PR #122, opened 2015-05-22
Use max-uploads instead of max-keys for ListMultipartUploadsResult and limit listing to 1000

harshavardhana:


harshavardhana:
LGTM

PR #128, opened 2015-05-29
SetBucketACL failes on an existing bucket on S3 - fixes #127

### LINKED ISSUE #127: SetBucketACL fails on an existing bucket on s3
The reason here is that we are not setting up "acl" url query string, so a setBucketACL operations ends up becoming a regular makeBucket operation.

harshavardhana:


harshavardhana:
LGTM

PR #172, opened 2015-06-10
Version number is pulled in via manifest variable. Defaults to 'dev' when executing directly from .class instead of .jar

fkautz:


fkautz:
LGTM

PR #208, opened 2015-07-12
getObject Offset and length are now range checked. Fixes #199

### LINKED ISSUE #199: get object with <= 0 length should fail before sending request

fkautz:


fkautz:
LGTM

PR #210, opened 2015-07-13
Getting object with byte range now works. Fixes #201

### LINKED ISSUE #201: get object range should not return full object
```java
        InputStream smallobject = client.getObject(bucket, "smallobject", 0, 5);
        byte[] bytes = streamToByteArray(smallobject);
        Assert.assertEquals("hello", new String(bytes, "UTF-8"));
```
Expected: hello
Actual: hello world

fkautz:


fkautz:
LGTM

PR #211, opened 2015-07-13
Adds get object with offset and implicit length. Fixes #200.

### LINKED ISSUE #200: Add get object with offset specifier
It is useful for the user to be able to say getObject(bucket, key, offset). Currently, the user must specify the offset and length getObject(bucket, key, offset, length).

fkautz:


fkautz:
LGTM

PR #213, opened 2015-07-13
Common prefixes are now returned in list objects. Fixes #202.

### LINKED ISSUE #202: list objects with recursion disabled doesn't return common prefixes

fkautz:


fkautz:
LGTM

PR #220, opened 2015-07-13
Removing date from list of signed headers

fkautz:


harshavardhana:
Did you see why user-agent was still being sent along with signedheaders, it is part of ignoredheader list. 

~~~
accept-encoding;date;host;user-agent;x-amz-content-sha256;x-amz-date
~~~

fkautz:
LGTM

PR #227, opened 2015-07-15
Removing ForbiddenException, using AccessDeniedException. BucketExists returns false only on error code BucketNotFound.

fkautz:


fkautz:
LGTM

PR #229, opened 2015-07-15
Setting redirect to false for all requests. Fixes #223.

### LINKED ISSUE #223: ForbiddenException should be thrown instead of boolean as false for bucketExists() on unauthenticated private bucket
~~~
scala> var client = Client.getClient("https://s3.amazonaws.com")
scala> client.bucketExists("ferenginar")
res2: Boolean = false

scala>
~~~

fkautz:


fkautz:
LGTM

PR #239, opened 2015-07-22
Migrate from google-http-client to okHttp library

harshavardhana:


harshavardhana:
Removed tests for now, will add them back.

PR #252, opened 2015-08-10
Signed headers list in canonical header must be in lower case. Was ca…

krishnasrinivas:
…using auth failure with s3.


harshavardhana:
LGTM

PR #268, opened 2015-09-03
Use Google Guava lib

vadmeste:


harshavardhana:
We shouldn't be making changes to minio-java project specifically for android support, but in general this is a nice dependency to have. 

It is best to create a new project to write android specific library as minio-java-android. Please change the title of this patch as.

harshavardhana:
For example here https://github.com/aws/aws-sdk-android/tree/master/aws-android-sdk-s3

vadmeste:
I do understand your point, but I have just checked and it seems that only this patch is needed to make minio-java compatible with Android (at least with the last API version) and I successfully tested MakeBucket, Put/Get Object examples under my smartphone

harshavardhana:
> I do understand your point, but I have just checked and it seems that only this patch is needed to make minio-java compatible with Android (at least with the last API version) and I successfully tested MakeBucket, Put/Get Object examples under my smartphone

It is good to know that it works, but regardless of that there are many other things which need to be addressed for android. Officially as it stands we are seeing android s3 library to be very limited use case.

vadmeste:
Gotcha ! I changed the commit message if you still want to merge this PR

harshavardhana:
Thanks, this is good.

PR #280, opened 2015-10-28
Implement presigned POST

krishnasrinivas:


harshavardhana:
> Let me know if you have any other comments, so that i can avoid sending multiple pushes.

There are none.

PR #303, opened 2015-12-09
fix endpoint validation check

balamurugana:
If endpoint is not an url, validation is done for a hostname, IPv4 or
IPv6 address


harshavardhana:
LGTM :+1:

PR #323, opened 2015-12-24
new getObject() to store downloaded object to given file

balamurugana:
various fixes related to object methods.


balamurugana:
Please share your thoughts for my replies.

harshavardhana:
> in short, Range header works like range selection of a list in python. Do we really need to support such things?

The question is about GetPartialObject is meant to be fully RFC style and yes "-rangeStart" is a valid case. 

On the other hand i was thinking through this - 

GetPartialObject should work like SectionReader - https://golang.org/pkg/io/#NewSectionReader , Since all you are reading is sections and it makes sense to write it in a manner that it takes offset and length

Offset to indicate starting point and length to indicate the amount of bytes to read.  This makes it a more cleaner API. What we expose here is too low level for a API user.

harshavardhana:
> It doesn't work eg. GET https://miniocloud.s3.amazonaws.com/hello-world range 5-0 gives entire object content.
> 
> I would say below are valid cases
> 1. if startRange and endRange are null, download entire object.
> 2. if startRange >= 0 and endRange >= 0, download specified range.
> 3. if startRange >=0 and endRange is null, download object from startRange till EOF
> 4. if startRange is null and endRange >= 0, download object from beginning of the object till endRange.
> 
> I feel we would avoid handling other types.

Yes it shouldn't work the case of 'stopRange' being '0' means that Range header has following values

```
Range: 5- 
```

not

```
Range: 5-0
```

The current implementation problem is you are expecting always to have the end length, it is not necessary since RFC allows and S3 honors it.  It also makes unnecessary statObject() operation when its not even needed. 

This will also simplify fgetObject() since you only have to request offsets not the end length, and let S3 return back the entire object starting from that 'offset'.

harshavardhana:
> i think 5L * 1024 * 1024 * 1024 * 1024 is good enough. need to check that.

Ah okay yes works both ways.

harshavardhana:
LGTM

PR #342, opened 2016-01-19
fix: use XML pull parser for every instance of XmlEntity

### LINKED ISSUE #339: parallel multipart put fails inconsitently
When ```MinioClient.putObject()``` is used in a multiple threads to achieve upload different objects throw XML parsing/EOF Exceptions inconsistently.
However ```MinioClient.putObject()``` works fine for smaller object ie <= 5MiB in many more threads.

balamurugana:
Previously there single XML pull parser which is used by XmlEntity and
their children classes.  This causes parsing error when doing threaded
execution.  This patch fixes this.

Closes #339


harshavardhana:
Other than the comment everything else LGTM

balamurugana:
I have fixed the comment in the latest sync.

PR #344, opened 2016-01-20
createRequest: Create request now handles buckets with '.'.

harshavardhana:
- accessing bucket names right after they were created. In current style
  if we try to access the buckets we would get 'Access Denied'.
  This can be solved by sending the requests directly to the endpoint
  where the bucket was created. For this we have to automatically use
  the relevant endpoint based on bucket location, similar to how it is
  achieved in 'aws-sdks'.

- bucket names with '.' cannot be used with virtual host style for
  HTTPS requests due to SSL certificate issue, so in this case we fallback
  to using path style instead for such requests, similar to how it is achieved
  in 'aws-sdks'.


balamurugana:
:+1: LGTM

PR #345, opened 2016-01-25
presigned: Input expires should be time.Duration

### LINKED ISSUE #300: All arguments for all functions should have native types
For example in Golang in PresignedGetObject

```
func (a API) PresignedGetObject(bucket, object string, expires time.Duration) (string, error) {
```

Expires argument is time.Duration which is a native type under Golang ecosystem, we should make sure that wherever possible we should take native types.

harshavardhana:
Fixes #300

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@ language: java
 sudo: false
 jdk:
 - oraclejdk8
-- oraclejdk7
-- openjdk7
 notifications:
   slack:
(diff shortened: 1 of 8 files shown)


harshavardhana:
@balamurugana ?

.travis.yml
```diff
@@ -2,8 +2,6 @@ language: java
 sudo: false
 jdk:
 - oraclejdk8
-- oraclejdk7
-- openjdk7
```
balamurugana: does it mean we don't support only oraclejdk8?

src/main/java/io/minio/MinioClient.java
```diff
@@ -1015,11 +1021,12 @@ public String presignedPutObject(String bucketName, String objectName, Integer e
    * @throws NoSuchAlgorithmException upon requested algorithm was not found during signature calculation
    * @throws InvalidExpiresRangeException upon input expires is out of range
    */
-  public String presignedPutObject(String bucketName, String objectName)
+  public String presignedDEFAULT_EXPIRY_TIMEPutObject(String bucketName, String objectName)
```
balamurugana: looks like a typo or wrong copy/paste :)

harshavardhana:
We will stick to JDK7 for now.

PR #369, opened 2016-05-08
doc: add javadoc to all methods and classes.

balamurugana:


src/main/java/io/minio/DateFormat.java
```diff
@@ -20,6 +20,9 @@
 import org.joda.time.format.DateTimeFormatter;
 
 
+/**
+ * Global constants for various date formats.  They are used to convert string to DateTime object and vise verse.
```
harshavardhana: Typo here - s/vise verse/vice versa/

PR #385, opened 2016-05-27
Portnumber should not be ignored if a valid endpoint is specified

sdkrum:
Hi! 

was easier then thought, and if you have a good test coverage, i did not even break something...

### DIFF
--- a/src/main/java/io/minio/MinioClient.java
+++ b/src/main/java/io/minio/MinioClient.java
@@ -382,26 +382,6 @@ public final class MinioClient {
       throw new InvalidEndpointException("(null)", "null endpoint");
     }
 
-    // for valid URL endpoint, port and secure are ignored
-    HttpUrl url = HttpUrl.parse(endpoint);
-    if (url != null) {
-      if (!"/".equals(url.encodedPath())) {
-        throw new InvalidEndpointException(endpoint, "no path allowed in endpoint");
-      }
-
-      // treat Amazon S3 host as special case
-      String amzHost = url.host();
-      if (amzHost.endsWith(".amazonaws.com") && !amzHost.equals("s3.amazonaws.com")) {
-        throw new InvalidEndpointException(endpoint, "for Amazon S3, host should be 's3.amazonaws.com' in endpoint");
-      }
-
-      this.baseUrl = url;
-      this.accessKey = accessKey;
-      this.secretKey = secretKey;
-
-      return;
-    }
-
     // endpoint may be a valid hostname, IPv4 or IPv6 address
     if (!this.isValidEndpoint(endpoint)) {
       throw new InvalidEndpointException(endpoint, "invalid host");
--- a/src/main/java/io/minio/MinioClient.java
+++ b/src/main/java/io/minio/MinioClient.java
@@ -152,7 +152,7 @@ public final class MinioClient {
    * @see #MinioClient(String endpoint, int port, String accessKey, String secretKey, boolean secure)
    */
   public MinioClient(String endpoint) throws InvalidEndpointException, InvalidPortException {
-    this(endpoint, 0, null, null, endpoint==null?false:(endpoint.startsWith("https://")));
+    this(endpoint, 0, null, null, endpoint==null?false:endpoint.startsWith("https://"));
   }
 
 
@@ -225,7 +225,7 @@ public final class MinioClient {
    */
   public MinioClient(String endpoint, String accessKey, String secretKey)
     throws InvalidEndpointException, InvalidPortException {
-    this(endpoint, 0, accessKey, secretKey, endpoint==null?false:(endpoint.startsWith("https://")));
+    this(endpoint, 0, accessKey, secretKey, endpoint==null?false:endpoint.startsWith("https://"));
   }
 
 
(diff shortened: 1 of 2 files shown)


src/main/java/io/minio/MinioClient.java
```diff
@@ -382,26 +382,6 @@ public MinioClient(String endpoint, int port, String accessKey, String secretKey
       throw new InvalidEndpointException("(null)", "null endpoint");
     }
 
-    // for valid URL endpoint, port and secure are ignored
-    HttpUrl url = HttpUrl.parse(endpoint);
```
harshavardhana: This won't be the right fix.  This is needed.

sdkrum: Ah, not sure, but at least all tests pass nicely. Here comes why i dropped the whole block: My first try was to just change the port number in the HttpUrl object, but, that field is final, so no chance there. So what i had to do in case the parameter port and the port in the HttpUrl object did not map, was to use the code following in the lines after (the case where url is null), what left the creation of the url object a kind of useless.

sdkrum: Oh sorry, now the tests are failing here... see your point...

sdkrum: hm, the problem is caused by some assumptions made in the different constructors. Seems not as easy:

 `
public MinioClient(String endpoint) throws InvalidEndpointException, InvalidPortException {
    this(endpoint, 0, null, null, true);
  }
`

Strange assumption, that this has to be a secure connection, if the endpoint is not interpreted. I try some more cleanup...

harshavardhana: > hm, the problem is caused by some assumptions made in the different constructors. Seems not as easy:
> 
>  public MinioClient(String endpoint) throws InvalidEndpointException, InvalidPortException { this(endpoint, 0, null, null, true); } 
> 
> Strange assumption, that this has to be a secure connection, if the endpoint is not interpreted. I try some more cleanup...

Yes it is tricky it is implemented all through for a reason.

sdkrum: Ok, there you should have a version which compiles and tests successfully. In my installation two tests are failing: 

- testSigningKey
- testGetObjectHeaders

But they do even if i try to run the untouched project, i assume that has something to do with my locale, as the code has problems to parse the last modified date correctly.

harshavardhana:
// cc @balamurugana for your review.

src/main/java/io/minio/MinioClient.java
```diff
@@ -152,7 +152,7 @@
    * @see #MinioClient(String endpoint, int port, String accessKey, String secretKey, boolean secure)
    */
   public MinioClient(String endpoint) throws InvalidEndpointException, InvalidPortException {
-    this(endpoint, 0, null, null, true);
+    this(endpoint, 0, null, null, endpoint==null?false:endpoint.startsWith("https://"));
```
balamurugana: As ie mentioned in the comment, if endpoint is `localhost` or `myminio.com`, they are valid and should use HTTPS by default.
This check make it false.

src/main/java/io/minio/MinioClient.java
```diff
@@ -172,7 +172,7 @@ public MinioClient(String endpoint) throws InvalidEndpointException, InvalidPort
    * @see #MinioClient(String endpoint, int port, String accessKey, String secretKey, boolean secure)
    */
   public MinioClient(URL url) throws NullPointerException, InvalidEndpointException, InvalidPortException {
-    this(url.toString(), 0, null, null, true);
+    this(url.toString(), 0, null, null, url.getProtocol().equals(Scheme.HTTPS.toString()));
```
balamurugana: Its better to use like `Scheme.fromString(url.getProtocol()) == Scheme.HTTPS` because `Scheme.fromString()` works case insensitively.

src/main/java/io/minio/MinioClient.java
```diff
@@ -306,7 +306,7 @@ public MinioClient(HttpUrl url, String accessKey, String secretKey)
    */
   public MinioClient(String endpoint, int port, String accessKey, String secretKey)
     throws InvalidEndpointException, InvalidPortException {
-    this(endpoint, port, accessKey, secretKey, true);
+    this(endpoint, port, accessKey, secretKey, endpoint==null?false:endpoint.startsWith("https://"));
```
balamurugana: same as above

src/main/java/io/minio/MinioClient.java
sdkrum: Ah, ok, that means in the end, that even if i specify "http://localhost" as the endpoint, i end up with an https connection? I think that is confusing, bu it is your api. And as long as your suggested changes makes it possible to get an connection when specifying a valid endpoint and a port number which really gets to the given port number and not to the default i am fine. 

But what about a policy to handle back https in any case where the endpoint does not explicitly specify something else? then 
`endpoint==null?true:endpoint.startsWith("http://")` would work just fine?

balamurugana: @sdkrum i have fixed those and sent a pull request https://github.com/UNINETT/minio-java/issues/1.  I guess you haven't seen my private message in gitter.  Can you please check whether that is ok?

balamurugana:
As per the talk with @sdkrum, this issue is covered in https://github.com/minio/minio-java/pull/392.
Closing this pull request.

PR #392, opened 2016-06-07
fix: override given port and scheme in the endpoint

balamurugana:


harshavardhana:
why @balamurugana both patches in the same PR?

balamurugana:
> why @balamurugana both patches in the same PR?

Made two pull request now.

I think we would need to change from 'Merge' to 'Cherry-picking' for minio-java repo.

balamurugana:
> continuous-integration/appveyor/pr — AppVeyor was unable to build non-mergeable pull

I think we need to run this manually.

harshavardhana:
> I think we need to run this manually.

You can edit the commit and resubmit the PR it will run.

harshavardhana:
> I think we would need to change from 'Merge' to 'Cherry-picking' for minio-java repo.

That is just toggling the github option.

harshavardhana:
@balamurugana can you also update the functional/unite test with this new change?

balamurugana:
> @balamurugana can you also update the functional/unite test with this new change?

Currently we don't have a way to test constructors ie this patch updates constructors to allow overriding given port/scheme and we can't validate it back with mock server.

I will check and update those tests.

balamurugana:
> I will check and update those tests.

I see in [MockWebServer.java](https://github.com/square/okhttp/blob/master/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java#L298) its recommended not to use fixed port no. 

Need to check more how to run mock server on specified port.

PR #400, opened 2016-06-15
Patch 3

deekoder:
- Brought it as close as possible to docs site.
- Qualified examples as QS Example vs Full Example vs Full Application Example
- Added Explore Further section
- Added Constructor and Table so that doc is fully complete and contained in itself.


harshavardhana:
We should keep Commit messages appropriate 'Update README.md' is too generic.

PR #405, opened 2016-06-16
Update Readme

deekoder:
1. Changed desc to description in the table heading.
2. Call it Full Docs Site  to Complete Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,8 +41,8 @@ MinioClient minioClient = new MinioClient("https://play.minio.io:9000", "Q3AM3UQ
 
 ```
 
-| Params     | Desc |  
-| :------- | :---- |  
+| Params     | Description |  
+| :------- | :------------ |  
 | Endpoint | URL to object storage service. |  
 | Access Key    | Access key is like user ID that uniquely identifies your account.   |   
 | Secret Key     | Secret key is the password to your account.    |
@@ -150,7 +150,7 @@ The full API Reference is available here.
 
  
 ## Explore Further
-* [docs.minio.io](https://docs.minio.io) - Full Docs Site
+* [Complete Documentation](https://docs.minio.io)
 * [Minio Java Client SDK API Reference](https://docs.minio.io/docs/java-client-api-reference) 
 * [Build your own Photo API Service - Full Application Example ](https://docs.minio.io/docs/java-photo-api-service)
 


harshavardhana:
We shouldn't merge patches with wrong 'commit' message 'Update Readme' lacks context.

If you look at the current set of commits - https://github.com/minio/minio-java/commits/master - its unclean any one pulling the project now will see a whole bunch of commits just editing one file with same commit messages for all. 

I merged it previously thinking that it is a one time thing which is okay, but its already 7 commits now with the same 'commit message' and it is not okay.  

Another thing is we didn't wait for the CI to finish, we should wait for the CI's to finish its job regardless of it being just a documentation edit - https://ci.appveyor.com/project/Harshavardhana/minio-java/build/256

GarimaKapoor:
It has the context in the description. Because we are updating the Github readme.md to be in sync with the docs.minio.io. It is natural to have multiple commits since readme.io is buggy and the copied text has quite a bit of errors. Also Minio - Java is the test sync up so it is natural to go through multiple iterations before getting it to the level we want it.

I did not realize that CI needs to be finished for the doc edit. Is it necessary to have it done since it is just text that we are editing. Is it just for discipline or does it do something for the text edits.

harshavardhana:
> It has the context in the description. Because we are updating the Github readme.md to be in sync with the docs.minio.io. It is natural to have multiple commits since readme.io is buggy and the copied text has quite a bit of errors. Also Minio - Java is the test sync up so it is natural to go through multiple iterations before getting it to the level we want it.
> 

That is not a proper practice, we don't commit saying 'Update file.java' and provide a description. A commit log should capture the essence of the description. For more examples we should look at other projects how they commit changes.

Now if its an iteration/update and we are reviewing should happen to the same patch using 'git push and git commit --amend' locally and the aforementioned patch set should be updated until it is ready to be merged and cleanly formatted. There is no reason to make multiple commits for each small edits, typos etc. 

> I did not realize that CI needs to be finished for the doc edit. Is it necessary to have it done since it is just text that we are editing. Is it just for discipline or does it do something for the text edits.

It is a practice otherwise we could hurry every single change, all it takes is another 10 mins of our time.

What i am going to do is squash all the changes for README.md into a single commit saying "docs: sync minio-java README with docs.minio.io" and force push to minio-java for now.

harshavardhana:
https://github.com/minio/minio-java/commit/9a7a6097da0c786e89b3925456edeea9fb3703e8 - commits are now squashed and now this looks clean. 

This is what is expected - https://github.com/minio/minio-java/commits/master

GarimaKapoor:
got it! thanks. But I am still not convinced about CI needs for the simple edits :)

PR #417, opened 2016-07-08
doc:API.md/Minor edits to bring it in sync with docs.minio.io

koolhead17:


harshavardhana:
The commit message seems to be the same as previous one, can you fix that?

PR #441, opened 2016-08-07
Doc: API.md/This patch will replace tables in existing HTML format to markdown.

koolhead17:
Changes:
* Table HTML changed to Markdown.
*  Replaced examples URL from play to s3, this is a uniform style we are following in all other client libraries.


harshavardhana:
// cc @balamurugana  can you review this?

PR #465, opened 2016-10-07
api: putObject() should properly save contentType.

### LINKED ISSUE #464: Content Type not updating on S3 - Java Client Lib
Hi,

I am using java client library of minio. I am trying to upload files using putObject method but i noticed that when i pass all the details along with content type. It is not changing content type of file on s3. It is always remain as binary/octet-steam. Because of this when i try to access any file in browser it directly go to download instead of displaying it on html page or display as image. 

Can you please fix this asap. 

Thank you,

Regard,
Jani Parthiv

harshavardhana:
Current code was not working when contentType is set
to a custom value.

Fixes #464


balamurugana:
LGTM :+1:

Issue #463, opened 2016-10-03
Implement CopyObject API


harshavardhana:
For reference visit https://docs.minio.io/docs/golang-client-api-reference#CopyObject

NitishT:
Hey @harshavardhana mind if I give this a shot?

harshavardhana:
> Hey @harshavardhana mind if I give this a shot?

Feel free to @NitishT - Thanks

NitishT:
> Feel free to @NitishT - Thanks

Cool. Thanks!

balamurugana:
@NitishT :+1:   If you have any trouble in current code base, feel free to ask me.

NitishT:
Sure! Will do that @balamurugana

NitishT:
Hi @balamurugana I have code & JUnit tests done, but I wanted to add a Functional test as well. 

I see the file FunctionalTest.java but not sure how to run those tests. Can you please tell me how to run those in eclipse?

balamurugana:
@NitishT Currently functional tests are run by hand using `javac` and `java` commands.  I am trying to add a new task for functional test which can be run on requirement like `./gradlew functionaltest`

balamurugana:
@NitishT FYI https://github.com/minio/minio-java/pull/473

GarimaKapoor:
@NitishT what is the status of the issue?

NitishT:
There were a few enhancements/review comments to be addressed. I will take it up this week @GarimaKapoor

PR #477, opened 2016-10-31
Implemented CopyObject API #463

NitishT:


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1381,25 +1382,106 @@ public void getObject(String bucketName, String objectName, String fileName)
...
+      InvalidArgumentException {
+    if (copyConditions == null) {
+      throw new InvalidArgumentException("Copy conditions should be a non null value");
+    }
+    if (objectSource == null) {
+      throw new InvalidArgumentException("Invalid object source");
+    }
+    String sourceBucketName = getBucketNameFromPath(objectSource);
+    String sourceObjectName = getObjectNameFromPath(objectSource);
+    ObjectStat objectStat = statObject(sourceBucketName, sourceObjectName);
```
harshavardhana: you don't need to do statObject(),  you can send the request to server and server will fail appropriately.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1381,25 +1382,106 @@ public void getObject(String bucketName, String objectName, String fileName)
...
+    String etag = objectStat.etag();
+    Date lastModified = objectStat.createdTime();
+    long length = objectStat.length();
+
+    // If the conditions specified are satisfied, get and put the object in new location
+    if (copyConditions.isUnmodifiedSince(lastModified) && copyConditions.isModifiedAfter(lastModified)
+        && copyConditions.etagMatches(etag) && copyConditions.etagDoesNotMatch(etag)) {
+      InputStream objectByteStream = getObject(sourceBucketName, sourceObjectName);
+      putObject(bucketName, objectName, objectByteStream, length, "application/octet-stream");
```
harshavardhana: CopyObject() is a server side API which does server side copy. This is inconsistent behavior and unexpected.  You have to take a look at the minio-go implementation or python implementation to understand. 

CopyConditions are part of HTTP headers which need to be set and server validates these conditions.

harshavardhana: https://docs.minio.io/docs/python-client-api-reference#copy_object - here you go. 

https://github.com/minio/minio-py/blob/master/minio/api.py#L769 - For the reply you need to parse CopyObjectResult XML.

harshavardhana:
Thanks for the patch - @NitishT lets discuss further on this tomorrow  http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectCOPY.html

NitishT:
Thanks for the quick review @harshavardhana 

Sure, lets discuss tomorrow.

api/src/main/java/io/minio/CopyConditions.java
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * A container class to hold all the Conditions to be checked before copying an object.
+ */
+public class CopyConditions {
+
+  Date dateUnmodifiedSince;
+  Date dateModifiedAfter;
+  String matchETag;
+  String exceptETag;
```
balamurugana: Are these names from minio-go?  It would be clear name like (as similar as the spec)
```
String matchEtag;
String noneMatchEtag;
String unmodifiedSince;
String modifiedSince;
```

Also, please have setter methods names prefixing `set` and member variable name.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -78,6 +78,7 @@
 import java.nio.file.StandardOpenOption;
 import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
+import java.util.Date;
```
balamurugana: Please use `import org.joda.time.DateTime;`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1381,25 +1382,106 @@ public void getObject(String bucketName, String objectName, String fileName)
...
+   * @throws InvalidBucketNameException
+   *           upon an invalid bucket name
+   * @throws InvalidKeyException
+   *           upon an invalid access key or secret key
+   * @throws IOException
+   *           upon signature calculation failure
+   * @throws NoSuchAlgorithmException
+   *           upon requested algorithm was not found during signature calculation
+   * @throws InvalidExpiresRangeException
+   *           upon input expires is out of range
```
balamurugana: IMO you don't need to format these javadoc.  If that's the standard, you could send a separate patch formatting every javadoc all over.

functional/FunctionalTest.java
```diff
@@ -802,6 +872,9 @@ public static void main(String[] args) {
 
       presignedPostPolicy_test();
 
+      copyObject_test1();
+      copyObject_test2();
```
balamurugana: you could have four or more functions here.

1. testing without any condition
2. testing with all condition
3. testing with specific condition
4. testing with combination of conditions like

```
Consideration 1 – If both of the x-amz-copy-source-if-match and x-amz-copy-source-if-unmodified-since headers are present in the request as follows:

x-amz-copy-source-if-match condition evaluates to true, and;

x-amz-copy-source-if-unmodified-since condition evaluates to false;

then, S3 returns 200 OK and copies the data.

 

Consideration 2 – If both of the x-amz-copy-source-if-none-match and x-amz-copy-source-if-modified-since headers are present in the request as follows:

x-amz-copy-source-if-none-match condition evaluates to false, and;

x-amz-copy-source-if-modified-since condition evaluates to true;

then, S3 returns 412 Precondition Failed response code. 
```

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1381,25 +1382,106 @@ public void getObject(String bucketName, String objectName, String fileName)
...
+      NoResponseException, ErrorResponseException, InternalException, IOException, XmlPullParserException,
+      InvalidArgumentException {
+    if (copyConditions == null) {
+      throw new InvalidArgumentException("Copy conditions should be a non null value");
+    }
+    if (objectSource == null) {
+      throw new InvalidArgumentException("Invalid object source");
+    }
+    String sourceBucketName = getBucketNameFromPath(objectSource);
+    String sourceObjectName = getObjectNameFromPath(objectSource);
```
NitishT: This is as per the ref given: https://docs.minio.io/docs/golang-client-api-reference#CopyObject

NitishT: But I agree, if possible, it will cleaner to have `srcObjectName` and `srcBucketName` here instead of `objectSource`.

PR #478, opened 2016-11-01
Implemented CopyObject API #463

NitishT:


api/src/main/java/io/minio/CopyConditions.java
```diff
@@ -0,0 +1,92 @@
...
+import org.joda.time.DateTime;
+
+import io.minio.errors.InvalidArgumentException;
+
+/**
+ * A container class to hold all the Conditions to be checked before copying an object.
+ */
+public class CopyConditions {
+
+  private Map<String, String> copyConditions = new HashMap<>();
```
balamurugana: 1. You could use Hashtable to avoid key/value as null.
2. just naming `conditions` is good enough.

api/src/main/java/io/minio/CopyConditions.java
```diff
@@ -0,0 +1,92 @@
...
+    copyConditions.put("x-amz-copy-source-if-match", etag);
+  }
+
+  /**
+   * Set matching ETag except condition.
+   * 
+   * @throws InvalidArgumentException
+   *           When invalid arguement is passed
+   */
+  public void setMatchETagExcept(String etag) throws InvalidArgumentException {
```
balamurugana: I think `setNoneMatchEtag()` would be better.

api/src/main/java/io/minio/CopyConditions.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+    copyConditions.put("x-amz-copy-source-if-none-match", etag);
+  }
+
+  /**
+   * Get copy conditions HashMap.
+   * 
+   */
+  public Map<String, String> getCopyConditions() {
+    return copyConditions;
```
balamurugana: 1. you could name this method as `conditions()`
2. return a copy of `return new Hashtable(copyConditions);`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1381,7 +1382,61 @@ public void getObject(String bucketName, String objectName, String fileName)
     }
   }
 
-
+  /**
+   * Copy a source object into a new object with the provided name in the provided bucket.
```
balamurugana: Copy source object into given object name.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1381,7 +1382,61 @@ public void getObject(String bucketName, String objectName, String fileName)
...
+   *          CopyConditions object with collection of supported CopyObject conditions.
+   *
+   * @throws InvalidBucketNameException
+   *           upon an invalid bucket name
+   * @throws NoSuchAlgorithmException
+   *           upon requested algorithm was not found during signature calculation
+   * @throws InvalidKeyException
+   *           upon an invalid access key or secret key
+   */
+  public CopyObjectResult copyObject(String bucketName, String objectName, String objectSource,
```
balamurugana: 1. Returning CopyObjectResult is not useful. The method should return void if success and throws exception on failure.
2. Having `String srcBucketName, String srcObjectName` than `String objectSource` is better.  You would need to validate srcBucketName as well.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1381,7 +1382,61 @@ public void getObject(String bucketName, String objectName, String fileName)
...
+    }
+
+    HttpResponse response = execute(Method.PUT, US_EAST_1, bucketName, objectName, headerMap, null, null, "", 0);
+    
+    CopyObjectResult result = new CopyObjectResult();
+    result.parseXml(response.body().charStream());
+    response.body().close();
+    return result;
+  }
+
```
balamurugana: We would need to have one more version like
```
public void copyObject(String bucketName, String objectName, String srcBucketName, String srcObjectName)
```

api/src/main/java/io/minio/CopyConditions.java
```diff
@@ -0,0 +1,92 @@
...
+   * Sets object modified after condition.
+   * 
+   * @throws InvalidArgumentException
+   *           When invalid arguement is passed
+   */
+  public void setModified(DateTime date) throws InvalidArgumentException {
+    if (date == null) {
+      throw new InvalidArgumentException("Date can not be null");
+    }
+    copyConditions.put("x-amz-copy-source-if-modified-since", date.toString());
```
balamurugana: What is the date format expected for this header?

api/src/main/java/io/minio/MinioClient.java
harshavardhana: Why is it useful?

api/src/main/java/io/minio/MinioClient.java
harshavardhana: We don't need to validate this server does the validate in this scenario.

api/src/main/java/io/minio/CopyConditions.java
bsideup: Hashtable shouldn't be used in a modern Java app.

api/src/main/java/io/minio/CopyConditions.java
bsideup: by convention in Java getters should start with "get", so "getConditions"
instead of Hashtable, it's better to use https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#unmodifiableMap(java.util.Map)

api/src/main/java/io/minio/CopyConditions.java
balamurugana: If so, why am I not getting errors when  `./gradlew build`.  I use oracle jdk8 `1.8.0_111`.

bsideup: Because your build is configured so :) You can configure some code linting tools, they'll cry about it.

From the official JavaDoc:
http://docs.oracle.com/javase/7/docs/api/java/util/Hashtable.html

> As of the Java 2 platform v1.2, this class was retrofitted to implement the Map interface, making it a member of the Java Collections Framework. Unlike the new collection implementations, Hashtable is synchronized. If a thread-safe implementation is not needed, it is recommended to use HashMap in place of Hashtable. If a thread-safe highly-concurrent implementation is desired, then it is recommended to use ConcurrentHashMap in place of Hashtable.

api/src/main/java/io/minio/CopyConditions.java
balamurugana: We use okhttp style of keeping the object immutable.  That's why point 2.

bsideup: I don't mind to make it immutable, just use `Collections.unmodifiableMap()` instead of `new Hashtable()`

api/src/main/java/io/minio/CopyConditions.java
balamurugana: Got it.  Thanks :smile: 

 @NitishT you could consider running some java linting tools and use it accordingly.

api/src/main/java/io/minio/CopyConditions.java
balamurugana: :+1:

harshavardhana:
Closing this PR and moving this to a new PR.

PR #490, opened 2016-11-29
multipart: send content-length in POST requests

vadmeste:
Send Content-Lenght header in POST requests to satisfy some proxies
(like squid) which require that header to work.

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -680,7 +680,15 @@ public final class MinioClient {
 
         @Override
         public long contentLength() {
+          if (body instanceof InputStream || body instanceof RandomAccessFile || body instanceof byte[]) {
             return length;
+          }
+
+          if (length == 0) {
+            return -1;
+          } else {
+            return length;
+          }
         }
 
         @Override
@@ -788,6 +796,11 @@ public final class MinioClient {
     String contentType = null;
     if (headerMap != null) {
       contentType = headerMap.get("Content-Type");
+    }
+    if (body != null && !(body instanceof InputStream || body instanceof RandomAccessFile || body instanceof byte[])) {
+      byte[] bytes = body.toString().getBytes(StandardCharsets.UTF_8);
+      body = bytes;
+      length = bytes.length;
     }
 
     Request request = createRequest(method, bucketName, objectName, region,
@@ -1045,13 +1058,6 @@ public final class MinioClient {
            InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
            InternalException {
     updateRegionCache(bucketName);
-    int len = -1;
-    if (data instanceof String) {
-      len = ((String) data).length();
-    }
-    if (data instanceof CompleteMultipartUpload) {
-      len = ((CompleteMultipartUpload) data).toString().length();
-    }
     return execute(Method.POST, BucketRegionCache.INSTANCE.region(bucketName),
                    bucketName, objectName, headerMap, queryParamMap,
                    data, 0);
@@ -2489,7 +2495,7 @@ public final class MinioClient {
 
     String policyJson = policy.getJson();
 
-    HttpResponse response = executePut(bucketName, null, headerMap, queryParamMap, policyJson, policyJson.length());
+    HttpResponse response = executePut(bucketName, null, headerMap, queryParamMap, policyJson, 0);
     response.body().close();
   }
 
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -680,15 +680,7 @@ public final class MinioClient {
 
         @Override
         public long contentLength() {
-          if (body instanceof InputStream || body instanceof RandomAccessFile || body instanceof byte[]) {
             return length;
-          }
-
-          if (length == 0) {
-            return -1;
-          } else {
-            return length;
-          }
         }
 
         @Override
@@ -1053,6 +1045,13 @@ public final class MinioClient {
            InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
            InternalException {
     updateRegionCache(bucketName);
+    int len = -1;
+    if (data instanceof String) {
+      len = ((String) data).length();
+    }
+    if (data instanceof CompleteMultipartUpload) {
+      len = ((CompleteMultipartUpload) data).toString().length();
+    }
     return execute(Method.POST, BucketRegionCache.INSTANCE.region(bucketName),
                    bucketName, objectName, headerMap, queryParamMap,
                    data, 0);


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1019,7 +1011,7 @@ private HttpResponse executeDelete(String bucketName, String objectName, Map<Str
     updateRegionCache(bucketName);
     HttpResponse response = execute(Method.DELETE, BucketRegionCache.INSTANCE.region(bucketName),
                                     bucketName, objectName, null,
-                                    queryParamMap, null, null, 0);
+                                    queryParamMap, null, null, -1);
```
harshavardhana: this might not be true for 'executeDelete()' when we implement Multi object delete API.  Change this to executeDelete() itself taking the content length.

vadmeste: > this might not be true for 'executeDelete()' when we implement Multi object delete API. Change this to executeDelete() itself taking the content length.

Multi Delete is a POST and not DELETE http operation. I think we should keep it like that until we implement the new multi delete API.

harshavardhana: ah correct.

balamurugana:
@vadmeste The fix should be at https://github.com/minio/minio-java/blob/master/api/src/main/java/io/minio/MinioClient.java#L2543. Instead of sending `""`, we would need to send `new byte[0]`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -671,15 +671,7 @@ public MediaType contentType() {
```
balamurugana: This is not a right fix.
`requestBody` is created only if `body != null`.  As per `createRequest()`, `body` is any object or `null`.  If `body` is non-null and other than `(body instanceof InputStream || body instanceof RandomAccessFile || body instanceof byte[])`, we won't need to compute md5sum and need to send `Transfer-Encoding: chunked`

vadmeste: @balamurugana, if you focus on the function contentLength() itself, you can notice that the only change I've made is to return 0 instead of -1 when length == 0. Otherwise the behavior will be still the same.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1040,9 +1032,16 @@ private HttpResponse executePost(String bucketName, String objectName, Map<Strin
            InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
            InternalException {
     updateRegionCache(bucketName);
+    int len = -1;
+    if (data instanceof String) {
+      len = ((String) data).length();
+    }
+    if (data instanceof CompleteMultipartUpload) {
+      len = ((CompleteMultipartUpload) data).toString().length();
+    }
```
balamurugana: Having execute{Head,Get,Post,Put,Delete} are just wrappers to execute() for convenience and readability across APIs  Spreading the logic of type checking will lead to confusion in future and its not a good idea.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1875,7 +1874,7 @@ public void makeBucket(String bucketName, String region)
       configString = config.toString();
     }
 
-    executePut(bucketName, null, headerMap, null, US_EAST_1, configString, 0);
+    executePut(bucketName, null, headerMap, null, US_EAST_1, configString, configString.length());
```
balamurugana: `String.length()` != `[].byte.length`.

We would need to convert string into UTF8 encoded byte array.

vadmeste:
> @vadmeste The fix should be at https://github.com/minio/minio-java/blob/master/api/src/main/java/io/minio/MinioClient.java#L2543. Instead of sending "", we would need to send new byte[0]

We also need to send content-length in complete multipart POST request. The reason of this patch is to send Content-Length header required for some proxies in POST requests.

vadmeste:
> Improvement to existing createRequest() is it work only with body as null, InputStream, RandomAccessFile, []byte than any object. This way length should be taken as is in RequestBody.contentLength(). As I mentioned previously this will lead to md5sum calculation for any non-null body.

Not sure about your exact meaning here, could you ping me when you are available ?

api/src/main/java/io/minio/MinioClient.java
vadmeste: This is for creating bucket, do we really need to convert to UTF8 ?

As @balamurugana, suggested, we need to enhance createRequest for a simpler logic.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1040,9 +1032,16 @@ private HttpResponse executePost(String bucketName, String objectName, Map<Strin
            InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
            InternalException {
     updateRegionCache(bucketName);
+    int len = -1;
+    if (data instanceof String) {
```
harshavardhana: This type of special handling is not a good idea..  There is some issue here.

api/src/main/java/io/minio/MinioClient.java
harshavardhana: it is largely `configString.getBytes("UTF-8").length()`

vadmeste: > it is largely configString.getBytes("UTF-8").length()

yes, I meant that if this is about bucket creation, we won't have any unicode character inside the string, am I correct?

balamurugana: * In java, its required to set encoding to convert string into byte array, else you will get build error.
* In minio-java, `StandardCharsets.UTF_8` is used transparently ie S3 Rest API methods work with objects than strings/byte arrays.

harshavardhana:
Is there any pending here? @vadmeste

vadmeste:
> Is there any pending here? @vadmeste

@harshavardhana, I addressed some issues mentioned in comments but it looks that we need more change to simplify the overall architecture, @balamurugana had some ideas about simplification so I'll let him decide if that change should be done in a separate PR or not.

balamurugana:
> @harshavardhana, I addressed some issues mentioned in comments but it looks that we need more change to simplify the overall architecture, @balamurugana had some ideas about simplification so I'll let him decide if that change should be done in a separate PR or not.

@vadmeste Any other changes need to be a separate PR.

harshavardhana:
what is pending in this PR? @vadmeste @balamurugana

vadmeste:
@harshavardhana, this PR is ready

balamurugana:
> @harshavardhana, this PR is ready

@vadmeste This PR hasn't been addressed the review comment.  Let me send you a PR to your fork with those changes.

balamurugana:
> @harshavardhana, this PR is ready
> 
> @vadmeste This PR hasn't been addressed the review comment. Let me send you a PR to your fork with those changes.

I sent a PR https://github.com/vadmeste/minio-java/pull/1 to your fork.  Please review and merge it accordingly.

balamurugana:
@vadmeste Please rebase

PR #499, opened 2016-12-16
S3Endpoints: Added eu-west-2 to the endpoint list.

koolhead17:


api/src/main/java/io/minio/AwsS3Endpoints.java
```diff
@@ -42,6 +42,8 @@
     endpoints.put("eu-central-1", "s3-eu-central-1.amazonaws.com");
     // eu-west-1
     endpoints.put("eu-west-1", "s3-eu-west-1.amazonaws.com");
+    // eu-west-2
+    endpoints.put("eu-west-1", "s3-eu-west-1.amazonaws.com");
```
harshavardhana: eu-west-2 not eu-west-1

koolhead17: sorry. just fixed.

PR #503, opened 2016-12-22
examples should use https play endpoint.

harshavardhana:


balamurugana:
Is https only supported in `play.minio.io`?

harshavardhana:
> Is https only supported in play.minio.io ie no http?

http is but it re-directs to https so it is better to use https endpoint.

PR #507, opened 2016-12-23
getObjectUrl: add new api to get object URL.

### LINKED ISSUE #502: Add MinioClient.getUrl
The MinioClient class only provides the method presignedGetObject which returns a signed url for an object.

It would be useful to have a getUrl method to get a simple url for public files.

The AWS SDK has such a method: http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/AmazonS3Client.html#getUrl-java.lang.String-java.lang.String-

balamurugana:
The URL is only useful to retrieve the object's data if the object has
public read permissions.

Fixes #502

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -1143,6 +1143,39 @@ public final class MinioClient {
 
 
   /**
+   * Gets object's URL in given bucket.  The URL is ONLY useful to retrieve the object's data if the object has
+   * public read permissions.
+   *
+   * <p><b>Example:</b>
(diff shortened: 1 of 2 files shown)


HeheJD:
The api works fine in version 6.0.11. However, after we change version to the latest (8.5.7) it no longer exists.

balamurugana:
> The api works fine in version 6.0.11. However, after we change version to the latest (8.5.7) it no longer exists.

Use https://minio-java.min.io/io/minio/MinioAsyncClient.html#getPresignedObjectUrl-io.minio.GetPresignedObjectUrlArgs- with no credential provider i.e. with anonymous `MinioAsyncClient`

PR #510, opened 2017-01-04
Allow setting content encoding and a region in Post Policy.

MastaP:
The current Post Policy class is very restrictive. AWS API allows more fields to be included in the conditions. http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPOST.html
This PR enables adding Content-Encoding and also a region.
As this class is public, it can be (and is) used without Minio Client in certain applications. Our use case: prepare the policy in one service and send the form data to another service which works both with S3 and Minio endpoints.

### DIFF
--- a/api/src/main/java/io/minio/PostPolicy.java
+++ b/api/src/main/java/io/minio/PostPolicy.java
@@ -35,11 +35,13 @@ import java.util.*;
 public class PostPolicy {
   private static final String ALGORITHM = "AWS4-HMAC-SHA256";
 
-  private String bucketName;
-  private String objectName;
-  private boolean startsWith;
-  private DateTime expirationDate;
+  private final String bucketName;
+  private final String objectName;
+  private final boolean startsWith;
+  private final DateTime expirationDate;
+  private String region;
   private String contentType;
+  private String contentEncoding;
   private long contentRangeStart;
   private long contentRangeEnd;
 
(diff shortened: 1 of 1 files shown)


api/src/main/java/io/minio/PostPolicy.java
```diff
@@ -187,7 +217,7 @@ public String bucketName() {
     formData.put("x-amz-algorithm", ALGORITHM);
 
     DateTime date = new DateTime();
-    String region = BucketRegionCache.INSTANCE.region(this.bucketName);
+    String region = this.region != null ? this.region : BucketRegionCache.INSTANCE.region(this.bucketName);
```
balamurugana: As we already got the bucket name and its region, could you tell me why do we need explicit region setting?

MastaP: No, if PostPolicy is used independently (without the Client), BucketRegionCache is inaccessible

balamurugana: :+1:

MastaP: Any chance to have it merged and released soon? :) Thanks!

balamurugana: Introducing region setter would confuse existing users as nowhere in minio-java we ask region.

To make it compatible, have `formData()` method to be overloaded with another version like
```
public Map<String,String> formData(String accessKey, String secretKey, String region)
public Map<String,String> formData(String accessKey, String secretKey)
```

where the second one uses the first one with `BucketRegionCache`.

Hope this works for you.

MastaP:
@balamurugana Done. In order to keep the original formData() intact, had to extract makeFormData() since the region null-check adds 'throws InvalidArgumentException' to the new method signature.

MastaP:
@harshavardhana 
there are two public methods,
1. `public Map<String,String> formData(String accessKey, String secretKey)`
works as before, fills the region from the cache,
2. `public Map<String,String> formData(String accessKey, String secretKey, String region)`
throws `new InvalidArgumentException("empty region")` if no region is provided.

PR #511, opened 2017-01-13
api: Regression contentType set by user was never set properly.

harshavardhana:
The reason is because contentType is overriden by this
operation even if the contentType is set through header.

```
    public MediaType contentType() {
          if (contentType != null && !contentType.isEmpty()) {
            return MediaType.parse(contentType);
          } else {
            return MediaType.parse("application/octet-stream");
          }
        }
```

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -664,11 +664,16 @@ public final class MinioClient {
       requestBody = new RequestBody() {
         @Override
         public MediaType contentType() {
+          MediaType mediaType = null;
+
           if (contentType != null) {
-            return MediaType.parse(contentType);
-          } else {
-            return MediaType.parse("application/octet-stream");
+            mediaType = MediaType.parse(contentType);
           }
+          if (mediaType == null) {
+            mediaType = MediaType.parse("application/octet-stream");
+          }
+
+          return mediaType;
         }
 
         @Override
@@ -782,10 +787,15 @@ public final class MinioClient {
    */
   private HttpResponse execute(Method method, String region, String bucketName, String objectName,
                                Map<String,String> headerMap, Map<String,String> queryParamMap,
-                               String contentType, Object body, int length)
+                               Object body, int length)
     throws InvalidBucketNameException, NoSuchAlgorithmException, InsufficientDataException, IOException,
            InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
            InternalException {
+    String contentType = null;
+    if (headerMap != null) {
+      contentType = headerMap.get("Content-Type");
+    }
+
     Request request = createRequest(method, bucketName, objectName, region,
                                     headerMap, queryParamMap,
                                     contentType, body, length);
@@ -919,7 +929,7 @@ public final class MinioClient {
       queryParamMap.put("location", null);
 
       HttpResponse response = execute(Method.GET, US_EAST_1, bucketName, null,
-                                      null, queryParamMap, null, null, 0);
+                                      null, queryParamMap, null, 0);
 
       // existing XmlEntity does not work, so fallback to regular parsing.
       XmlPullParser xpp = xmlPullParserFactory.newPullParser();
@@ -983,7 +993,7 @@ public final class MinioClient {
     updateRegionCache(bucketName);
     return execute(Method.GET, BucketRegionCache.INSTANCE.region(bucketName),
                    bucketName, objectName, headerMap, queryParamMap,
-                   null, null, 0);
+                   null, 0);
   }
 
 
@@ -1000,7 +1010,7 @@ public final class MinioClient {
     updateRegionCache(bucketName);
     HttpResponse response = execute(Method.HEAD, BucketRegionCache.INSTANCE.region(bucketName),
                                     bucketName, objectName, null,
-                                    null, null, null, 0);
+                                    null, null, 0);
     response.body().close();
     return response;
   }
@@ -1020,7 +1030,7 @@ public final class MinioClient {
     updateRegionCache(bucketName);
     HttpResponse response = execute(Method.DELETE, BucketRegionCache.INSTANCE.region(bucketName),
                                     bucketName, objectName, null,
-                                    queryParamMap, null, null, 0);
+                                    queryParamMap, null, 0);
     response.body().close();
     return response;
   }
@@ -1043,7 +1053,7 @@ public final class MinioClient {
     updateRegionCache(bucketName);
     return execute(Method.POST, BucketRegionCache.INSTANCE.region(bucketName),
                    bucketName, objectName, headerMap, queryParamMap,
-                   null, data, 0);
+                   data, 0);
   }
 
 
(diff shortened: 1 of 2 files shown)


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1036,14 +1037,14 @@ private HttpResponse executeDelete(String bucketName, String objectName, Map<Str
    * @param data           HTTP request body data.
    */
   private HttpResponse executePost(String bucketName, String objectName, Map<String,String> headerMap,
-                                   Map<String,String> queryParamMap, Object data)
+                                   Map<String,String> queryParamMap, String contentType, Object data)
```
balamurugana: 1. As `headerMap` already carries `Content-Type`, we would need to use it.
2. You could remove `contentType` argument in `execute()` and get `contentType` from `headerMap` and pass it to `createRequest()` or `createRequest()` could get `contentType` from `headerMap`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1938,7 +1940,8 @@ public void makeBucket(String bucketName, String region)
       configString = config.toString();
     }
 
-    executePut(bucketName, null, headerMap, null, US_EAST_1, configString, 0);
+    execute(Method.PUT, US_EAST_1, bucketName, null, headerMap, null,
+            "application/octet-stream", configString, 0);
```
balamurugana: Avoid using `execute()` directly

api/src/main/java/io/minio/MinioClient.java
harshavardhana: The way it is written it is not possible we need to remove the MediaType behavior. Do you agree that is okay to do ?

api/src/main/java/io/minio/MinioClient.java
harshavardhana: This is on purpose for `makeBucket()`  knows what to send there is no automatic region handling etc. So i see this is correct.

api/src/main/java/io/minio/MinioClient.java
harshavardhana: This leads to a lot more involved, i would do a cleanup in a separate patch @balamurugana - because this patch at-least fixes the original problem of regression.

balamurugana: You can use this PR https://github.com/harshavardhana/minio-java/pull/1 which fixes these things.

api/src/main/java/io/minio/MinioClient.java
balamurugana: No. You can use another `executePut()` which accepts region

balamurugana:
> @balamurugana - because this patch at-least fixes the original problem of regression.

It looks like content-type support wasn't working earlier also.  That means its not a regression.  The actual fix we are doing now.

api/src/main/java/io/minio/MinioClient.java
balamurugana: We could use `execute()` directly, but these wrappers prevent errors at compile time. Currently response body is not used in any method using `executePut()`.  Closing at `executePut()` prevents file stream leaks else each method should properly `close()`

harshavardhana:
> It looks like content-type support wasn't working earlier also. That means its not a regression. The actual fix we are doing now

It did when minio-java was first released.

harshavardhana:
@balamurugana fixed all the changes..

api/src/main/java/io/minio/MinioClient.java
harshavardhana: What i am saying its confusing for newer patches. Nitish tried to use executePut for copyObject it didn't work and then had to use lower level execute() function - what i am saying is abstraction has a limit and it is not addressing all scenarios so it breaking. 

It is better for each caller to know if and when they need to close. Closing internally and not knowing  that it would be closed is hard to follow. For example executeGet and executePost don't close the body underneath - all abstracted wrappers for http methods should have same behavior otherwise we won't remember.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1989,6 +1997,57 @@ public void removeBucket(String bucketName)
...
+   * @param contentType File content type of the object, user supplied.
+   *
+   * @throws InvalidBucketNameException  upon invalid bucket name is given
+   * @throws NoResponseException         upon no response from server
+   * @throws IOException                 upon connection error
+   * @throws XmlPullParserException      upon parsing response xml
+   * @throws ErrorResponseException      upon unsuccessful execution
+   * @throws InternalException           upon internal library error
+   */
+  public void putObject(String bucketName, String objectName, String fileName, String contentType)
```
balamurugana: Make `public void putObject(String bucketName, String objectName, String fileName)` method to use this version of method to avoid code duplication.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1989,6 +1997,57 @@ public void removeBucket(String bucketName)
...
+    Path filePath = Paths.get(fileName);
+    if (!Files.isRegularFile(filePath)) {
+      throw new InvalidArgumentException("'" + fileName + "': not a regular file");
+    }
+
+    long size = Files.size(filePath);
+
+    RandomAccessFile file = new RandomAccessFile(filePath.toFile(), "r");
+    try {
+      putObject(bucketName, objectName, contentType, size, file);
```
balamurugana: If `contentType` is `null` we could probe the content type than `octet-stream`

```java
    String contentType = Files.probeContentType(filePath);
```

balamurugana:
> What i am saying its confusing for newer patches. Nitish tried to use executePut for copyObject it didn't work and then had to use lower level execute() function. Abstraction has a limit and it is not addressing all scenarios so it breaking.

`executePut()` should be fixed than using `execute()` directly.


> 
> It is better for each caller to know if and when they need to close. Closing internally and not knowing that it would be closed is hard to follow. For example executeGet and executePost don't close the body underneath - all abstracted wrappers for http methods should have same behavior otherwise we won't remember.

Till current usage of `executePut()` and `executePost()` no one is reading the body hence those methods close the body.  If there is a requirement of reading body and they are private methods we would need to fix it at the time.

api/src/main/java/io/minio/MinioClient.java
harshavardhana: Done

api/src/main/java/io/minio/MinioClient.java
harshavardhana: Done

harshavardhana:
> Till current usage of executePut() and executePost() no one is reading the body hence those methods close the body. If there is a requirement of reading body and they are private methods we would need to fix it at the time.

I don't see why you can't see the problem in current code.  The detail is harder to remember if a programmer looks at `executeGet()` and `executePost()`  they see that body is closed by the caller.  So it is natural one would expect such a behavior with `executePut()` .  Basically these `executeMethods()` should return the response and let the caller close the body so that they are all consistent in behavior.

harshavardhana:
> I don't see why you can't see the problem in current code. The detail is harder to remember if a programmer looks at executeGet() and executePost() they see that body is closed by the caller. So it is natural one would expect such a behavior with executePut() . Basically these executeMethods() should return the response and let the caller close the body so that they are all consistent in behavior.

This issue i will send a separate fix `makeBucket()` is still using the `executePut()` wrapper.

harshavardhana:
ping @balamurugana take a look please. We need to make a fix soon.

PR #524, opened 2017-02-10
API: Add new constructors with region parameter

vadmeste:
A developer can avoid a bucket location call to the server by
using this API and setting a region to any bucket.

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -34,6 +34,7 @@ import io.minio.errors.InvalidExpiresRangeException;
 import io.minio.errors.InvalidObjectPrefixException;
 import io.minio.errors.InvalidPortException;
 import io.minio.errors.NoResponseException;
+import io.minio.errors.RegionConflictException;
 import io.minio.http.HeaderParser;
 import io.minio.http.Method;
 import io.minio.http.Scheme;
@@ -160,6 +161,8 @@ public final class MinioClient {
   private String accessKey;
   // Secret key to sign all requests with
   private String secretKey;
+  // Region to sign all requests with
+  private String region;
 
   private String userAgent = DEFAULT_USER_AGENT;
 
@@ -187,10 +190,12 @@ public final class MinioClient {
    *
    * @see #MinioClient(URL url)
    * @see #MinioClient(String endpoint, String accessKey, String secretKey)
+   * @see #MinioClient(String endpoint, String accessKey, String secretKey, String region)
    * @see #MinioClient(URL url, String accessKey, String secretKey)
    * @see #MinioClient(String endpoint, int port, String accessKey, String secretKey)
    * @see #MinioClient(String endpoint, String accessKey, String secretKey, boolean secure)
    * @see #MinioClient(String endpoint, int port, String accessKey, String secretKey, boolean secure)
+   * @see #MinioClient(String endpoint, int port, String accessKey, String secretKey, String region, boolean secure)
    */
   public MinioClient(String endpoint) throws InvalidEndpointException, InvalidPortException {
     this(endpoint, 0, null, null);
@@ -207,10 +212,12 @@ public final class MinioClient {
    *
    * @see #MinioClient(String endpoint)
    * @see #MinioClient(String endpoint, String accessKey, String secretKey)
+   * @see #MinioClient(String endpoint, String accessKey, String secretKey, String region)
    * @see #MinioClient(URL url, String accessKey, String secretKey)
    * @see #MinioClient(String endpoint, int port, String accessKey, String secretKey)
    * @see #MinioClient(String endpoint, String accessKey, String secretKey, boolean secure)
    * @see #MinioClient(String endpoint, int port, String accessKey, String secretKey, boolean secure)
+   * @see #MinioClient(String endpoint, int port, String accessKey, String secretKey, String region, boolean secure)
    */
   public MinioClient(URL url) throws InvalidEndpointException, InvalidPortException {
     this(url.toString(), 0, null, null);
@@ -227,10 +234,12 @@ public final class MinioClient {
    * @see #MinioClient(String endpoint)
    * @see #MinioClient(URL url)
    * @see #MinioClient(String endpoint, String accessKey, String secretKey)
+   * @see #MinioClient(String endpoint, String accessKey, String secretKey, String region)
    * @see #MinioClient(URL url, String accessKey, String secretKey)
    * @see #MinioClient(String endpoint, int port, String accessKey, String secretKey)
    * @see #MinioClient(String endpoint, String accessKey, String secretKey, boolean secure)
    * @see #MinioClient(String endpoint, int port, String accessKey, String secretKey, boolean secure)
+   * @see #MinioClient(String endpoint, int port, String accessKey, String secretKey, String region, boolean secure)
    */
   public MinioClient(HttpUrl url) throws InvalidEndpointException, InvalidPortException {
     this(url.toString(), 0, null, null);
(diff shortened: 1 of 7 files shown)


balamurugana:
@vadmeste As all our SDKs use region cache, none of our SDKs provide such method by the design.  This design change needs to be reviewed by @abperiasamy

harshavardhana:
> @vadmeste As all our SDKs use region cache, none of our SDKs provide such method by the design. This design change needs to be reviewed by @abperiasamy

This we are going to do and decided to be added - since it is an important requirement.  It was decided that we take 'region' with the constructor but not as an `setRegion` API and default it to for all buckets if set. If not which is usually the default case we employ bucket-cache.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -497,6 +497,19 @@ public MinioClient(String endpoint, int port, String accessKey, String secretKey
...
+  /**
+   * Sets the region of a given bucket.
+   *
+   * </p><b>Example:</b><br>
+   * <pre>{@code minioClient.setRegion("my-bucket", "us-west-2")}</pre>
+   *
+   * @param bucketName   The bucket name.
+   * @param region       The region name.
+   */
+  public void setRegion(String bucketName, String region) {
```
harshavardhana: Take region as part of the constructor.

balamurugana:
> It was decided that we take 'region' with the constructor

If region is taken in the constructor, does it mean any S3 call uses passed region of any given bucket?

harshavardhana:
> If region is taken in the constructor, does it mean any S3 call uses passed region of any given bucket?

Yes that is correct - if a user has buckets in different regions then they are supposed to use this constructor. 

The reason to implement this is for applications which are generating only presignedURLs which do not wish to communicate to server to fetch the region.

balamurugana:
> Yes that is correct - if a user has buckets in different regions then they are supposed to use this constructor.

Do you mean
```java
MinioClient allregionMinioClient = new MinioClient(...); //allregionMinioClient fetches region for cache misses.
MinioClient uswest2regionMinioClient = new MinioClient(..., "us-west-2"); // all S3 calls goes to 'us-west-2'
MinioClient useast1regionMinioClient = new MinioClient(..., "us-east-1"); // all S3 calls goes to 'us-east-1'
```
?

harshavardhana:
> Do you mean
> 
> MinioClient allregionMinioClient = new MinioClient(...); //allregionMinioClient fetches region for cache misses.
> MinioClient uswest2regionMinioClient = new MinioClient(..., "us-west-2"); // all S3 calls goes to 'us-west-2'
> MinioClient useast1regionMinioClient = new MinioClient(..., "us-east-1"); // all S3 calls goes to 'us-east-1'

Yes @balamurugana - and following this across all libraries.

balamurugana:
> Yes @balamurugana - and following this across all libraries.

Got it.

vadmeste:
I added setRegion() API because we have too many constructors, so this way, setting region will be possible for any constructor the developer chooses to use.

harshavardhana:
> I added setRegion() API because we have too many constructors, so this way, setting region will be possible for any constructor the developer chooses to use.

Actually with function overloading we shouldn't be worried about that in Java. Constructor is the right place because having an active caching and some other calls actively replacing your region - can create a confusing situation.

Having it defaulted allows for lesser mutation and more predictable outcome. It is also natural that region is provided in the constructor for any S3 API sdk.

vadmeste:
I made new changes, ready for review

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -421,12 +475,55 @@ public MinioClient(String endpoint, String accessKey, String secretKey, boolean
...
    * @see #MinioClient(String endpoint, String accessKey, String secretKey)
+   * @see #MinioClient(String endpoint, String accessKey, String secretKey, String region)
    * @see #MinioClient(URL url, String accessKey, String secretKey)
    * @see #MinioClient(String endpoint, int port, String accessKey, String secretKey)
    * @see #MinioClient(String endpoint, String accessKey, String secretKey, boolean secure)
+   * @see #MinioClient(String endpoint, int port, String accessKey, String secretKey, String region, boolean secure)
    */
   public MinioClient(String endpoint, int port, String accessKey, String secretKey, boolean secure)
     throws InvalidEndpointException, InvalidPortException {
+    this(endpoint, port, accessKey, secretKey, "", secure);
```
balamurugana: you could pass `null` instead of `""`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -932,43 +1030,49 @@ private void updateRegionCache(String bucketName)
...
+            region = "eu-west-1";
+          } else {
+            region = location;
+          }
         }
       }
 
       // Add the new location.
-      BucketRegionCache.INSTANCE.add(bucketName, region);
+      BucketRegionCache.INSTANCE.set(bucketName, region);
```
balamurugana: I thought of BucketRegionCache.region() could return passed region for non-empty bucket name.  I am OK with this approach too.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -932,43 +1030,49 @@ private void updateRegionCache(String bucketName)
...
-      XmlPullParser xpp = xmlPullParserFactory.newPullParser();
-      String location = null;
-
-      xpp.setInput(response.body().charStream());
-      while (xpp.getEventType() != xpp.END_DOCUMENT) {
-        if (xpp.getEventType() ==  xpp.START_TAG && xpp.getName() == "LocationConstraint") {
+      // Compute bucket region but always pick this.region if the latter is set,
+      // otherwise, call the bucket location API.
+      String region;
+      if (!this.region.equals("")) {
```
balamurugana: We would need to check null or empty check.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -421,12 +475,55 @@ public MinioClient(String endpoint, String accessKey, String secretKey, boolean
...
+   * @see #MinioClient(String endpoint)
+   * @see #MinioClient(URL url)
+   * @see #MinioClient(String endpoint, String accessKey, String secretKey)
+   * @see #MinioClient(String endpoint, String accessKey, String secretKey, String region)
+   * @see #MinioClient(URL url, String accessKey, String secretKey)
+   * @see #MinioClient(String endpoint, int port, String accessKey, String secretKey)
+   * @see #MinioClient(String endpoint, String accessKey, String secretKey, boolean secure)
+   * @see #MinioClient(String endpoint, int port, String accessKey, String secretKey, String region, boolean secure)
+   */
+  public MinioClient(String endpoint, int port, String accessKey, String secretKey, String region, boolean secure)
```
balamurugana: Please add functional test for this constructor.

vadmeste:
@balamurugana, I addressed your comments, besides I added RegionConflictException when makeBucket() is invoked with incompatible region parameter

build.gradle
```diff
@@ -269,7 +269,13 @@ project(':functional') {
             ext.secretKey = rootProject.properties['secretKey']
         }
 
-        args = [ext.endPoint, ext.accessKey, ext.secretKey]
+        ext.region = 'us-east-1'
+        if (project.properties.containsKey('region')) {
+            ext.secretKey = rootProject.properties['region']
```
harshavardhana: shouldn't this be `ext.region' ?

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -932,43 +1031,49 @@ private void updateRegionCache(String bucketName)
...
-      String location = null;
-
-      xpp.setInput(response.body().charStream());
-      while (xpp.getEventType() != xpp.END_DOCUMENT) {
-        if (xpp.getEventType() ==  xpp.START_TAG && xpp.getName() == "LocationConstraint") {
+      // Compute bucket region but always pick this.region if the latter is set,
+      // otherwise, call the bucket location API.
+      String region;
+      if (this.region != null && !this.region.equals("")) {
+        region = this.region;
```
harshavardhana: This code looks incorrect - bucketCache is still used when the region is set. we should refactor it so that when region is given we never ever use the bucket cache instead the return value is directly from `this.region` 

Region cache should only be used when `this.region` is not set - this would indicate the future code is not added on top to have a mutating state underneath.

vadmeste: Done

harshavardhana:
This looks good to me @balamurugana can you review?.

functional/FunctionalTest.java
```diff
@@ -1075,85 +1076,120 @@ public static void copyObject_test7() throws Exception {
...
+    try {
+      client = new MinioClient(endpoint, accessKey, secretKey);
+      // Enable trace for debugging.
+      // client.traceOn(System.out);
+      FunctionalTest.runTests();
 
-      threadedPutObject();
+      // Run fast test with region parameter passed to the constructor
+      client = new MinioClient(endpoint, accessKey, secretKey, region);
+      FunctionalTest.runFastTests();
```
balamurugana: I think you can name `runTestsWithRegion()` or something to convey these tests with region than `runFastTests()`.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -968,10 +1066,26 @@ private void updateRegionCache(String bucketName)
...
-      BucketRegionCache.INSTANCE.add(bucketName, region);
+      BucketRegionCache.INSTANCE.set(bucketName, region);
     }
   }
 
+  /**
+   * Computes region of a given bucket name. If set, this.region is considered. Otherwise,
+   * resort to the server location API.
+   */
+  private String bucketToRegion(String bucketName) throws InvalidBucketNameException, NoSuchAlgorithmException,
```
balamurugana: I think `getRegion()` is more appropriate than `bucketToRegion()` which looks like a conversion method.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2118,9 +2212,17 @@ public void makeBucket(String bucketName)
...
-           InternalException {
+    throws InvalidBucketNameException, RegionConflictException, NoSuchAlgorithmException, InsufficientDataException,
+                IOException, InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
+                InternalException {
+    // If region param is not provided, set it with the one provided by constructor
+    if (region == null) {
+      region = this.region;
+    }
+    // If constructor already sets a region, check if it is equal to region param if provided
+    if (this.region != null && !this.region.equals(region)) {
```
balamurugana: `region != null` check is not needed.

functional/FunctionalTest.java
vadmeste: > I think you can name runTestsWithRegion() or something to convey these tests with region than runFastTests().

The purpose of runFastTest() is to test some basic S3 operations and not only with region. We can use it later to test multiple combinations of constructors parameters.

api/src/main/java/io/minio/MinioClient.java
vadmeste: > I think getRegion() is more appropriate than bucketToRegion() which looks like a conversion method.

:+1:

api/src/main/java/io/minio/MinioClient.java
vadmeste: > region != null check is not needed.

Do you mean this.region ?

harshavardhana: it is the `region` as it can be `null`

vadmeste: > it is the region as it can be null

Actually, String.equals() can work with null parameter

api/src/main/java/io/minio/PostPolicy.java
```diff
@@ -167,14 +167,6 @@ public String bucketName() {
   }
 
   /**
-   * Returns form data of this post policy.
-   */
-  public Map<String,String> formData(String accessKey, String secretKey)
```
vadmeste: > Removal makes backward incompatible. Is it expected?

You are right, PostPolicy is really used by the developer. However, this function doesn't seem to be working properly before, as it uses "us-east-1" by default unless another API was called before and updated the region cache with the correct value.

balamurugana:
> Do you mean this.region ?

Oops! I didn't notice `this.` properly.  Sorry for the confusion.

api/src/main/java/io/minio/PostPolicy.java
harshavardhana: This is okay to break since it was used incorrectly, better for the users.

balamurugana: > However, this function doesn't seem to be working properly before, as it uses "us-east-1" by default unless another API was called before and updated the region cache with the correct value.

I couldn't able to understand this.  When `MinioClient.presignedPostPolicy(PostPolicy)` method is called, it does `updateRegionCache(policy.bucketName());` then calls `policy.formData(this.accessKey, this.secretKey);` where bucket region cache is used.

Do you say this is not working?

balamurugana: I am OK about backward incompatible.  You should increment major version in this PR.

harshavardhana: Actually wait @vadmeste you should keep the previous code its harmless and there is no reason to remove it.

balamurugana: Reason of `public formData()` is to access it in `class MinioClient`.  Majority of use cases do not touch/use this method unless some special use case like @MastaP in https://github.com/minio/minio-java/pull/510.

As you use `public Map<String,String> formData(String accessKey, String secretKey, String region)`, you could mark `public Map<String,String> formData(String accessKey, String secretKey)` as deprecated and we will remove it in next major update.

harshavardhana: This looks like a good idea. We can use the decorator here.

vadmeste:
formData(String accessKey, String secretKey) is deprecated now.

PR #534, opened 2017-03-08
Fix scanner object leaks close them appropriately.

harshavardhana:
Currently scanner was never closed which would leak fds.

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -63,7 +63,6 @@ import org.xmlpull.v1.XmlPullParserException;
 import org.xmlpull.v1.XmlPullParserFactory;
 
 import java.io.BufferedInputStream;
-import java.io.EOFException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -89,7 +88,6 @@ import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Scanner;
 import java.util.concurrent.TimeUnit;
-import java.util.logging.Level;
 import java.util.logging.Logger;
 
 /**
@@ -629,8 +627,7 @@ public final class MinioClient {
    * Validates if given objectPrefix is valid.
    */
   private void checkObjectPrefix(String prefix) throws InvalidObjectPrefixException {
-    // TODO(nl5887): what to do with wildcards in objectPrefix?
-    //
+    // TODO(nl5887): what to do with wild-cards in objectPrefix?
     if (prefix.length() > 1024) {
       throw new InvalidObjectPrefixException(prefix, "Object prefix cannot be greater than 1024 characters.");
     }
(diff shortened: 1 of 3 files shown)


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -967,6 +965,7 @@ private HttpResponse execute(Method method, String region, String bucketName, St
       } finally {
         response.body().close();
       }
+      scanner.close();
```
balamurugana: Adding `scanner.close();` in `finally` is good enough than moving scanner out of `try` block.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2938,20 +2937,18 @@ private void completeMultipart(String bucketName, String objectName, String uplo
...
-    } catch (EOFException e) {
-      // Getting EOF exception is not an error.
-      // Just log it.
-      LOGGER.log(Level.WARNING, "EOF exception occured: " + e);
-    }  finally {
+    } finally {
       response.body().close();
     }
-
+    scanner.close();
```
balamurugana: same as above

api/src/main/java/io/minio/MinioClient.java
harshavardhana: Done.

harshavardhana:
ping @balamurugana

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -950,11 +947,12 @@ private HttpResponse execute(Method method, String region, String bucketName, St
 
     // HEAD returns no body, and fails on parseXml
     if (!method.equals(Method.HEAD)) {
+      Scanner scanner = new Scanner(response.body().charStream());
+      scanner.useDelimiter("\\A");
```
balamurugana: You missed scanner to be moved inside `try` block.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2938,18 +2937,16 @@ private void completeMultipart(String bucketName, String objectName, String uplo
 
     // Fixing issue https://github.com/minio/minio-java/issues/391
     String bodyContent = "";
+    Scanner scanner = new Scanner(response.body().charStream());
```
balamurugana: same as above

api/src/main/java/io/minio/MinioClient.java
harshavardhana: It won't work @balamurugana there is a compilation error..

harshavardhana: so what is done right now is the correct style..

api/src/main/java/io/minio/MinioClient.java
harshavardhana: Read the above comment.. try doesn't work.

harshavardhana:
ping @balamurugana

PR #537, opened 2017-03-11
fix: close scanner always including when exception occurred

harshavardhana:


harshavardhana:
> fix: close scanner always including when exception occurred

Done!

PR #547, opened 2017-04-09
api: Set correct (un)modified condition header

### LINKED ISSUE #546: Copy unmodified condition fails with AWS S3
```java
    CopyConditions copyConditions = new CopyConditions();
    DateTime dateRepresentation = new DateTime(2015, Calendar.MAY, 3, 10, 10);
    copyConditions.setUnmodified(dateRepresentation);
    client.copyObject(bucketName, filename, destBucketName, copyConditions);
```

when tested against AWS S3, copy is performed though it shouldn't.

The reason is that X-Amz-Copy-Source-If-Unmodified-Since is not sent with the right date time format

vadmeste:
This PR sets the correct date time format for headers
x-amz-copy-if-modified-since and x-amz-copy-source-if-unmodified-since.
In consequence, modified/unmodified copy preconditions will just work.

Fixes #546

### DIFF
--- a/api/src/main/java/io/minio/DateFormat.java
+++ b/api/src/main/java/io/minio/DateFormat.java
@@ -38,7 +38,7 @@ public class DateFormat {
       DateTimeFormat.forPattern("yyyyMMdd").withZoneUTC().withLocale(Locale.US);
 
   public static final DateTimeFormatter HTTP_HEADER_DATE_FORMAT =
-      DateTimeFormat.forPattern("EEE',' dd MMM yyyy HH':'mm':'ss zzz").withZoneUTC().withLocale(Locale.US);
+      DateTimeFormat.forPattern("EEE',' dd MMM yyyy HH':'mm':'ss 'GMT'").withZoneUTC().withLocale(Locale.US);
 
   private DateFormat() {}
 }
--- a/functional/FunctionalTest.java
+++ b/functional/FunctionalTest.java
@@ -1055,11 +1055,13 @@ public class FunctionalTest {
       client.copyObject(bucketName, filename, destBucketName, copyConditions);
     } catch (ErrorResponseException e) {
       // File should not be copied as object was modified after date set in copyConditions.
-      ignore();
+      if (!e.errorResponse().code().equals("PreconditionFailed")) {
+        throw e;
+      }
     }
 
     client.removeObject(bucketName, filename);
-    client.removeObject(destBucketName, filename);
+    // Destination bucket is expected to be empty, otherwise it will trigger an exception.
     client.removeBucket(destBucketName);
   }
 


api/src/main/java/io/minio/DateFormat.java
```diff
@@ -38,7 +38,7 @@
       DateTimeFormat.forPattern("yyyyMMdd").withZoneUTC().withLocale(Locale.US);
 
   public static final DateTimeFormatter HTTP_HEADER_DATE_FORMAT =
-      DateTimeFormat.forPattern("EEE',' dd MMM yyyy HH':'mm':'ss zzz").withZoneUTC().withLocale(Locale.US);
+      DateTimeFormat.forPattern("EEE',' dd MMM yyyy HH':'mm':'ss 'GMT'").withZoneUTC().withLocale(Locale.US);
```
balamurugana: What is the time sent with/without this change?  I suspect it should be the same.

functional/FunctionalTest.java
```diff
@@ -1074,11 +1074,17 @@ public static void copyObject_test7() throws Exception {
...
       // File should not be copied as object was modified after date set in copyConditions.
-      ignore();
+      expectedException = e;
+    }
+
+    if (expectedException == null || !expectedException.errorResponse().code().equals("PreconditionFailed")) {
+      throw new Exception("[FAILED] copyObject_test7(): copying should fail with unsatisified precondition.");
     }
 
     client.removeObject(bucketName, fileName);
```
balamurugana: As there is no object copied, you should remove `client.removeObject(secondBucketName, fileName);`. below. You will see the failure.

functional/FunctionalTest.java
```diff
@@ -1074,11 +1074,17 @@ public static void copyObject_test7() throws Exception {
...
+
     try {
       client.copyObject(bucketName, fileName, secondBucketName, copyConditions);
     } catch (ErrorResponseException e) {
       // File should not be copied as object was modified after date set in copyConditions.
-      ignore();
+      expectedException = e;
+    }
+
+    if (expectedException == null || !expectedException.errorResponse().code().equals("PreconditionFailed")) {
```
balamurugana: 1. You would need to check in `catch()` not outside.
2. `expectedException` is not necessary.

api/src/main/java/io/minio/DateFormat.java
harshavardhana: It is not the same 

Without 
```
x-amz-copy-source-if-unmodified-since: Fri, 03 Apr 2015 17:10:00 +00:00
``` 

With
```
x-amz-copy-source-if-unmodified-since: Fri, 03 Apr 2015 17:10:00 GMT
```

functional/FunctionalTest.java
harshavardhana: You won't see any failure because removeObject never returns any error, with or without object exists or not.

api/src/main/java/io/minio/DateFormat.java
balamurugana: If so, we need to create new formatter for this than modifying response header date formatter.

functional/FunctionalTest.java
balamurugana: That's known.

As per the test, there is no object copied and you should not remove.  This is the bug in the test which causes no failure.

api/src/main/java/io/minio/DateFormat.java
balamurugana: `'GMT' ` change is good to go. You would need

1. modify commit log
2. have a separate PR for functional test fix.

functional/FunctionalTest.java
vadmeste: Done.

functional/FunctionalTest.java
vadmeste: > You would need to check in catch() not outside.
> expectedException is not necessary.

Checking outside catch() is better since it can cover the case when copyObject() doesn't trigger an exception as intended.

balamurugana: Its not java style. Can you tell me a case where its useful?
Catching and raising its own exception is unwanted thing in functional test.  The scope of the test is up to what expected. If anything happens beyond that is an error.

vadmeste: copyObject_test7() is intended to test if copyObject() is always raising an exception when we pass an unsatisfying `x-amz-copy-source-if-unmodified-since` condition. We have a bug that this PR is actually fixing but this test wasn't actually detecting it before.

balamurugana: Yes. Due to the bug in the test, it didn't catch the actual problem.  However how is it related to moving exception handling outside of `catch()`?

vadmeste: The test wasn't detecting the bug because it is doing the check inside catch(), copyObject wasn't triggering an exception as it should have done.

balamurugana: The actual bug is having `client.removeObject(destBucketName, filename);` in `copyObject_test7()`.  If you remove that code, there is no special exception handling required.

vadmeste: > The actual bug is having client.removeObject(destBucketName, filename); in copyObject_test7(). If you remove that code, there is no special exception handling required.

Do you mean that the change in this test function is useless and the only needed thing to do is to remove `client.removeObject(destBucketName, filename);` ?

functional/FunctionalTest.java
```diff
@@ -1055,11 +1055,13 @@ public static void copyObject_test7() throws Exception {
...
       // File should not be copied as object was modified after date set in copyConditions.
-      ignore();
+      if (!e.errorResponse().code().equals("PreconditionFailed")) {
+        throw e;
+      }
     }
 
     client.removeObject(bucketName, filename);
-    client.removeObject(destBucketName, filename);
+    // Destination bucket is expected to be empty, otherwise it will trigger an exception.
```
balamurugana: :+1:

vadmeste:
https://github.com/minio/minio-java/pull/548 needs to be merged first

vadmeste:
Ready and tested against Minio and AWS S3

PR #550, opened 2017-04-11
examples: Implement console progress bar for PutObject

harshavardhana:


build.gradle
```diff
@@ -239,6 +239,7 @@ project(':examples') {
 
 project(':functional') {
     dependencies {
+        compile "me.tongfei:progressbar:0.5.3"
```
balamurugana: Why is this dependency in `functional`?

harshavardhana: oops

harshavardhana: Fixed

PR #551, opened 2017-04-11
api: add new bucket notification APIs.

### LINKED ISSUE #437: api: Implement SetBucketNotification, GetBucketNotification, RemoveAllBucketNotification

balamurugana:
Below methods are added
* public NotificationConfiguration getBucketNotification(String bucketName)
* public void setBucketNotification(String bucketName,
                  NotificationConfiguration notificationConfiguration)
* public void removeAllBucketNotification(String bucketName)

Fixes #437

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -57,6 +57,7 @@ import io.minio.messages.ListPartsResult;
 import io.minio.messages.Part;
 import io.minio.messages.Prefix;
 import io.minio.messages.Upload;
+import io.minio.messages.NotificationConfiguration;
 import io.minio.org.apache.commons.validator.routines.InetAddressValidator;
 import io.minio.policy.PolicyType;
 import io.minio.policy.BucketPolicy;
@@ -2987,6 +2988,97 @@ public final class MinioClient {
   }
 
 
+  /**
+   * Get bucket notification configuration
+   *
+   * @param bucketName   Bucket name.
+   *
+   * </p><b>Example:</b><br>
+   * <pre>{@code NotificationConfiguration notificationConfig = minioClient.getBucketNotification("my-bucketname");
+   * }</pre>
+   *
+   * @throws InvalidBucketNameException  upon invalid bucket name is given
+   * @throws NoResponseException         upon no response from server
+   * @throws IOException                 upon connection error
+   * @throws XmlPullParserException      upon parsing response xml
+   * @throws ErrorResponseException      upon unsuccessful execution
+   * @throws InternalException           upon internal library error
+   *
+   */
+  public NotificationConfiguration getBucketNotification(String bucketName)
+    throws InvalidBucketNameException, InvalidObjectPrefixException, NoSuchAlgorithmException,
+           InsufficientDataException, IOException, InvalidKeyException, NoResponseException,
+           XmlPullParserException, ErrorResponseException, InternalException {
+    Map<String,String> queryParamMap = new HashMap<>();
+    queryParamMap.put("notification", "");
+  
+    HttpResponse response = executeGet(bucketName, null, null, queryParamMap);
+    NotificationConfiguration result = new NotificationConfiguration();
+    try {
+      result.parseXml(response.body().charStream());
+    } finally {
+      response.body().close();
+    }
+
+    return result;
+  }
+  
+  
+  /**
+   * Set bucket notification configuration
+   *
+   * @param bucketName   Bucket name.
+   * @param notificationConfiguration   Notification configuration to be set.
+   *
+   * </p><b>Example:</b><br>
+   * <pre>{@code minioClient.setBucketNotification("my-bucketname", notificationConfiguration);
+   * }</pre>
+   *
+   * @throws InvalidBucketNameException  upon invalid bucket name is given
+   * @throws NoResponseException         upon no response from server
+   * @throws IOException                 upon connection error
+   * @throws XmlPullParserException      upon parsing response xml
+   * @throws ErrorResponseException      upon unsuccessful execution
+   * @throws InternalException           upon internal library error
+   *
+   */
+  public void setBucketNotification(String bucketName, NotificationConfiguration notificationConfiguration)
+    throws InvalidBucketNameException, InvalidObjectPrefixException, NoSuchAlgorithmException,
+           InsufficientDataException, IOException, InvalidKeyException, NoResponseException,
+           XmlPullParserException, ErrorResponseException, InternalException {
+    Map<String,String> queryParamMap = new HashMap<>();
+    queryParamMap.put("notification", "");
+    HttpResponse response = executePut(bucketName, null, null, queryParamMap, notificationConfiguration.toString(), 0);
+    response.body().close();
+  }
+  
+  
+  /**
+   * Remove all bucket notification.
+   *
+   * @param bucketName   Bucket name.
+   *
+   * </p><b>Example:</b><br>
+   * <pre>{@code minioClient.removeAllBucketNotification("my-bucketname");
+   * }</pre>
+   *
+   * @throws InvalidBucketNameException  upon invalid bucket name is given
+   * @throws NoResponseException         upon no response from server
+   * @throws IOException                 upon connection error
+   * @throws XmlPullParserException      upon parsing response xml
+   * @throws ErrorResponseException      upon unsuccessful execution
+   * @throws InternalException           upon internal library error
+   *
+   */
+  public void removeAllBucketNotification(String bucketName)
+    throws InvalidBucketNameException, InvalidObjectPrefixException, NoSuchAlgorithmException,
+           InsufficientDataException, IOException, InvalidKeyException, NoResponseException,
+           XmlPullParserException, ErrorResponseException, InternalException {
+    NotificationConfiguration notificationConfiguration = new NotificationConfiguration();
+    setBucketNotification(bucketName, notificationConfiguration);
+  }
+  
+  
   /**
    * Returns next part if exists.
    */
--- a/api/src/main/java/io/minio/messages/CloudFunctionConfiguration.java
+++ b/api/src/main/java/io/minio/messages/CloudFunctionConfiguration.java
@@ -1,0 +1,111 @@
+/*
+ * Minio Java SDK for Amazon S3 Compatible Cloud Storage, (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.minio.messages;
+
+import java.util.LinkedList;
+import java.util.List;
+
+import org.xmlpull.v1.XmlPullParserException;
+
+import com.google.api.client.util.Key;
+
+import io.minio.errors.InvalidArgumentException;
+
+
+/**
+ * Helper class to parse Amazon AWS S3 response XML containing cloud function configuration.
+ */
+public class CloudFunctionConfiguration extends XmlEntity {
+  @Key("Id")
+  private String id;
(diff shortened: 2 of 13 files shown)


api/src/main/java/io/minio/messages/CloudFunctionConfiguration.java
```diff
@@ -0,0 +1,109 @@
...
+  private String id;
+  @Key("CloudFunction")
+  private String cloudFunction;
+  @Key("Event")
+  private List<String> events = new LinkedList<>();
+  @Key("Filter")
+  private Filter filter;
+
+
+  public CloudFunctionConfiguration() throws XmlPullParserException {
```
vadmeste: A constructor without any parameters can trigger an exception ?

vadmeste:
@balamurugana, can you add unit tests ?

examples/SetBucketNotification.java
```diff
@@ -0,0 +1,68 @@
...
+
+      // Get current notification configuration.
+      NotificationConfiguration notificationConfiguration = minioClient.getBucketNotification("my-bucketname");
+
+      // Add a new topic configuration.
+      List<TopicConfiguration> topicConfigurationList = notificationConfiguration.topicConfigurationList();
+      TopicConfiguration topicConfiguration = new TopicConfiguration();
+      topicConfiguration.setTopic("arn:aws:sns:us-west-2:444455556666:sns-topic-xyz");
+      List<String> eventList = new LinkedList<>();
+      eventList.add("s3:ObjectCreated:Put");
```
vadmeste: `s3:ObjectCreated:Put`, I think it would be better if we will have some constants that represent all possible events and we put them in java doc. That way, it would be easier for the developer to see what it can add as events.

examples/SetBucketNotification.java
```diff
@@ -0,0 +1,68 @@
...
+      eventList.add("s3:ObjectCreated:Copy");
+      topicConfiguration.setEvents(eventList);
```
vadmeste: One thing missing in the example is setting filters, can you add them? Besides, in minio-go, we have an API for adding a prefix filter and a suffix filter. We probably want to do the same.

api/src/main/java/io/minio/messages/TopicConfiguration.java
```diff
@@ -0,0 +1,91 @@
...
+   */
+  public Filter filter() {
+    return filter;
+  }
+
+
+  /**
+   * Sets filter.
+   */
+  public void setFilter(Filter filter) {
```
vadmeste: For all configuration types (cloudfunction, queue and topic) we can always set at most one suffix filter and one prefix.

api/src/main/java/io/minio/messages/CloudFunctionConfiguration.java
balamurugana: You missed `super();`

api/src/main/java/io/minio/messages/TopicConfiguration.java
balamurugana: Can you point me to such restriction in the spec?

examples/SetBucketNotification.java
balamurugana: Done

examples/SetBucketNotification.java
balamurugana: I will check how to do with java way.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2631,6 +2632,73 @@ public void setBucketPolicy(String bucketName, String objectPrefix, PolicyType p
...
 
   /**
+   * Get bucket notification configuration
+   *
+   * @param bucketName   Bucket name.
+   *
+   * </p><b>Example:</b><br>
+   * <pre>{@code NotificationConfiguration notificationConfig = minioClient.getBucketNotification("my-bucketname");
+   * }</pre>
+   */
```
NitishT: You can add the Exceptions thrown list here.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2631,6 +2632,73 @@ public void setBucketPolicy(String bucketName, String objectPrefix, PolicyType p
...
+  /**
+   * Set bucket notification configuration
+   *
+   * @param bucketName   Bucket name.
+   * @param notificationConfiguration   Notification configuration to be set.
+   *
+   * </p><b>Example:</b><br>
+   * <pre>{@code minioClient.setBucketNotification("my-bucketname", notificationConfiguration);
+   * }</pre>
+   */
```
NitishT: You can add the Exceptions thrown list here.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2631,6 +2632,73 @@ public void setBucketPolicy(String bucketName, String objectPrefix, PolicyType p
...
+  
+  /**
+   * Remove all bucket notification.
+   *
+   * @param bucketName   Bucket name.
+   *
+   * </p><b>Example:</b><br>
+   * <pre>{@code minioClient.removeAllBucketNotification("my-bucketname");
+   * }</pre>
+   */
```
NitishT: You can add the Exceptions thrown list here.

api/src/main/java/io/minio/messages/FilterRule.java
```diff
@@ -0,0 +1,70 @@
...
+package io.minio.messages;
+
+import org.xmlpull.v1.XmlPullParserException;
+
+import com.google.api.client.util.Key;
+
+
+/**
+ * Helper class to parse Amazon AWS S3 response XML containing filter rule.
+ */
```
NitishT: Instead of AWS S3 response, can we say Minio server response here? It is kind of known that we are S3 compatible, but IMHO we should use Minio in comments etc.

api/src/main/java/io/minio/messages/NotificationConfiguration.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.List;
+
+import org.xmlpull.v1.XmlPullParserException;
+
+import com.google.api.client.util.Key;
+
+
+/**
+ * Helper class to parse Amazon AWS S3 response XML containing notification configuration.
+ */
```
NitishT: Same as above

api/src/main/java/io/minio/messages/TopicConfiguration.java
vadmeste: Not exactly, it is something that is probably not written in the spec, I remember seeing that, and minio-go already works that way.

api/src/main/java/io/minio/MinioClient.java

api/src/main/java/io/minio/MinioClient.java

api/src/main/java/io/minio/MinioClient.java

api/src/main/java/io/minio/messages/FilterRule.java
balamurugana: We would need to convey the specification ie Amazon AWS S3 is the specification.

api/src/main/java/io/minio/messages/TopicConfiguration.java
balamurugana: If its not in the spec, then we shouldn't need to do that.

vadmeste: @balamurugana, S3 will just return an error if you send more than one suffix filter for example.

balamurugana: @vadmeste Unfortunately I am not able to understand. 

I read `Object key name prefix or suffix identifying one or more objects to which the filtering rule applies. Maximum prefix length can be up to 1,024 characters. Overlapping prefixes and suffixes are not supported. ` in the spec.

What I understood is below is valid
```
      <Filter>
          <S3Key>
              <FilterRule>
                  <Name>prefix</Name>
                  <Value>images/</Value>
              </FilterRule>
              <FilterRule>
                  <Name>prefix</Name>
                  <Value>photos/</Value>
              </FilterRule>
              <FilterRule>
                  <Name>suffix</Name>
                  <Value>jpg</Value>
              </FilterRule>
          </S3Key>
     </Filter>
```

Overlapping prefix like `photos` and `photos-2010`  are error.

Anyhow I will try how Amazon S3 behaves and add accordingly.

vadmeste: > Anyhow I will try how Amazon S3 behaves and add accordingly.

Yes, it would be better to test again, Thanks. I am relying on my memory and minio-go code (https://github.com/minio/minio-go/blob/master/bucket-notification.go#L107)

api/src/main/java/io/minio/messages/QueueConfiguration.java
```diff
@@ -0,0 +1,112 @@
...
+   */
+  public void setQueue(String queue) {
+    this.queue = queue;
+  }
+
+
+  /**
+   * Returns events.
+   */
+  public List<String> events() {
```
vadmeste: We probably want to make events() returns List<EventType> as setEvents()

api/src/main/java/io/minio/messages/TopicConfiguration.java
```diff
@@ -0,0 +1,94 @@
...
+   */
+  public void setTopic(String topic) {
+    this.topic = topic;
+  }
+
+
+  /**
+   * Returns events.
+   */
+  public List<String> events() {
```
vadmeste: events() needs to return List<EventType> ?

api/src/main/java/io/minio/messages/QueueConfiguration.java
balamurugana: Done

api/src/main/java/io/minio/messages/TopicConfiguration.java
balamurugana: I think you wanted to the same in your comment above?

api/src/main/java/io/minio/messages/FilterRule.java
NitishT: Okay, that is valid too.

api/src/main/java/io/minio/messages/TopicConfiguration.java
vadmeste: > I think you wanted to the same in your comment above?
yes exactly sorry for the typo.

api/src/main/java/io/minio/messages/TopicConfiguration.java
balamurugana: @vadmeste  I have confirmed below XML is invalid as per Amazon AWS S3 server.
```xml
      <Filter>
          <S3Key>
              <FilterRule>
                  <Name>prefix</Name>
                  <Value>images/</Value>
              </FilterRule>
              <FilterRule>
                  <Name>prefix</Name>
                  <Value>photos/</Value>
              </FilterRule>
              <FilterRule>
                  <Name>suffix</Name>
                  <Value>jpg</Value>
              </FilterRule>
          </S3Key>
     </Filter>
```

Accordingly to the behavior I am changing helper methods.

balamurugana:
> @balamurugana, can you add unit tests ? or functional tests

Functional tests can not be added because of Topic/Queue/CloudFunction are very specific to a target.  those target/bucket have to be permanently there in Amazon AWS S3.

I defer unit test WRT minio-java which does not work/help the API behavior.

As notification system is out of S3 storage, functional/unit tests can't be done.

harshavardhana:
> This PR is not tested yet. Till then

This PR is ready @balamurugana to be merged?

balamurugana:
@harshavardhana 

> This PR is ready @balamurugana to be merged?

Please look my previous comments here about functional/unit tests.

harshavardhana:
> Please look my previous comments here about functional/unit tests.

IT won't be a problem @balamurugana , we can add a test case like how we have done in minio-go but only run it when requested ?

balamurugana:
The functional tests are run for validating minio-java APIs.  But the problem is notification APIs is it requires pre-configured topic/queue/cloudfunction targets in prior which is completely external entity.

We can't assume and hard-code these values which may lead to respective functional tests failure.  Currently I don't know solution for that.

However examples available for those APIs can be modified/run for further validation by hand.

harshavardhana:
> We can't assume and hard-code these values which may lead to respective functional tests failure. Currently I don't know solution for that.

Not saying we should hard-code minio-go implements a way to test it, but requires you to enable some ENVs if those ENVs are not found the tests are skipped.  We can do something similar in minio-java . So that when one needs to test notification part it can be tested as well. 

It is non conventional but serves the purpose IMHO.

balamurugana:
All functional tests pass in `s3.amazonaws.com`

PR #552, opened 2017-04-13
api: add list object version 2 support and set to default

### LINKED ISSUE #413: api: Implement support for ListObjectsV2. supporting new S3 list version 2 type.

balamurugana:
Now ListObjects() API uses ReST API List Objects Version 2
(http://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html)
as a default.

Added new method
listObjects(String bucketName, String prefix, boolean recursive,
            boolean useVersion1)

which can be used for ReST API List Objects Version 1 by passing true
to useVersion1 flag.

Fixes #413

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -51,6 +51,7 @@ import io.minio.messages.InitiateMultipartUploadResult;
 import io.minio.messages.Item;
 import io.minio.messages.ListAllMyBucketsResult;
 import io.minio.messages.ListBucketResult;
+import io.minio.messages.ListBucketResultV1;
 import io.minio.messages.ListMultipartUploadsResult;
 import io.minio.messages.ListPartsResult;
 import io.minio.messages.Part;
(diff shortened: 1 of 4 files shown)


balamurugana:
All functional tests pass in `play.minio.io` and `s3.amazonaws.com`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1931,13 +1932,73 @@ public void removeObject(String bucketName, String objectName)
...
+    return listObjects(bucketName, prefix, recursive, false);
+  }
+
+
+  /**
+   * Lists object information as {@code Iterable<Result><Item>>} in given bucket, prefix, recursive flag and S3 API
+   * version to use.
+   *
+   * </p><b>Example:</b><br>
+   * <pre>{@code Iterable<Result<Item>> myObjects = minioClient.listObjects("my-bucketname");
```
NitishT: This example is not updated.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1931,13 +1932,73 @@ public void removeObject(String bucketName, String objectName)
...
+
+    return listObjectsV2(bucketName, prefix, recursive);
+  }
+
+  
+  /**
+   * Lists object information as {@code Iterable<Result><Item>>} in given bucket, prefix and recursive flag.
+   *
+   * </p><b>Example:</b><br>
+   * <pre>{@code Iterable<Result<Item>> myObjects = minioClient.listObjects("my-bucketname");
```
NitishT: Example not updated.

api/src/main/java/io/minio/MinioClient.java

api/src/main/java/io/minio/MinioClient.java
balamurugana: Removed

PR #553, opened 2017-04-15
api: add new API to delete multiple objects.

### LINKED ISSUE #467: Implement RemoveObjects API to remove objects in batches
Refer minio-go for further API details.

balamurugana:
This patch adds new API

public Iterable<Result<DeleteError>> removeObject(final String bucketName,
    final Iterable<String> objectNames)

which uses "Delete Multiple Objects" ReST API
(http://docs.aws.amazon.com/AmazonS3/latest/API/multiobjectdeleteapi.html).

Fixes #467

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -42,6 +42,10 @@ import io.minio.messages.Bucket;
 import io.minio.messages.CompleteMultipartUpload;
 import io.minio.messages.CopyObjectResult;
 import io.minio.messages.CreateBucketConfiguration;
+import io.minio.messages.DeleteError;
+import io.minio.messages.DeleteObject;
+import io.minio.messages.DeleteRequest;
+import io.minio.messages.DeleteResult;
 import io.minio.messages.ErrorResponse;
 import io.minio.messages.InitiateMultipartUploadResult;
 import io.minio.messages.Item;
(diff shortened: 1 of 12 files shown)


balamurugana:
The PR succeeds in `play.minio.io` and `s3.amazonaws.com`

api/src/main/java/io/minio/messages/ErrorResponse.java
```diff
@@ -33,21 +33,21 @@
 @SuppressWarnings("unused")
 public class ErrorResponse extends XmlEntity {
   @Key("Code")
-  private String code;
+  protected String code;
```
harshavardhana: why is this changed in this PR? any specific reason?

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1884,6 +1888,156 @@ public void removeObject(String bucketName, String objectName)
...
+
+    return errorList;
+  }
+
+
+  /**
+   * Removes multiple objects from a bucket.
+   *
+   * </p><b>Example:</b><br>
+   * <pre>{@code minioClient.removeObject("my-bucketname", "my-objectname"); }</pre>
```
harshavardhana: This might need an update to reflect to the new API changes.

functional/FunctionalTest.java
```diff
@@ -435,6 +435,26 @@ public static void removeObject_test() throws Exception {
...
+    String[] filenames = new String[4];
+    for (int i = 0; i < 3; i++) {
+      String filename = createFile(1 * MB);
+      client.putObject(bucketName, filename, filename);
+      Files.delete(Paths.get(filename));
+      filenames[i] = filename;
+    }
+    filenames[3] = "nonexistent-object";
+
+    for (Result<?> r : client.removeObject(bucketName, Arrays.asList(filenames))) {
```
harshavardhana: Is this the right style of usage of this newer API? i thought we are supposed to send an iterator down not an Array list.

api/src/main/java/io/minio/messages/ErrorResponse.java
balamurugana: Yes. Check `DeleteError.java`

api/src/main/java/io/minio/MinioClient.java

functional/FunctionalTest.java
balamurugana: The API requires `Iterable<String>`.  `Arrays,asList()` creates an `iterable` object.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1884,6 +1888,157 @@ public void removeObject(String bucketName, String objectName)
...
+           InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
+           InternalException {
+    Map<String,String> queryParamMap = new HashMap<>();
+    queryParamMap.put("delete", "");
+
+    DeleteRequest request = new DeleteRequest(objectList);
+    HttpResponse response = executePost(bucketName, null, null, queryParamMap, request);
+
+    String bodyContent = "";
+    Scanner scanner = new Scanner(response.body().charStream());
```
vadmeste: I am not sure what's the purpose of this scanner? can you add a comment ?

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1884,6 +1888,157 @@ public void removeObject(String bucketName, String objectName)
   }
 
 
+  private List<DeleteError> removeObject(String bucketName, List<DeleteObject> objectList)
```
vadmeste: Do you mean removeObjects() ? since it takes a list of delete object requests.

api/src/main/java/io/minio/MinioClient.java

api/src/main/java/io/minio/MinioClient.java
balamurugana: Nope.  I wanted to keep a overloaded `removeObject()`

api/src/main/java/io/minio/MinioClient.java
vadmeste: > Its already commented below.

yeah I didn't expect all these lines are just to read the body, there is not an easier way? ``// Use scanner to read the entire body`` will be a better comment probably.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1884,6 +1888,156 @@ public void removeObject(String bucketName, String objectName)
...
+      }
+    } finally {
+      response.body().close();
+      scanner.close();
+    }
+
+    List<DeleteError> errorList = null;
+
+    bodyContent = bodyContent.trim();
+    DeleteError error = new DeleteError(new StringReader(bodyContent));
```
vadmeste: DeleteError as the S3 only returns errors ?

api/src/main/java/io/minio/MinioClient.java
balamurugana: Done.

Many ways to do that http://stackoverflow.com/questions/309424/read-convert-an-inputstream-to-a-string

api/src/main/java/io/minio/MinioClient.java
balamurugana: `DeleteError` is multi object removal error in various cases as per the spec.

vadmeste: > DeleteError is multi object removal error in various cases as per the spec.

yes thanks, the spec is little confusing, especially how it is returning 200 OK also for errors

vadmeste: Can you also add a comment for this, to say that we need sometimes to parse the body twice since the xml tree structure is not the same in some circumstances.

PR #560, opened 2017-04-21
Fix isValidEndpoint function to match spec

donatello:
Existing code had a bug. For example "minio-.example.com" would be permitted by the existing code, but this is clearly not allowed by the spec.

Spec - https://en.wikipedia.org/wiki/Hostname#Restrictions_on_valid_host_names


donatello:
> Perhaps bring a set of Unit tests for this? @donatello
Done @harshavardhana

Issue #536, opened 2017-03-10
Support encryption for GetObject() and PutObject()


NitishT:
Minio server supports the header required. The SDK needs to support client side encryption and decryption.

deekoder:
Is there a bug on the server that this one depends on.

harshavardhana:
> Is there a bug on the server that this one depends on.

There is no server side changed needed for this. Its purely a client enhancement issue.

NitishT:
ETA 15 April

PR #542, opened 2017-04-06
[WIP] Implement client side encryption

NitishT:
- Added getObject() to fetch encrypted objects

- Added putObject() to put encrypted objects

PR #561, opened 2017-04-23
api: client side encryption

NitishT:
- Supports RSA and AES encryption on client side. 
- Took care of review comments in https://github.com/minio/minio-java/pull/542
- Fixes https://github.com/minio/minio-java/issues/536

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,9 @@ The full API Reference is available here.
 
 #### Full Examples: Object Operations
 * [PutObject.java](https://github.com/minio/minio-java/tree/master/examples/PutObject.java)
+* [PutObjectEncrypted.java](https://github.com/minio/minio-java/tree/master/examples/PutObjectEncrypted.java)
 * [GetObject.Java](https://github.com/minio/minio-java/tree/master/examples/GetObject.java)
+* [GetObjectEncrypted.Java](https://github.com/minio/minio-java/tree/master/examples/GetObjectEncrypted.java)
 * [GetPartialObject.java](https://github.com/minio/minio-java/tree/master/examples/GetPartialObject.java)
 * [RemoveObject.java](https://github.com/minio/minio-java/tree/master/examples/RemoveObject.java)
 * [RemoveObjects.java](https://github.com/minio/minio-java/tree/master/examples/RemoveObjects.java)
--- a/api/src/main/java/io/minio/Crypto.java
+++ b/api/src/main/java/io/minio/Crypto.java
@@ -1,0 +1,183 @@
+/*
+ * Minio Java SDK for Amazon S3 Compatible Cloud Storage,
+ * (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.minio;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.security.InvalidAlgorithmParameterException;
+import java.security.InvalidKeyException;
+import java.security.Key;
+import java.security.NoSuchAlgorithmException;
+
+import javax.crypto.BadPaddingException;
+import javax.crypto.Cipher;
+import javax.crypto.CipherInputStream;
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.NoSuchPaddingException;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.IvParameterSpec;
+
+class Crypto {
+
+  /**
+   * Decrypts an encrypted input stream using secret key and IV.
+   * 
+   * @param encryptedInputStream
+   *          Encrypted {@link InputStream}
+   * @param contentKey
+   *          {@link SecretKey} for decryption
+   * @param iv
+   *          IV for decryption
+   * 
+   * @return {@link CipherInputStream}
+   * 
+   * @throws InvalidKeyException
+   *           upon invalid keys
+   * @throws InvalidAlgorithmParameterException
+   *           upon invalid algorithm
+   * @throws NoSuchAlgorithmException
+   *           upon requested algorithm was not found
+   * @throws NoSuchPaddingException
+   *           upon incorrect padding
+   * @throws IllegalBlockSizeException
+   *           upon incorrect block size
+   * @throws BadPaddingException
+   *           upon wrong padding
+   */
+  public static CipherInputStream decrypt(InputStream encryptedInputStream, SecretKey contentKey,
+      String transformation, byte[] iv) throws InvalidKeyException, NoSuchAlgorithmException, NoSuchPaddingException,
+      IllegalBlockSizeException, BadPaddingException, InvalidAlgorithmParameterException {
+
+    // Get the cipher
+    Cipher cipher = Cipher.getInstance(transformation);
+
+    // init cipher with mode and data encryption key
+    cipher.init(Cipher.DECRYPT_MODE, contentKey, new IvParameterSpec(iv));
+
+    // create cipherinputstream with encrypted stream and new initialized cipher
+    return new CipherInputStream(encryptedInputStream, cipher);
+  }
+
+  /**
+   * Decrypts cipher text byte array using RSA Private key. Returns plain data byte array.
+   * 
+   * @param cipherText
+   *          cipher text data to be decrypted
+   * 
+   * @return byte array with the plain text data
+   * 
+   * @throws BadPaddingException
+   *           upon wrong padding
+   * @throws IllegalBlockSizeException
+   *           upon incorrect block size
+   * @throws NoSuchAlgorithmException
+   *           upon requested algorithm was not found during signature calculation
+   * @throws BadPaddingException
+   *           upon wrong padding
+   * @throws NoSuchPaddingException
+   *           upon incorrect padding
+   */
+  public static byte[] decrypt(byte[] cipherText, Key key, String transformation) throws NoSuchAlgorithmException,
+      NoSuchPaddingException, InvalidKeyException, IllegalBlockSizeException, BadPaddingException {
+
+    // Create encryption cipher
+    Cipher cipher = Cipher.getInstance(transformation);
+
+    // init encryption cipher with mode and public key
+    cipher.init(Cipher.DECRYPT_MODE, key);
+
+    // decrypt cipher text data
+    return cipher.doFinal(cipherText);
+  }
+
+  /**
+   * Encrypts an plain input stream using secret key and IV.
+   * 
+   * @param plainInputStream
+   *          Plain {@link InputStream}
+   * @param contentKey
+   *          {@link SecretKey} for encryption
+   * @param iv
+   *          IV for encryption
+   * 
+   * @return {@link CipherInputStream}
+   * 
+   * @throws InvalidKeyException
+   *           upon invalid keys
+   * @throws InvalidAlgorithmParameterException
+   *           upon invalid algorithm
+   * @throws NoSuchAlgorithmException
+   *           upon requested algorithm was not found
+   * @throws NoSuchPaddingException
+   *           upon incorrect padding
+   * @throws IllegalBlockSizeException
+   *           upon incorrect block size
+   * @throws BadPaddingException
+   *           upon wrong padding
+   * @throws IOException
+   *           upon IO error
+   */
+  public static CipherInputStream encrypt(InputStream plainInputStream, SecretKey contentKey,
+      String transformation, byte[] iv) throws InvalidKeyException, NoSuchAlgorithmException, NoSuchPaddingException,
+      IllegalBlockSizeException, BadPaddingException, InvalidAlgorithmParameterException, IOException {
+
+    // Get the cipher
+    Cipher cipher = Cipher.getInstance(transformation);
+
+    // init cipher with mode and data encryption key
+    cipher.init(Cipher.ENCRYPT_MODE, contentKey, new IvParameterSpec(iv));
+
+    // create cipherinputstream with plain stream and new initialized cipher
+    return new CipherInputStream(plainInputStream, cipher);
+  }
+
+  /**
+   * Encrypts plain text byte array using RSA Public key. Returns encrypted data byte array.
+   * 
+   * @param plainText
+   *          plain text data to be encrypted
+   * 
+   * @return byte array with the encrypted data
+   * 
+   * @throws BadPaddingException
+   *           upon wrong padding
+   * @throws IllegalBlockSizeException
+   *           upon incorrect block size
+   * 
+   * @throws NoSuchPaddingException
+   *           In case of wrong padding scheme
+   * @throws NoSuchAlgorithmException
+   *           In case of wrong algorithm
+   * @throws InvalidKeyException
+   *           In case of wrong key
+   * 
+   */
+  public static byte[] encrypt(byte[] plainText, Key key, String transformation)
+      throws IllegalBlockSizeException, BadPaddingException, NoSuchAlgorithmException, NoSuchPaddingException,
+      InvalidKeyException {
+
+    // Create encryption cipher
+    Cipher cipher = Cipher.getInstance(transformation);
+
+    // init encryption cipher with mode and public key
+    cipher.init(Cipher.ENCRYPT_MODE, key);
+
+    // encrypt plain text data and return
+    return cipher.doFinal(plainText);
+  }
+}
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -17,6 +17,7 @@
 
 package io.minio;
 
+import com.google.common.io.BaseEncoding;
 import com.google.common.io.ByteStreams;
 import com.squareup.okhttp.HttpUrl;
 import com.squareup.okhttp.MediaType;
@@ -24,11 +25,13 @@ import com.squareup.okhttp.OkHttpClient;
 import com.squareup.okhttp.Request;
 import com.squareup.okhttp.RequestBody;
 import com.squareup.okhttp.Response;
+
 import io.minio.errors.ErrorResponseException;
 import io.minio.errors.InsufficientDataException;
 import io.minio.errors.InternalException;
 import io.minio.errors.InvalidArgumentException;
 import io.minio.errors.InvalidBucketNameException;
+import io.minio.errors.InvalidEncryptionMetadataException;
 import io.minio.errors.InvalidEndpointException;
 import io.minio.errors.InvalidExpiresRangeException;
 import io.minio.errors.InvalidObjectPrefixException;
@@ -63,6 +66,7 @@ import io.minio.policy.PolicyType;
 import io.minio.policy.BucketPolicy;
 import okio.BufferedSink;
 import okio.Okio;
+
 import org.joda.time.DateTime;
 import org.xmlpull.v1.XmlPullParser;
 import org.xmlpull.v1.XmlPullParserException;
@@ -84,8 +88,12 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
 import java.nio.file.StandardOpenOption;
+import java.security.InvalidAlgorithmParameterException;
 import java.security.InvalidKeyException;
+import java.security.Key;
+import java.security.KeyPair;
 import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedList;
@@ -95,6 +103,15 @@ import java.util.NoSuchElementException;
 import java.util.Scanner;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Logger;
+
+import javax.crypto.BadPaddingException;
+import javax.crypto.Cipher;
+import javax.crypto.CipherInputStream;
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.KeyGenerator;
+import javax.crypto.NoSuchPaddingException;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.SecretKeySpec;
 
 /**
  * <p>
@@ -1268,8 +1285,9 @@ public final class MinioClient {
            InternalException {
     HttpResponse response = executeHead(bucketName, objectName);
     ResponseHeader header = response.header();
-    return new ObjectStat(bucketName, objectName, header.lastModified(), header.contentLength(),
-                          header.etag(), header.contentType());
+    ObjectStat objectStat = new ObjectStat(bucketName, objectName, header);
+
+    return objectStat;
   }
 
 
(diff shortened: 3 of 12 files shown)


README.md
```diff
@@ -134,7 +134,11 @@ The full API Reference is available here.
 
 #### Full Examples: Object Operations
 * [PutObject.java](https://github.com/minio/minio-java/tree/master/examples/PutObject.java)
+* [PutObjectEncryptedAes.java](https://github.com/minio/minio-java/tree/master/examples/PutObjectEncryptedAes.java)
```
vadmeste: Aes is an acronym, let's capitalize it entirely.

README.md
```diff
@@ -134,7 +134,11 @@ The full API Reference is available here.
...
+* [PutObjectEncryptedRsa.java](https://github.com/minio/minio-java/tree/master/examples/PutObjectEncryptedRsa.java)
```
vadmeste: Rsa is an acronym, let's capitalize it entirely.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2774,27 +2901,198 @@ public void putObject(String bucketName, String objectName, InputStream stream,
...
+    // Prepare data to be put to the object header
+    String encDataKey = new String(Base64.encodeBase64(encryptedDataKey), StandardCharsets.UTF_8);
+    String ivString = new String(Base64.encodeBase64(iv), StandardCharsets.UTF_8);
+
+    // Set the headermap with encryption related metadata
+    Map<String, String> headerMap = new HashMap<>();
+    headerMap.put("x-amz-meta-x-amz-key", encDataKey);
+    headerMap.put("x-amz-meta-x-amz-iv", ivString);
+    // matdesc is unused
+    headerMap.put("x-amz-meta-x-amz-matdesc", "");
```
vadmeste: I always see this value as '{}', let's do the same so we give other clients what they expect.

api/src/main/java/io/minio/encryption/EncryptionMaterials.java
```diff
@@ -0,0 +1,214 @@
...
+   *          plain text byte array
+   * 
+   * @return encrypted byte array
+   * 
+   * @throws IllegalBlockSizeException
+   *           upon incorrect block size
+   * @throws BadPaddingException
+   *           upon wrong padding
+   */
+  public byte[] encryptContentKeys(byte[] plainTextKey) throws IllegalBlockSizeException, BadPaddingException {
```
vadmeste: encryptContentKey (without 's')

api/src/main/java/io/minio/encryption/EncryptionMaterials.java
```diff
@@ -0,0 +1,214 @@
...
+   * @throws NoSuchAlgorithmException
+   *           upon requested algorithm was not found
+   * @throws NoSuchPaddingException
+   *           upon incorrect padding
+   * @throws IllegalBlockSizeException
+   *           upon incorrect block size
+   * @throws BadPaddingException
+   *           upon wrong padding
+   */
+  public byte[] decryptContentKeys(byte[] encryptedDataKey) throws IllegalBlockSizeException, BadPaddingException,
```
vadmeste: decryptContentKey (without 's')

api/src/main/java/io/minio/encryption/EncryptionMaterials.java
```diff
@@ -0,0 +1,214 @@
...
+
+import javax.crypto.BadPaddingException;
+import javax.crypto.Cipher;
+import javax.crypto.CipherInputStream;
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.NoSuchPaddingException;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.IvParameterSpec;
+
+public class EncryptionMaterials {
```
vadmeste: Can you add one abstraction level so EncryptionMaterials can support more than one encryption algorithm ?

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1260,8 +1277,17 @@ public ObjectStat statObject(String bucketName, String objectName)
            InternalException {
     HttpResponse response = executeHead(bucketName, objectName);
     ResponseHeader header = response.header();
-    return new ObjectStat(bucketName, objectName, header.lastModified(), header.contentLength(),
-                          header.etag(), header.contentType());
+    ObjectStat objectStat = null;
+
+    // Return object stat with key data if header has key and iv
+    if ((header.xamzMetaKey() != null) && (header.xamzMetaIv() != null)) {
```
vadmeste: Can't you merge 'if' with 'else' here ? It looks we can only support ObjectStat(bucket, object, headers)

README.md
NitishT: Checkstyle complains if I do that, as `allowedAbbreviationLength` is set to 1 (https://github.com/minio/minio-java/blob/master/checkstyle.xml#L157). If @harshavardhana @balamurugana agree, we can change this to 3.

README.md
NitishT: Checkstyle complains if I do that, as `allowedAbbreviationLength` is set to 1 (https://github.com/minio/minio-java/blob/master/checkstyle.xml#L157). If @harshavardhana @balamurugana agree, we can change this to 3.

api/src/main/java/io/minio/MinioClient.java
NitishT: I had it like that, but @balamurugana has a valid comment about it breaking backwards compatibility - https://github.com/minio/minio-java/pull/542#discussion_r110333060

README.md
balamurugana: The style is copied from google.  Not sure whether we need to touch these settings.

harshavardhana: You can just keep it PutEncryptedObject no need to mention `AES` and `RSA` separately, perhaps AES is the example and RSA commented out?

api/src/test/java/io/minio/MinioClientTest.java
```diff
@@ -253,6 +270,273 @@ public void testGetObject()
...
+    // Create data key encryption cipher
+    Cipher keyEncryptionCipher = Cipher.getInstance("AES/ECB/PKCS5Padding");
+
+    // init cipher with mode and master key
+    keyEncryptionCipher.init(Cipher.ENCRYPT_MODE, symMasterKey);
+    // Encrypt the data key
+    byte[] encryptedDataKey = keyEncryptionCipher.doFinal(plainDataKey);
+
+    String encDataKey = new String(Base64.encodeBase64(encryptedDataKey), StandardCharsets.UTF_8);
+    String ivString = new String(Base64.encodeBase64(iv), StandardCharsets.UTF_8);
```
balamurugana: `Digest.java` has some methods related to base64 hashing? Is it possible to use/enhance them?

api/src/main/java/io/minio/MinioClient.java
balamurugana: As `public class ObjectStat`, we would need to keep backward compatible constructors (You can deprecate those constructors too and we will remove in next major release), but usage side within `minio-java`, we could use better version.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1509,6 +1535,107 @@ public void getObject(String bucketName, String objectName, String fileName)
...
+   * @throws BadPaddingException
+   *           upon wrong padding
+   * @throws IllegalBlockSizeException
+   *           upon incorrect block size
+   * @throws NoSuchPaddingException
+   *           upon incorrect padding
+   * @throws InvalidAlgorithmParameterException
+   *           upon incorrect algorithm
+   */
+  public CipherInputStream getObject(String bucketName, String objectName, EncryptionMaterials encryptionMaterials)
```
balamurugana: If `CipherInputStream` is interface compatible with `InputStream`, we would need to return `InputStream`.  This way there is no difference to API user.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1509,6 +1535,107 @@ public void getObject(String bucketName, String objectName, String fileName)
...
+
+    if (encryptionMaterials == null) {
+      throw new InvalidArgumentException("empty encryption properties not allowed");
+    }
+
+    // get encrypted object metadata and verify
+    ObjectStat stat = statObject(bucketName, objectName);
+
+    // check if encryption metadata is not present
+    if ((stat.contentKey() == null) || (stat.encryptionIv() == null)) {
```
balamurugana: If object meta data are null and `encryptionMeterials` is also null, you could fall back to regular get object.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1509,6 +1535,107 @@ public void getObject(String bucketName, String objectName, String fileName)
...
+    InputStream stream = getObject(bucketName, objectName);
+
+    // Get the encrypted key from response metadata
+    byte[] encryptedDataKey = Base64.decodeBase64(stat.contentKey().getBytes(StandardCharsets.UTF_8));
+
+    // Get the iv from the response metadata
+    byte[] iv = Base64.decodeBase64(stat.encryptionIv().getBytes(StandardCharsets.UTF_8));
+
+    // Decrypt the encrypted data key using master key
+    byte[] plainDataKey = encryptionMaterials.decryptContentKeys(encryptedDataKey);
```
balamurugana: `dataKey` is better than `plainDataKey`.  If S3 Spec says it as `content key`, I prefer the variable names as `encryptedContentKey` and `contentKey` and we would name that method in `EncryptionMaterials` accordingly.

api/src/main/java/io/minio/encryption/EncryptionMaterials.java
NitishT: done.

api/src/main/java/io/minio/encryption/EncryptionMaterials.java
NitishT: done

api/src/main/java/io/minio/MinioClient.java
NitishT: done

README.md
NitishT: > You can just keep it PutEncryptedObject no need to mention AES and RSA separately, perhaps AES is the example and RSA commented out?

Okay, I will do this

api/src/main/java/io/minio/encryption/EncryptionMaterials.java
NitishT: @vadmeste I am not sure if I understood this, can you pls explain this further?

api/src/main/java/io/minio/MinioClient.java
balamurugana: You could use this `getObject()` as primary method like you used `putObject()` below.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2774,27 +2901,198 @@ public void putObject(String bucketName, String objectName, InputStream stream,
...
+   * @see #putObject(String bucketName, String objectName, InputStream stream, long size, String contentType)
+   *
+   */
+  public void putObject(String bucketName, String objectName, InputStream stream, long size, String contentType,
+      EncryptionMaterials encryptionMaterials) throws InvalidBucketNameException, NoSuchAlgorithmException,
+      InsufficientDataException, IOException, InvalidKeyException, NoResponseException, XmlPullParserException,
+      ErrorResponseException, InternalException, InvalidArgumentException, NoSuchPaddingException,
+      IllegalBlockSizeException, BadPaddingException, InvalidAlgorithmParameterException {
+
+    if (encryptionMaterials == null) {
```
balamurugana: You can fall back to regular `putObject()`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2774,27 +2901,198 @@ public void putObject(String bucketName, String objectName, InputStream stream,
...
+      throw new InvalidArgumentException("empty encryption properties not allowed");
+    }
+
+    // Generate symmetric 256 bit AES key for data encryption
```
balamurugana: Is 256 bit AES key default in S3 Spec?

api/src/main/java/io/minio/encryption/EncryptionMaterials.java
vadmeste: Here, EncryptionMaterials uses CBC + PAD5 as an encryption algorithm. To be able to support more algorithms, EncryptionMaterials needs to be an abstract class, and we can write a new class called EncryptionMaterialsCBC for example which does the real encryption.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2774,27 +2901,198 @@ public void putObject(String bucketName, String objectName, InputStream stream,
...
+    KeyGenerator symKeyGenerator = KeyGenerator.getInstance("AES");
+    symKeyGenerator.init(256);
+    SecretKey dataEncryptionKey = symKeyGenerator.generateKey();
+
+    // Generate an iv to be used for data encryption
+    SecureRandom ivSeed = new SecureRandom();
```
balamurugana: you could move this statement below for logical grouping.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2774,27 +2901,198 @@ public void putObject(String bucketName, String objectName, InputStream stream,
...
+    byte[] iv = new byte[16];
+    ivSeed.nextBytes(iv);
+
+    // Get CipherStream after encryption of input stream
+    CipherInputStream cipherInputStream = encryptionMaterials.encryptInputStream(stream, dataEncryptionKey, iv);
```
balamurugana: If `CipherInputStream` is interface compatible with `InputStream`, you could pass it to a new `putObject(String bucketName, String objectName, InputStream stream, long size, Map<String, String> customHeaders)` as primary method for all put object versions.

README.md
NitishT: done as discussed

README.md
NitishT: done as discussed above

api/src/main/java/io/minio/encryption/MasterKey.java
```diff
@@ -0,0 +1,33 @@
...
+import javax.crypto.BadPaddingException;
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.NoSuchPaddingException;
+
+public abstract class MasterKey {
+
+  protected abstract byte[] encrypt(byte[] plainText) throws IllegalBlockSizeException, BadPaddingException;
+
+  protected abstract byte[] decrypt(byte[] cipherText) throws NoSuchAlgorithmException, NoSuchPaddingException,
+      InvalidKeyException, IllegalBlockSizeException, BadPaddingException;
```
balamurugana: I would prefer below in `Crypto.java`
```java
public static byte[] encrypt(byte[] plainTextKeys, KeyPair key) throws IllegalBlockSizeException, BadPaddingException;
public static byte[] encrypt(byte[] plainTextKeys, SecretKey key) throws IllegalBlockSizeException, BadPaddingException;

public static byte[] decrypt(byte[] cipherTextKeys, KeyPair key) throws NoSuchAlgorithmException, NoSuchPaddingException,
      InvalidKeyException, IllegalBlockSizeException, BadPaddingException;
public static byte[] decrypt(byte[] cipherTextKeys, SecretKey key) throws NoSuchAlgorithmException, NoSuchPaddingException,
      InvalidKeyException, IllegalBlockSizeException, BadPaddingException;

public InputStream decryptInputStream(InputStream encryptedInputStream, KeyPair key, String contentKey, byte[] iv)
      throws InvalidKeyException, NoSuchAlgorithmException, NoSuchPaddingException, IllegalBlockSizeException,
      BadPaddingException, InvalidAlgorithmParameterException;
public InputStream decryptInputStream(InputStream encryptedInputStream, SecretKey key, String contentKey, byte[] iv)
      throws InvalidKeyException, NoSuchAlgorithmException, NoSuchPaddingException, IllegalBlockSizeException,
      BadPaddingException, InvalidAlgorithmParameterException;

public InputStream encryptInputStream(InputStream plainInputStream, KeyPair key, String contentKey, byte[] iv)
      throws InvalidKeyException, NoSuchAlgorithmException, NoSuchPaddingException, IllegalBlockSizeException,
      BadPaddingException, InvalidAlgorithmParameterException;
public InputStream encryptInputStream(InputStream plainInputStream, SecretKey key, String contentKey, byte[] iv)
      throws InvalidKeyException, NoSuchAlgorithmException, NoSuchPaddingException, IllegalBlockSizeException,
      BadPaddingException, InvalidAlgorithmParameterException;
```

This would be plain/simple/straightforward methods without duplicate logic than interface/inheritance.

api/src/main/java/io/minio/encryption/EncryptionMaterials.java
harshavardhana: For example @NitishT here - https://github.com/minio/minio-go/tree/master/pkg/encrypt

NitishT: Okay, got it. Thanks @harshavardhana @vadmeste

api/src/main/java/io/minio/MinioClient.java
NitishT: I dont think falling back to regular `getObject` is a good option here. If an application needs plain object it should call corresponding `getObject` API. 

If we fall back to default `getObject` in cases where `encryptionMaterials` is null due to some error, we'll be silently discarding the error and returning wrong object. That is not correct.

api/src/main/java/io/minio/MinioClient.java
NitishT: As explained above, falling back to regular `putObject` can be confusing. I think it is best to throw error.

api/src/main/java/io/minio/MinioClient.java
NitishT: S3 spec calls it `data encryption key` or `data key` [(ref)](http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingClientSideEncryption.html). I can rename the vars to `dataKey` and `encryptedDataKey`

api/src/test/java/io/minio/MinioClientTest.java
NitishT: `Digest.java` has methods related to SHA/MD5 hashing, which are not reqd here. But as `Digest.java` uses `com.google.common.io.BaseEncoding` as the encoding library, I will use the same library for encoding here.

api/src/main/java/io/minio/MinioClient.java
NitishT: Spec doesn't specifically mention the size of randomly generated data encryption key. It just says `one-time-use symmetric key (also known as a data encryption key or data key).` [(ref)](http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingClientSideEncryption.html). 

AES 256 bit key is one of the strongest symmetric keys available, also Minio-go uses 256 bit symmetric data key.  So, 256 bit key is used.

api/src/main/java/io/minio/MinioClient.java
NitishT: Agreed, done.

api/src/main/java/io/minio/MinioClient.java
NitishT: Not sure about primary method meaning here. Could you pls explain bit more?

api/src/main/java/io/minio/MinioClient.java
NitishT: done.

api/src/main/java/io/minio/MinioClient.java
NitishT: This can be taken up as separate PR, IMO this is a general improvement, not related to client encryption per se.

api/src/main/java/io/minio/MinioClient.java
balamurugana: Its like 
```java
public void getObject(String bucketName, String objectName, String fileName);
public InputStream getObject(String bucketName, String objectName);
public InputStream getObject(String bucketName, String objectName, long offset);
public InputStream getObject(String bucketName, String objectName, long offset, Long length);
```

These methods uses primarily one method which is `public InputStream getObject(String bucketName, String objectName, long offset, Long length);`, rest of other methods are just wrappers.

WRT encrypt/decrypt methods, we would need below
```java
public InputStream getObject(String bucketName, String objectName, SecretKey key);
public InputStream getObject(String bucketName, String objectName, SecretKey key, Cipher cipher);
public InputStream getObject(String bucketName, String objectName, KeyPair key);
public InputStream getObject(String bucketName, String objectName, KeyPair key, Cipher cipher);
```

I think its hard to implement below
```java
public InputStream getObject(String bucketName, String objectName, long offset, Long length,
                             SecretKey key, Cipher cipher);
public InputStream getObject(String bucketName, String objectName, long offset, Long length,
                             KeyPair key, Cipher cipher);
```

Similarly for `putObject()`
```java
public void putObject(String bucketName, String objectName, String fileName, String contentType, SecretKey key,
                      Cipher cipher);
public void putObject(String bucketName, String objectName, String fileName, String contentType, KeyPair key,
                      Cipher cipher);
public void putObject(String bucketName, String objectName, InputStream stream, long size, String contentType,
                      Map<String, String> headerMap, SecretKey key, Cipher cipher);
public void putObject(String bucketName, String objectName, InputStream stream, long size, String contentType,
                      Map<String, String> headerMap, KeyPair key, Cipher cipher);
```

api/src/main/java/io/minio/MinioClient.java
balamurugana: How about providing an optional parameters for `encryptionKey` and `IV` (optional parameters are done by method overloading)? This way user has more control than a fixed algo.

api/src/main/java/io/minio/MinioClient.java
NitishT: Yes, `CipherStream` is a part of this PR only, but the changes you are proposing will be needed in all the `put object versions`. I meant this is not specific to client side encryption.

api/src/main/java/io/minio/MinioClient.java
NitishT: Changed the return type to `InputStream` in getObject.

api/src/main/java/io/minio/encryption/EncryptionMaterials.java
NitishT: This is done @vadmeste please take a look.

api/src/main/java/io/minio/encryption/MasterKey.java
NitishT: Working on this, will get back with updated code.

api/src/main/java/io/minio/MinioClient.java
balamurugana: `minio-go` may have lack of support or limitation to do so.  But its clean to do in java with method overloading.  Its not a stopper I guess.

api/src/main/java/io/minio/ObjectStat.java
```diff
@@ -29,7 +29,34 @@
...
+  private String contentKey;
+  // Encryption IV available in the object header
+  private String encryptionIv;
+  // Encryption material description available in the object header
+  private String matDesc;
 
+  /**
+   * Creates ObjectStat with given bucket name, object name, and available response header information.
+   */
+  public ObjectStat(String bucketName, String name, ResponseHeader header) {
```
vadmeste: Does this class really need to be public?

harshavardhana: It is better to be public.. since it is returned back in statObject()

NitishT:
@balamurugana changed as per your comments, ready for another review round.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2776,27 +2998,333 @@ public void putObject(String bucketName, String objectName, InputStream stream,
...
+    // Create byte array stream to hold encrypted stream
+    ByteArrayOutputStream encryptedStream = new ByteArrayOutputStream();
+
+    // Read bytes from cipher stream
+    int numRead = 0;
+    byte[] readBytes = new byte[8192];
+    while ((numRead = cipherInputStream.read(readBytes)) != -1) {
+      // write to byte array stream
+      encryptedStream.write(readBytes, 0, numRead);
+    }
```
balamurugana: I have few questions here
* Why do we need intermittent `ByteArrayOutputStream/ByteArrayInputStream` as `cipherInputStream` is already an `InputStream`?
* We encrypt each 8192 bytes from the input stream (note that `numRead` might not be guaranteed to have 8192) and stores to object.  What will happen if `getObject()` reads less than 8192 bytes or higher than 8192 bytes to decrypt?
* Does encryption increase/decrease total size of input stream? In short, does `encryptedStream` contains same amount of data as passed input stream?
* Assume the input stream is having GiBs of data.  This code holds same amount of memory for `encryptedStream`, we should use `PipedInputStream/PipedOutputStream`  or our own input/output stream to handle encryption on the fly.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2776,27 +2998,333 @@ public void putObject(String bucketName, String objectName, InputStream stream,
...
+    // Prepare data to be put to the object header
+    String encDataKey = BaseEncoding.base64().encode(encryptedDataKey);
+    String ivString = BaseEncoding.base64().encode(iv);
+    
+    // Set the headermap with encryption related metadata
+    Map<String, String> headerMap = new HashMap<>();
+    headerMap.put("x-amz-meta-x-amz-key", encDataKey);
+    headerMap.put("x-amz-meta-x-amz-iv", ivString);
+    // matdesc is unused
+    headerMap.put("x-amz-meta-x-amz-matdesc", "{}");
```
harshavardhana: It's aws-sdk behavior kept as empty json.  @vadmeste can fill why is it purposefully needed.

vadmeste: > It's aws-sdk behavior kept as empty json. @vadmeste can fill why is it purposefully needed.

x-amz-meta-x-amz-matdesc is used with another crypo algorithm ( AES/GCM/NoPadding). aws-java sdk works with current crypt algorithm if it doesn't find x-amz-meta-x-amz-matdesc header but I am not sure for other aws libraries. So for peace of mind, let's keep it as it is now.

api/src/main/java/io/minio/encryption/Crypto.java
```diff
@@ -0,0 +1,189 @@
+/*
```
NitishT: done

api/src/main/java/io/minio/encryption/Crypto.java
```diff
@@ -0,0 +1,189 @@
...
+   *           upon requested algorithm was not found
+   * @throws NoSuchPaddingException
+   *           upon incorrect padding
+   * @throws IllegalBlockSizeException
+   *           upon incorrect block size
+   * @throws BadPaddingException
+   *           upon wrong padding
+   */
+  public static CipherInputStream decryptInputStream(InputStream encryptedInputStream, SecretKey contentKey,
+      String mode, byte[] iv) throws InvalidKeyException, NoSuchAlgorithmException, NoSuchPaddingException,
```
NitishT: done

api/src/main/java/io/minio/encryption/Crypto.java
```diff
@@ -0,0 +1,189 @@
...
+      IllegalBlockSizeException, BadPaddingException, InvalidAlgorithmParameterException {
+
+    // Get the cipher
+    Cipher inputStrDecryptionCipher = Cipher.getInstance(mode);
```
NitishT: done

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1511,6 +1531,208 @@ public void getObject(String bucketName, String objectName, String fileName)
...
+   * 
+   */
+  public InputStream getObject(String bucketName, String objectName, SecretKey key)
+      throws InvalidBucketNameException, NoSuchAlgorithmException, InsufficientDataException, IOException,
+      InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException, InternalException,
+      InvalidArgumentException, NoSuchPaddingException, IllegalBlockSizeException,
+      BadPaddingException, InvalidAlgorithmParameterException, InvalidEncryptionMetadataException {
+
+    if (key == null) {
+      throw new InvalidArgumentException("empty decryption key not allowed");
```
NitishT: It is better to have proper exception thrown with detailed cause of exception.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1511,6 +1531,208 @@ public void getObject(String bucketName, String objectName, String fileName)
...
+    }
+    String keyDecryptionCipherMode = "AES/ECB/PKCS5Padding";
```
NitishT: Then how do you propose sending this info down to the private method? 

We need this variable to pass the transformation to be used for key encryption in the private function.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1511,6 +1531,208 @@ public void getObject(String bucketName, String objectName, String fileName)
...
+   */
+  public InputStream getObject(String bucketName, String objectName, KeyPair key) throws InvalidBucketNameException,
+      NoSuchAlgorithmException, InsufficientDataException, IOException, InvalidKeyException, NoResponseException,
+      XmlPullParserException, ErrorResponseException, InternalException, InvalidArgumentException,
+      NoSuchPaddingException, IllegalBlockSizeException, BadPaddingException, InvalidAlgorithmParameterException,
+      InvalidEncryptionMetadataException {
+    if (key == null) {
+      throw new InvalidArgumentException("empty decryption key pair not allowed");
+    }
+    String keyDecryptionCipherMode = "RSA";
```
NitishT: Then how do you propose sending this info down to the private method? 

We need this variable to pass the transformation to be used for key encryption in the private function.

api/src/main/java/io/minio/Crypto.java
```diff
@@ -0,0 +1,189 @@
...
+   * @throws NoSuchPaddingException
+   *           upon incorrect padding
+   * @throws IllegalBlockSizeException
+   *           upon incorrect block size
+   * @throws BadPaddingException
+   *           upon wrong padding
+   * @throws IOException
+   *           upon IO error
+   */
+  public static CipherInputStream encryptInputStream(InputStream plainInputStream, SecretKey contentKey,
```
balamurugana: same as above

api/src/main/java/io/minio/Crypto.java
```diff
@@ -0,0 +1,189 @@
...
+      IllegalBlockSizeException, BadPaddingException, InvalidAlgorithmParameterException, IOException {
+
+    // Get the cipher
+    Cipher cipher = Cipher.getInstance(transformation);
+
+    // init cipher with mode and data encryption key
+    cipher.init(Cipher.ENCRYPT_MODE, contentKey, new IvParameterSpec(iv));
+
+    // create cipherinputstream with plain stream and new initialized cipher
+    CipherInputStream cipherInputStream = new CipherInputStream(plainInputStream, cipher);
```
balamurugana: same as above.

Sorry for the late catch.

api/src/main/java/io/minio/MinioClient.java
balamurugana: I was talking about like `return getObject(bucketName, objectName, key, "AES/ECB/PKCS5Padding");`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1511,6 +1530,207 @@ public void getObject(String bucketName, String objectName, String fileName)
...
+    // Fetch encrypted object if metadata present
+    InputStream stream = getObject(bucketName, objectName);
+
+    // get the master key type and typecast
+    if (key instanceof KeyPair) {
+      masterKey = ((KeyPair) key).getPrivate();
+    } else if (key instanceof SecretKey) {
+      masterKey = (SecretKey) key;
+    } else {
+      throw new InternalException("key should be an instance of KeyPair or SecretKey");
```
balamurugana: below is better

```java
throw new InternalException("Unknown key object. This should not happen. Please report this issue at https://github.com/minio/minio-java/issues");
```

functional/FunctionalTest.java
```diff
@@ -247,6 +252,64 @@ public static void putObject_test6() throws Exception {
...
+    InputStream is = Files.newInputStream(Paths.get(fileName));
+
+    // Generate symmetric 256 bit AES key.
+    KeyGenerator symKeyGenerator = KeyGenerator.getInstance("AES");
+    symKeyGenerator.init(128);
+    SecretKey symKey = symKeyGenerator.generateKey();
+
+    try {
+      client.putObject(bucketName, fileName, is, 20 * 1024 * 1024, null, symKey);
+    } catch (InsufficientDataException e) {
```
balamurugana: You should not pass excessive size value (pass 13 * MB) and handle `InsufficientDataException` for this test.

functional/FunctionalTest.java
```diff
@@ -247,6 +252,64 @@ public static void putObject_test6() throws Exception {
...
+    InputStream is = Files.newInputStream(Paths.get(fileName));
+
+    // Generate RSA key pair
+    KeyPairGenerator keyGenerator = KeyPairGenerator.getInstance("RSA");
+    keyGenerator.initialize(1024, new SecureRandom());
+    KeyPair keypair = keyGenerator.generateKeyPair();
+
+    try {
+      client.putObject(bucketName, fileName, is, 20 * 1024 * 1024, null, keypair);
+    } catch (InsufficientDataException e) {
```
balamurugana: same as above.

functional/FunctionalTest.java
```diff
@@ -327,6 +390,152 @@ public static void getObject_test5() throws Exception {
...
+
+    // Generate symmetric 128 bit AES key.
+    KeyGenerator symKeyGenerator = KeyGenerator.getInstance("AES");
+    symKeyGenerator.init(128);
+    SecretKey symKey = symKeyGenerator.generateKey();
+
+    try {
+      // Put an encrypted object
+      client.putObject(bucketName, fileName, is, 20 * 1024 * 1024, null, symKey);
+    } catch (InsufficientDataException e) {
```
balamurugana: same as above.

functional/FunctionalTest.java
```diff
@@ -327,6 +390,152 @@ public static void getObject_test5() throws Exception {
...
+    os.close();
+    ois.close();
+
+    // Check if two files are not equal
+    File file1 = new File(fileName);
+    File file2 = new File(plainFileName);
+
+    if (FileUtils.contentEquals(file1, file2)) {
+      throw new MinioException("Files should not be equal");
+    }
```
balamurugana: You can use like below.  Test is not that critical to be optimized.

```java
    if (!Arrays.equals(Files.readAllBytes(Paths.get(filename)),
                       Files.readAllBytes(Paths.get(filename + ".downloaded")))) {
      throw new Exception("[FAILED] Test: presignedGetObject(String bucketName, String objectName)"
                          + ", Error: <Content differs>");
    }
```

You can refer `presignedGetObject_test1()`, how things are done simpler.

functional/FunctionalTest.java
```diff
@@ -327,6 +390,152 @@ public static void getObject_test5() throws Exception {
...
+
+    try {
+      // Put an encrypted object
+      client.putObject(bucketName, fileName, is, 20 * 1024 * 1024, null, symKey);
+    } catch (InsufficientDataException e) {
+      ignore();
+    }
+    is.close();
+
+    // Get the object without decryption
```
balamurugana: Why do we need test this?

functional/FunctionalTest.java
```diff
@@ -247,6 +252,64 @@ public static void putObject_test6() throws Exception {
   }
 
   /**
+   * Test: client side encryption: putObject(String bucketName, String objectName, InputStream stream, long size, String
+   * contentType, EncryptionMaterials encryptionMaterials).
```
balamurugana: You would need to fix these comments everywhere.

NitishT:
@balamurugana made the changes for unknown size put, please check

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2873,7 +3394,14 @@ private String putObject(String bucketName, String objectName, int length,
       queryParamMap.put(UPLOAD_ID, uploadId);
     }
 
-    HttpResponse response = executePut(bucketName, objectName, headerMap, queryParamMap, data, length);
+    if (uploadId == null) {
```
balamurugana: As this method is rest wrapper, its better to do this logic outside of this method.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3441,11 +3970,15 @@ private ListMultipartUploadsResult listIncompleteUploads(String bucketName, Stri
   /**
    * Initializes new multipart upload for given bucket name, object name and content type.
    */
-  private String initMultipartUpload(String bucketName, String objectName, String contentType)
+  private String initMultipartUpload(String bucketName, String objectName, String contentType,
+      Map<String, String> headerMap)
```
balamurugana: Remove `contentType` because we receive `headerMap`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2889,7 +3417,8 @@ private String putObject(String bucketName, String objectName, int length,
    * @param size         Size of object data.
    * @param data         Object data.
    */
-  private void putObject(String bucketName, String objectName, String contentType, Long size, Object data)
+  private void putObject(String bucketName, String objectName, String contentType, Long size, Object data,
+      Map<String, String> headerMap)
```
balamurugana: Remove `contentType` as we are receiving `headerMap`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2839,27 +3056,331 @@ public void putObject(String bucketName, String objectName, InputStream stream,
...
+   *          Object data.
+   * @param uploadId
+   *          Upload ID of multipart put object.
+   * @param partNumber
+   *          Part number of multipart put object.
    */
   private String putObject(String bucketName, String objectName, int length,
-                           Object data, String uploadId, String contentType,
-                           int partNumber)
+      Object data, String uploadId, String contentType, int partNumber, Map<String, String> headerMap)
```
balamurugana: remove `contentType` as we are receiving `headerMap` ie we would have `headerMap` updated with `contentType` prior to this call.

NitishT:
Done with all the changes @balamurugana

NitishT:
@harshavardhana @vadmeste please take a look.

NitishT:
Verified on AWS

PR #564, opened 2017-04-26
Fix: redact signature/creds values on printing

### LINKED ISSUE #563: Do not expose credentials and signature in requests trace and exception messages
Replace access key and signature with '*EXPOSED*' in both exception description and http tracing.

vadmeste:
Access key and signature were exposed in Exception description
and http tracing. This commit hides them.

Fixes #563

### DIFF
--- a/api/src/main/java/io/minio/errors/ErrorResponseException.java
+++ b/api/src/main/java/io/minio/errors/ErrorResponseException.java
@@ -57,7 +57,9 @@ public class ErrorResponseException extends MinioException {
         + "request={"
         + "method=" + request.method() + ", "
         + "url=" + request.httpUrl() + ", "
-        + "headers=" + request.headers()
+        + "headers=" + request.headers().toString()
+                .replaceAll("Signature=([0-9a-f]+)", "Signature=*REDACTED*")
+                .replaceAll("Credential=([^/]+)", "Credential=*REDACTED*")
         + "}\n"
         + "response={"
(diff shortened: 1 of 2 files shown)


api/src/main/java/io/minio/errors/ErrorResponseException.java
```diff
@@ -57,7 +57,9 @@ public String toString() {
         + "request={"
         + "method=" + request.method() + ", "
         + "url=" + request.httpUrl() + ", "
-        + "headers=" + request.headers()
+        + "headers=" + request.headers().toString()
+                .replaceAll("Signature=([0-9a-f]+)", "Signature=*REDACTED*")
+                .replaceAll("Credential=([^/]+)", "Credential=*REDACTED*")
```
balamurugana: minio-java doesn't add `Credential` header on request.
What does S3 call return `Credential` header?

harshavardhana: Credential is part of the Authorization header @balamurugana . This while making the request not what s3 sent back.

balamurugana: Got it.  We would need to change the commit log like
```
fix: redact signature/credential values in request headers on printing
```

harshavardhana: :+1:

PR #569, opened 2017-05-03
fix: conditionally calculate sha256/md5 hash for request.

### LINKED ISSUE #567: For all signed payloads with sha256 avoid calculating Content-Md5
Don't compute md5sum signatures when uploading for the first time, only compute when resuming the upload operation.

balamurugana:
* For authenticated access
  - For HTTPS (secure) request, SHA-256 hash is set to "UNSIGNED-PAYLOAD" and MD5 hash is calculated.
  - For HTTP (insecure) request, only SHA-256 hash is calculated.
* For anonymous access, for any secure/insecure request, only MD5 hash is calculated.

Fixes #567

### DIFF
--- a/api/src/main/java/io/minio/Digest.java
+++ b/api/src/main/java/io/minio/Digest.java
@@ -67,6 +67,19 @@ class Digest {
 
 
   /**
+   * Returns SHA-256 of given input stream and it's length.
+   *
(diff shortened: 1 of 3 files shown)


balamurugana:
Functional tests pass in `play.minio.io` and Amazon AWS S3.

harshavardhana:
@balamurugana rebase and conflict fix is needed..

PR #572, opened 2017-05-08
api: add MinioClient.ignoreCertCheck() method

balamurugana:
This patch fixes below
* ignoreCertCheck() enables MinioClient to ignore server certificate
  verification for HTTPS.
* Upgrade OkHttp and Okio libraries to the latest version.

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -19,12 +19,6 @@ package io.minio;
 
 import com.google.common.io.BaseEncoding;
 import com.google.common.io.ByteStreams;
-import com.squareup.okhttp.HttpUrl;
-import com.squareup.okhttp.MediaType;
-import com.squareup.okhttp.OkHttpClient;
-import com.squareup.okhttp.Request;
-import com.squareup.okhttp.RequestBody;
-import com.squareup.okhttp.Response;
 
 import io.minio.errors.ErrorResponseException;
 import io.minio.errors.InsufficientDataException;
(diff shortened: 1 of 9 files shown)


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -686,11 +695,54 @@ private void checkBucketName(String name) throws InvalidBucketNameException {
...
+      .hostnameVerifier(new HostnameVerifier() {
+          @Override
+          public boolean verify(String hostname, SSLSession session) {
+            return true;
+          }
+        })
+      .build();
+  }
+  
+
```
poornas: Readme.md should have this publicly available method documented.

balamurugana:
> Do you want to upgrade okhttp in one PR and then do ignoreCertCheck method with the subsequent one?

upgrading okhttp patch would have no purpose.  I don't like a PR with no purpose.  Do you want it in that way?

api/src/main/java/io/minio/MinioClient.java
balamurugana: Currently apis doing ReST calls are documented into README.md/API.md and others are covered in javadoc.

harshavardhana:
> upgrading okhttp patch would have no purpose. I don't like a PR with no purpose. Do you want it in that way?

Yes @balamurugana if we revert this PR for some reason, it would revert the changes of migrating to okhttp3 and vice versa. So it is easier if they are separate PRs. Unless you depend on the APIs present in okhttp3 not available okhttp2 to implement ignoreCertCheck() in that case it makes sense. what do you think?

balamurugana:
Yes @balamurugana if we revert this PR for some reason, it would revert the changes of migrating to okhttp3 and vice versa. So it is easier if they are separate PRs. Unless you depend on the APIs present in okhttp3 not available okhttp2 to implement ignoreCertCheck() in that case it makes sense. what do you think?

okhttp2 has a slightly different way of do that.

harshavardhana:
> okhttp2 has a slightly different way of do that.

Ok then i think its fine..

harshavardhana:
@poornas do you have anything to add?

PR #573, opened 2017-05-09
api: add support to putObject() with unknown sized input stream.

### LINKED ISSUE #341: make putobject to support unknown sized stream

balamurugana:
This patch brings new putObject() API.
```java
public void putObject(String bucketName, String objectName,
                      InputStream stream, String contentType)
```
Fixes #341

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -2917,7 +2917,9 @@ public final class MinioClient {
     String uploadId = getLatestIncompleteUploadId(bucketName, objectName);
     Iterator<Result<Part>> existingParts = null;
     Part part = null;
+    boolean isResumeMultipart = false;
     if (uploadId != null) {
+      isResumeMultipart = true;
       // resume previous multipart upload
       existingParts = listObjectParts(bucketName, objectName, uploadId).iterator();
       if (existingParts.hasNext()) {
@@ -2938,17 +2940,20 @@ public final class MinioClient {
       // For unknown sized stream, check available size.
       int availableSize = 0;
       if (unknownSize) {
-        availableSize = getAvailableSize(data, expectedReadSize);
-        if (availableSize != expectedReadSize) {
-          // the stream has less data
+        // Check whether data is available one byte more than expectedReadSize.
+        availableSize = getAvailableSize(data, expectedReadSize + 1);
+        // If availableSize is less or equal to expectedReadSize, then we reached last part.
+        if (availableSize <= expectedReadSize) {
+          // If it is first part, do single put object.
           if (partNumber == 1) {
-            // We do single put object here.
             putObject(bucketName, objectName, availableSize, data, null, contentType, 0);
-            abortMultipartUpload(bucketName, objectName, uploadId);
+            // if its not resuming previous multipart, remove newly created multipart upload.
+            if (!isResumeMultipart) {
+              abortMultipartUpload(bucketName, objectName, uploadId);
+            }
             return;
           }
 
-          // We have reached last part.
           expectedReadSize = availableSize;
           partCount = partNumber;
         }
@@ -3725,6 +3730,10 @@ public final class MinioClient {
   }
 
 
+  /**
+   * Return available size of given input stream up to given expected read size.  If less data is available than
+   * expected read size, it returns how much data available to read.
+   */
   private int getAvailableSize(Object inputStream, int expectedReadSize) throws IOException, InternalException {
     RandomAccessFile file = null;
     BufferedInputStream stream = null;
@@ -3733,9 +3742,11 @@ public final class MinioClient {
     } else if (inputStream instanceof BufferedInputStream) {
       stream = (BufferedInputStream) inputStream;
     } else {
-      throw new InternalException("Unknown input stream. This should not happen.  Please report");
+      throw new InternalException("Unknown input stream. This should not happen.  "
+                                  + "Please report to https://github.com/minio/minio-java/issues/");
     }
 
+    // hold current position of file/stream to reset back to this position.
     long pos = 0;
     if (file != null) {
       pos = file.getFilePointer();
@@ -3767,6 +3778,7 @@ public final class MinioClient {
       totalBytesRead += bytesRead;
     }
 
+    // reset back to saved position.
     if (file != null) {
       file.seek(pos);
     } else {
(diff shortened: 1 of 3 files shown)


functional/FunctionalTest.java
```diff
@@ -247,6 +247,25 @@ public static void putObject_test6() throws Exception {
   }
 
   /**
+   * Test: putObject(String bucketName, String objectName, String contentType, InputStream body).
```
NitishT: Comment can be updated.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3638,6 +3725,58 @@ private void skipStream(Object inputStream, long n)
   }
 
 
+  private int getAvailableSize(Object inputStream, int expectedReadSize) throws IOException, InternalException {
```
NitishT: You can add comments to explain what this method does.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3638,6 +3725,58 @@ private void skipStream(Object inputStream, long n)
...
+    RandomAccessFile file = null;
+    BufferedInputStream stream = null;
+    if (inputStream instanceof RandomAccessFile) {
+      file = (RandomAccessFile) inputStream;
+    } else if (inputStream instanceof BufferedInputStream) {
+      stream = (BufferedInputStream) inputStream;
+    } else {
+      throw new InternalException("Unknown input stream. This should not happen.  Please report");
```
NitishT: Add Minio Java issues link https://github.com/minio/minio-java/issues/

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2867,6 +2935,25 @@ private void putObject(String bucketName, String objectName, String contentType,
         expectedReadSize = lastPartSize;
       }
 
+      // For unknown sized stream, check available size.
+      int availableSize = 0;
+      if (unknownSize) {
+        availableSize = getAvailableSize(data, expectedReadSize);
+        if (availableSize != expectedReadSize) {
```
NitishT: Shouldn't the condition here be `<` instead of `!=`

functional/FunctionalTest.java
```diff
@@ -247,6 +247,25 @@ public static void putObject_test6() throws Exception {
...
+    is.close();
+    Files.delete(Paths.get(filename));
+    ObjectStat objectStat = client.statObject(bucketName, filename);
+    if (!customContentType.equals(objectStat.contentType())) {
+      throw new Exception("[FAILED] Test: putObject(String bucketName, String objectName, String contentType, "
+                          + "long size, InputStream body)");
+    }
+    client.removeObject(bucketName, filename);
+  }
+
```
NitishT: Can you also add a test case that will hit the multi-part upload flow, meaning filesize greater that 5 MiB

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3638,6 +3725,58 @@ private void skipStream(Object inputStream, long n)
...
+    BufferedInputStream stream = null;
+    if (inputStream instanceof RandomAccessFile) {
+      file = (RandomAccessFile) inputStream;
+    } else if (inputStream instanceof BufferedInputStream) {
+      stream = (BufferedInputStream) inputStream;
+    } else {
+      throw new InternalException("Unknown input stream. This should not happen.  Please report");
+    }
+
+    long pos = 0;
```
vadmeste: Can you add a comment here to explain why we are marking file/stream ?

api/src/main/java/io/minio/MinioClient.java
vadmeste: I don't think we should say "This should not happen" if the user can pass his input stream which is not RandomAccessFile or BufferedInputStream

vadmeste:
LGTM except only few comments

functional/FunctionalTest.java
balamurugana: Done

api/src/main/java/io/minio/MinioClient.java
balamurugana: Its not possible to have `availableSize > expectedReadSize`.  So `!=` means the same thing.  May be `<` is more readable.

api/src/main/java/io/minio/MinioClient.java

api/src/main/java/io/minio/MinioClient.java
balamurugana: > Add Minio Java issues link https://github.com/minio/minio-java/issues/

Done

balamurugana: > I don't think we should say "This should not happen" if the user can pass his input stream which is not RandomAccessFile or BufferedInputStream

Its not user passed input stream.  User passed input stream should have been wrapped with `BufferedInputStream` previously.  As its an internal error by programmers, the text is correct.

api/src/main/java/io/minio/MinioClient.java

functional/FunctionalTest.java
balamurugana: Done.
As mentioned in the commit log, for unknown sized stream, multipart size is not 5MiB, it is 525MiB.

api/src/main/java/io/minio/MinioClient.java
NitishT: `<=` will be better IMO

balamurugana:
All functional tests pass in play.minio.io and Amazon AWS S3.

functional/FunctionalTest.java
andersflemmen: @balamurugana Would it be possible to make it possible to reduce the multipart size for unknown sized streams? I want to avoid having to keep a lot data in memory, and will probably never have any files larger than 1 GB. Does the max object size or max number of multiparts have to be a constant value?

balamurugana: @andersflemmen Part size 525MiB is chosen for unknown sized stream multipart by calculation.  As we know maximum object size is 5TiB and maximum part count is 10000.  With these numbers, 525MiB is chosen.

As multipart upload is automatic in minio-java SDK, user has no control on part size.  Supporting custom part size requires design discussion.

Is there a possibility to stage the object data and call `putObject()` with size?

andersflemmen: @balamurugana I see! Would be nice to have the option to set the maximum object size, as we do not have file sizes close to 5TiB. We want to limit memory usage, and stream the data all the way from user upload and into Minio, so we will probably have to implement the multipart upload ourselves if it can't be provided through the default API.

PR #576, opened 2017-05-11
tests: Fix functional tests to run with Java1.7.

harshavardhana:
java1.7 defaults to using TLSv1.1 but Minio server
only support TLSv1.2 by default.

This PR addresses a problem which surfaced recently
while using Java 7 with Minio server.

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -21,11 +21,14 @@ import com.google.common.io.BaseEncoding;
 import com.google.common.io.ByteStreams;
 import com.squareup.okhttp.HttpUrl;
 import com.squareup.okhttp.MediaType;
-import com.squareup.okhttp.OkHttpClient;
 import com.squareup.okhttp.Request;
 import com.squareup.okhttp.RequestBody;
 import com.squareup.okhttp.Response;
+import com.squareup.okhttp.OkHttpClient;
+import com.squareup.okhttp.TlsVersion;
+import com.squareup.okhttp.ConnectionSpec;
 
+import io.minio.errors.InternalException;
 import io.minio.errors.ErrorResponseException;
 import io.minio.errors.InsufficientDataException;
 import io.minio.errors.InternalException;
@@ -89,14 +92,18 @@ import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
 import java.nio.file.StandardOpenOption;
 import java.security.InvalidAlgorithmParameterException;
+
 import java.security.InvalidKeyException;
 import java.security.Key;
 import java.security.KeyPair;
 import java.security.NoSuchAlgorithmException;
 import java.security.SecureRandom;
+import java.security.KeyManagementException;
+
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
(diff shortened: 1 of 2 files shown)


harshavardhana:
While this works @balamurugana , i need to make this selectable i.e need to change it use cleartext with HTTP v/s HTTPS. Until then :-1:

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -927,6 +937,23 @@ private HttpResponse execute(Method method, String region, String bucketName, St
...
+    }
+
+    SSLSocketFactory socketFac = context.getSocketFactory();
+
+    ConnectionSpec spec = new ConnectionSpec.Builder(ConnectionSpec.MODERN_TLS)
+        .tlsVersions(TlsVersion.TLS_1_2)
+        .build();
+
+    this.httpClient.setConnectionSpecs(Collections.singletonList(spec));
+    this.httpClient.setSslSocketFactory(socketFac);
```
balamurugana: * You could make this as a separate public method `public void enableTLSv12()`.  This make a choice to consumer to choose what they want than forcing to TLSv1.2.
* Its not required for each `execute()`
* you can directly like `this.httpClient.setSslSocketFactory(context.getSocketFactory());`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -927,6 +937,23 @@ private HttpResponse execute(Method method, String region, String bucketName, St
       this.traceStream.println(headers);
     }
 
+    SSLContext context = SSLContext.getInstance("TLSv1.2");
+
+    try {
+      context.init(null, null, null);
+    } catch (KeyManagementException e) {
```
balamurugana: You can raise `KeyManagementException` itself than eating with `InternalException`

functional/FunctionalTest.java
```diff
@@ -649,7 +654,24 @@ public static void presignedGetObject_test1() throws Exception {
...
+    }
+
+    SSLSocketFactory socketFac = context.getSocketFactory();
+    ConnectionSpec spec = new ConnectionSpec.Builder(ConnectionSpec.MODERN_TLS)
+        .tlsVersions(TlsVersion.TLS_1_2)
+        .build();
+
+    transport.setConnectionSpecs(Collections.singletonList(spec));
+    transport.setSslSocketFactory(socketFac);
+
```
balamurugana: You could have one `transport` as global where TLSv1.2 is enabled only for `play.minio.io` in `main()`.  This way repetition code is removed.

functional/FunctionalTest.java
```diff
@@ -649,7 +654,24 @@ public static void presignedGetObject_test1() throws Exception {
         .url(HttpUrl.parse(urlString))
         .method("GET", null)
         .build();
+
     OkHttpClient transport = new OkHttpClient();
+    SSLContext context = SSLContext.getInstance("TLSv1.2");
+
+    try {
+      context.init(null, null, null);
+    } catch (KeyManagementException e) {
```
balamurugana: You can throw `KeyManagementException` itself.

functional/FunctionalTest.java
```diff
@@ -899,7 +921,24 @@ public static void presignedGetObject_test2() throws Exception {
...
+    }
+
+    SSLSocketFactory socketFac = context.getSocketFactory();
+    ConnectionSpec spec = new ConnectionSpec.Builder(ConnectionSpec.MODERN_TLS)
+        .tlsVersions(TlsVersion.TLS_1_2)
+        .build();
+
+    transport.setConnectionSpecs(Collections.singletonList(spec));
+    transport.setSslSocketFactory(socketFac);
+
```
poornas: This code block can be refactored into a separate method, instead of repeating for each presigned url test cases.

PR #583, opened 2017-06-25
api: Enhance copyObject to support metadata.

harshavardhana:
This change allows copyObject() to replace metadata
with new metadata on the destination object. Additionally
this also allows for replacing metadata on existing objects
as well.

### DIFF
--- a/api/src/main/java/io/minio/CopyConditions.java
+++ b/api/src/main/java/io/minio/CopyConditions.java
@@ -30,6 +30,9 @@ import io.minio.errors.InvalidArgumentException;
  * before copying an object.
  */
 public class CopyConditions {
+  // Metadata directive "REPLACE" used to replace metadata on
+  // destination object in copyObject().
+  private static final String METADATA_DIRECTIVE_REPLACE = "REPLACE";
 
   private Map<String, String> copyConditions = new HashMap<>();
 
@@ -86,6 +89,15 @@ public class CopyConditions {
       throw new InvalidArgumentException("ETag cannot be empty");
     }
     copyConditions.put("x-amz-copy-source-if-none-match", etag);
+  }
+
+  /**
+   * Set replace metadata directive which specifies that
+   * destination object after copyObject() sets new
+   * metadata provided in the request.
+   */
+  public void setReplaceMetadataDirective() {
+    copyConditions.put("x-amz-metadata-directive", METADATA_DIRECTIVE_REPLACE);
   }
 
   /**
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -1807,7 +1807,7 @@ public final class MinioClient {
       NoResponseException, ErrorResponseException, InternalException, IOException, XmlPullParserException,
       InvalidArgumentException {
 
-    copyObject(bucketName, objectName, destBucketName, null, null);
+    copyObject(bucketName, objectName, destBucketName, null, null, null);
   }
 
   /**
@@ -1842,7 +1842,7 @@ public final class MinioClient {
       NoResponseException, ErrorResponseException, InternalException, IOException, XmlPullParserException,
       InvalidArgumentException {
 
-    copyObject(bucketName, objectName, destBucketName, destObjectName, null);
+    copyObject(bucketName, objectName, destBucketName, destObjectName, null, null);
   }
 
   /**
@@ -1880,7 +1880,7 @@ public final class MinioClient {
         NoResponseException, ErrorResponseException, InternalException, IOException, XmlPullParserException,
         InvalidArgumentException {
 
-    copyObject(bucketName, objectName, destBucketName, null, copyConditions);
+    copyObject(bucketName, objectName, destBucketName, null, copyConditions, null);
   }
 
   /**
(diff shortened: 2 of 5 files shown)


api/src/main/java/io/minio/CopyConditions.java
```diff
@@ -89,6 +89,20 @@ public void setMatchETagNone(String etag) throws InvalidArgumentException {
...
   /**
+   * Set metadata directive to specify whether the metadata
+   * is copied from the source object or replaced with metadata
+   * provided in the request.
+   */
+  public void setMetadataDirective(String value) {
+    if (value == null) {
+      // http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectCOPY.html
+      // Default value is - COPY.
+      value = "COPY";
```
balamurugana: Can you add respective enum typeto denote the value? I see only `COPY` and `REPLACE` are permitted.

functional/FunctionalTest.java
```diff
@@ -1391,6 +1391,44 @@ public static void copyObject_test7() throws Exception {
...
+                        copyConditions, metadata);
+    } catch (ErrorResponseException e) {
+      // File should not be copied as object was modified after date set in copyConditions.
+      if (!e.errorResponse().code().equals("PreconditionFailed")) {
+        throw e;
+      }
+    }
+
+    ObjectStat objectStat = client.statObject(bucketName, filename);
+    if (!customContentType.equals(objectStat.contentType())) {
```
balamurugana: Can you add test for `COPY` metadata directive?

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3158,58 +3209,58 @@ public void putObject(String bucketName, String objectName, InputStream stream,
...
    *
    * @see #putObject(String bucketName, String objectName, String fileName, String contentType)
-   * 
+   *
    * @see #putObject(String bucketName, String objectName, String fileName)
-   * 
+   *
    * @see #putObject(String bucketName, String objectName, InputStream stream, long size, String contentType)
-   * 
+   *
```
harshavardhana: I think this automatic in my `emacs` not sure if i can turn this off..

functional/FunctionalTest.java
```diff
@@ -692,7 +692,7 @@ public static void listObject_test5() throws Exception {
 
     // Check the number of uploaded objects
     if (i != objCount) {
-      throw new Exception("[FAILED] Test: recursive: listObject_test5(), number of items, expected: " 
+      throw new Exception("[FAILED] Test: recursive: listObject_test5(), number of items, expected: "
```
harshavardhana: As explained above..

functional/FunctionalTest.java
harshavardhana: COPY is default @balamurugana so didn't think that was necessary.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1877,7 +1879,8 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
         NoResponseException, ErrorResponseException, InternalException, IOException, XmlPullParserException,
         InvalidArgumentException {
 
-    copyObject(bucketName, objectName, destBucketName, null, copyConditions);
+    copyObject(bucketName, objectName, destBucketName, null, copyConditions,
+               null);
```
harshavardhana: This is not a cleanup @balamurugana it is reflecting the change in function signature.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1945,6 +1993,9 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
     if (copyConditions != null) {
       headerMap.putAll(copyConditions.getConditions());
     }
+    if (metadata != null) {
+      headerMap.putAll(Collections.unmodifiableMap(metadata));
```
harshavardhana: Kept it similar to getConditions()

api/src/main/java/io/minio/CopyConditions.java
harshavardhana: The reason i didn't add enum is because it wasn't needed,  server returns an error anyways if the wrong metadata directive was set.  Doing an enum disadvantage is that we are defining new types where none exists from a server perspective. 

Users of aws-sdk would have a hard time looking up these enums. Perhaps a strings constant is better..

api/src/main/java/io/minio/MinioClient.java
balamurugana: Got it.   The line could fit into 120 char long support.   The breakage makes me think of a cleanup.

api/src/main/java/io/minio/MinioClient.java
balamurugana: Yep. No need to wrap.

api/src/main/java/io/minio/MinioClient.java
balamurugana: My `emacs` doesn't do automatically.  I think there could be a way to do so.

api/src/main/java/io/minio/CopyConditions.java
harshavardhana: Adding a new enum anyways. to keep it similar in behavior to other types of similar code.

balamurugana: As we provide a method to set a value for `x-amz-metadata-directive` header, its required to validate given arg.  String constants are not java style.  enum provides a strict typing.  By looking into javadoc, everybody knows what to pass to the method.

api/src/main/java/io/minio/MinioClient.java
harshavardhana: Will see..

harshavardhana: Its basically whitespace cleanup.

balamurugana: Unwanted line modification makes review harder.  Its always recommended to send a separate PR for such cleanup than mingling with new API addition.

api/src/main/java/io/minio/CopyConditions.java
```diff
@@ -24,13 +24,13 @@
 import org.joda.time.DateTime;
 
 import io.minio.errors.InvalidArgumentException;
+import io.minio.MetadataDirective;
```
balamurugana: Its necessary in same package.

api/src/main/java/io/minio/CopyConditions.java
```diff
@@ -89,6 +89,20 @@ public void setMatchETagNone(String etag) throws InvalidArgumentException {
   }
 
   /**
+   * Set metadata directive to specify whether the metadata
+   * is copied from the source object or replaced with metadata
+   * provided in the request.
+   */
+  public void setMetadataDirective(MetadataDirective value) {
+    if (value == null) {
```
balamurugana: I don't think its necessary for an enum argument.

api/src/main/java/io/minio/MetadataDirective.java
```diff
@@ -0,0 +1,48 @@
...
+  private MetadataDirective(String value) {
+    this.value = value;
+  }
+
+  /**
+   * Returns MetadataDirective enum of given string.
+   */
+  public static MetadataDirective fromString(String md) {
+    if (md == null) {
+      throw new IllegalArgumentException("null md");
```
balamurugana: the error supposes to be `null metadata value`

api/src/main/java/io/minio/CopyConditions.java
balamurugana: Two kinds of implementation:
1. the current style.  user is required to pass correct enum even though copy is default.
2. another style is `setReplaceMetadataDirective()`.  This will add the header with `REPLACE` value. No enum necessary.

My choice is less code.  But its your call.

api/src/main/java/io/minio/CopyConditions.java
harshavardhana: you want me to move this to CopyCondtions?

api/src/main/java/io/minio/CopyConditions.java
harshavardhana: Like the setReplaceMetadataDirective behavior.

harshavardhana:
@balamurugana can you approve this? if nothing is pending.

PR #586, opened 2017-07-05
tests: Use content generation input stream

### LINKED ISSUE #585: Functional tests should use streaming
Minio-java functional tests will be used in Mint framework to test Minio server. This calls for uniform usage of Files across Minio-SDK functional tests. 

It was concluded that (https://github.com/minio/mint/pull/67/files)

- Tests should generate their own data by using input/output streams. These streams should be used for testing upload cases. 

This makes tests flexible, removes too much dependency on `/mint/data` directory and reduces the size of `mint` Docker image. 

Once these changes are done in minio-java, `mint` will be updated to use these test cases instead of maintaining copy of minio-java functional tests.

balamurugana:
This patch addes ContentInputStream which generates (repeated) random
data on read().  This helps avoid using temporary files and memories.

Fixes #585

### DIFF
--- a/functional/FunctionalTest.java
+++ b/functional/FunctionalTest.java
@@ -35,7 +35,6 @@ import okhttp3.Request;
 import okhttp3.RequestBody;
 import okhttp3.MultipartBody;
 import okhttp3.Response;
-import com.google.common.io.ByteStreams;
 
 import io.minio.*;
 import io.minio.messages.*;
@@ -47,6 +46,9 @@ public class FunctionalTest {
   private static final Random random = new Random(new SecureRandom().nextLong());
   private static final String bucketName = getRandomName();
   private static final String customContentType = "application/javascript";
+  private static boolean mintEnv = false;
+  private static Path dataFile1Mb;
+  private static Path dataFile65Mb;
   private static String endpoint;
   private static String accessKey;
   private static String secretKey;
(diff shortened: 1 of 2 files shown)


functional/MintDataFiles.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+enum MintDataFiles {
+  FILE_1_MB("/mnt/data/datafile-1-MB"), FILE_65_MB("/mnt/data/datafile-65-MB");
```
NitishT: Mint data dir path is `/mint/data`

harshavardhana: We should use and honor `DATA_DIR`

balamurugana: Done

functional/FunctionalTest.java
```diff
@@ -1692,6 +1697,13 @@ public static void main(String[] args) {
       System.exit(-1);
     }
 
+    if (System.getenv("MINIO_MINT") != null) {
```
NitishT: There is no such variable `MINIO_MINT` in Mint env . I think we can use `DATA_DIR` here.

balamurugana: As we discussed previously, `DATA_DIR` is too generic name ie functional test will fail in non-mint environment.  Its required to update `run.sh` like `MINIO_MINT= ./gradlew runFunctionalTest`

harshavardhana: We can create files in non mint env @balamurugana and continue the tests. Or skip those tests which require files.

functional/FunctionalTest.java
```diff
@@ -185,99 +233,106 @@ public static void putObject_test1() throws Exception {
...
-                       + "InputStream body)");
-    String filename = createFile(3 * MB);
-    InputStream is = Files.newInputStream(Paths.get(filename));
-    client.putObject(bucketName, filename, is, 1024 * 1024, customContentType);
+    System.out.println("Test: putObject(String bucketName, String objectName, InputStream body, "
+                       + "long size, String contentType)");
+
+    String objectName = getRandomName();
+    InputStream is = new ContentInputStream(3 * MB);
+    client.putObject(bucketName, objectName, is, 1 * MB, customContentType);
```
NitishT: Why 3 MB stream for 1 MB input?

functional/FunctionalTest.java
```diff
@@ -671,32 +653,30 @@ public static void listObject_test4() throws Exception {
...
+    System.out.println("Test: recursive: listObjects(final String bucketName, final String prefix, " 
+                       + "final boolean recursive)");
 
-    String baseFilename = createFile(1);
-    for (i = 0; i < objCount; i++) {
-      objectNames[i] = baseFilename + "-" + i;
-      client.putObject(bucketName, objectNames[i], baseFilename);
+    String[] objectNames = new String[1050];
+    int i = 0;
+    for (i = 0; i < 1050; i++) {
```
NitishT: why hardcode `objCount`?

functional/FunctionalTest.java
balamurugana: It is to indicate smaller content from larger stream can also be uploaded

functional/FunctionalTest.java
balamurugana: Done

PR #587, opened 2017-07-14
Enable gradle check in travis-ci builds

NitishT:
With this change, minio-java examples will be built with each PR. This is important to ensure the examples are working as expected.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ jdk:
 - oraclejdk7
 - openjdk7
 script:
+- ./gradlew check
 - if java -version 2>&1 | grep -q 'java version "1.8.'; then ./gradlew runFunctionalTest; else true; fi
 notifications:
   slack:


NitishT:
@harshavardhana noted.

.travis.yml
```diff
@@ -5,7 +5,7 @@ jdk:
 - oraclejdk7
 - openjdk7
 script:
-- if java -version 2>&1 | grep -q 'java version "1.8.'; then ./gradlew runFunctionalTest; else true; fi
+- if java -version 2>&1 | grep -q 'java version "1.8.'; then ./gradlew check; ./gradlew runFunctionalTest; else true; fi
```
balamurugana: By default travis does `check` for gradle projects. https://docs.travis-ci.com/user/languages/java/#Projects-Using-Gradle.  Looks like `script` overrides this.  If so, you would need to run `./gradlew check` for all java versions not only for `java version 1.8`

NitishT:
Updated @balamurugana

PR #590, opened 2017-07-21
api: add putObject() accepting additional meta data.

### LINKED ISSUE #450: How to set custom metadata to Object as part of putObject call
AWS S3 provides an API to set custom ObjectMetadata using which you can set custom properties such as etag / cache control headers. Is it possible to have ability to set such metadata via MinioClient API?

balamurugana:
Below method is added which facilitates user supplying additional meta data.
```java
public void putObject(String bucketName, String objectName, InputStream stream,
                      long size, Map<String, String> headerMap)
```

fixes #450

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -3147,7 +3147,7 @@ public final class MinioClient {
    * // create object
    * minioClient.putObject("my-bucketname", "my-objectname", bais, bais.available(), "application/octet-stream");
    * bais.close();
-   * System.out.println("my-bucketname is uploaded successfully"); }</pre>
+   * System.out.println("my-objectname is uploaded successfully"); }</pre>
    *
    * @param bucketName  Bucket name.
    * @param objectName  Object name to create in the bucket.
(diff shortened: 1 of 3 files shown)


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3109,6 +3109,77 @@ public void putObject(String bucketName, String objectName, InputStream stream,
...
+   *   builder.append("brave ghost pled; Watch Jeopardy!, Alex Trebeks fun TV quiz game.\n");
+   *   builder.append("---\n");
+   * }
+   * ByteArrayInputStream bais = new ByteArrayInputStream(builder.toString().getBytes("UTF-8"));
+   * // create object
+   * Map<String, String> headerMap = new HashMap<>();
+   * headerMap.put("Content-Type", "application/octet-stream");
+   * minioClient.putObject("my-bucketname", "my-objectname", bais, bais.available(), headerMap);
+   * bais.close();
+   * System.out.println("my-bucketname is uploaded successfully"); }</pre>
```
NitishT: s/my-bucketname/my-objectname

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3109,6 +3109,77 @@ public void putObject(String bucketName, String objectName, InputStream stream,
...
+   * @throws InternalException           upon internal library error
+   *
+   * @see #putObject(String bucketName, String objectName, String fileName)
+   */
+  public void putObject(String bucketName, String objectName, InputStream stream, long size,
+                        Map<String, String> headerMap)
+    throws InvalidBucketNameException, NoSuchAlgorithmException, InsufficientDataException, IOException,
+           InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
+           InternalException,
+           InvalidArgumentException, InsufficientDataException {
```
NitishT: Looks like an indentation problem here.

api/src/main/java/io/minio/MinioClient.java

api/src/main/java/io/minio/MinioClient.java
balamurugana: What is the problem?

NitishT: I think both `InvalidArgumentException, InsufficientDataException` can be on the same line as `InternalException`. I have noticed Checkstyle doesn't report such issues, but it doesn't look very good, so I pointed out

balamurugana: I see that up to `InternalException` its thrown for all APIs,   After `InternalException`, exceptions listed in new line are specifics to that API.   Its a same thing borrowed from other `putObject()` variants.

NitishT: ah..ok, then it's understandable.

balamurugana:
@NitishT No. examples carry commonly used `putObject()` variant i.e. not covering each variant.

PR #591, opened 2017-07-21
add constructor to accept custom http client

### LINKED ISSUE #589: Make MinioClient.httpClient extensible
Refer the actual issue filed at https://github.com/minio/minio/issues/4707

Currently `MinioClient.httpClient` field is `private` which limits of customization with interceptor.  To enable this feature, either one of following can be done
1. make `httpClient` field `protected`
2. have a method to accept interceptor
3. have a method to accept `OkHttpClient` which replaces existing `httpClient`

balamurugana:
Newly added constructor accepts customized http client which replaces
default http client used.

```java
public MinioClient(String endpoint, int port, String accessKey,
                   String secretKey, String region, boolean secure,
                   OkHttpClient httpClient)
```

Fixes #589

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -227,6 +227,8 @@ public final class MinioClient {
    * @see #MinioClient(String endpoint, String accessKey, String secretKey, boolean secure)
    * @see #MinioClient(String endpoint, int port, String accessKey, String secretKey, boolean secure)
    * @see #MinioClient(String endpoint, int port, String accessKey, String secretKey, String region, boolean secure)
+   * @see #MinioClient(String endpoint, int port, String accessKey, String secretKey, String region, boolean secure,
+   *          OkHttpClient httpClient)
    */
   public MinioClient(String endpoint) throws InvalidEndpointException, InvalidPortException {
     this(endpoint, 0, null, null);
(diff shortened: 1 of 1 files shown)


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -762,6 +762,19 @@ public boolean verify(String hostname, SSLSession session) {
...
   /**
+   * Set customized HTTP client.
+   *
+   * </p><b>Example:</b><br>
+   * <pre>{@code minioClient.setHttpClient(customHttpClient); }</pre>
+   *
+   * @param customHttpClient Customized HTTP client object.
+   */
+  public void setHttpClient(OkHttpClient customHttpClient) {
+    this.httpClient = customHttpClient;
```
poornas: shouldn't we guard here against null customHttpClient passed in?

balamurugana: Passing null will throw `NullPointerException` at runtime.  This method is for specific use case I think its OK.

balamurugana:
> This public method needs to be documented in Api docs.

`docs/API.md` scope is to document S3 methods.  Helper methods are documented in javadoc.

poornas:
@balamurugana , discussed this with @abperiasamy .He said the correct approach would be to provide a constructor with custom OkHttpClient as additional argument rather than providing a setHttpClient method. The Api gives the user a guarantee of thread-safe efficient HttpClient. Providing  a way for user to change the instantiated Http client gives room for underlying connection pooling, thread management and context that might be maintained by current instance of HttpClient to be replaced by another HttpClient midway through a bunch of requests. It would be safer to accept a custom HttpClient at the time of instantiating the constructor.

balamurugana:
I will have a constructor accepting http client.

poornas:
@balamurugana, will you be updating the Api.md as well?

balamurugana:
@poornas No. Only commonly used constructors are documented in API.md.  Rest of things are covered in javadoc.

PR #609, opened 2017-09-13
fix: handle > 1000 objects properly in batch removeObject()

balamurugana:


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2413,6 +2408,10 @@ public boolean hasNext() {
               populate();
```
kannappanr: Can these lines be changed to 

            if (this.error != null) {
              return true;
            }
            if (this.errorIterator == null || !this.errorIterator.hasNext()) {
              populate();
            }

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2435,6 +2434,10 @@ public boolean hasNext() {
...
```
kannappanr: Can this also be changed as suggested in the previous comment?

api/src/main/java/io/minio/MinioClient.java
poornas: Lines 2407 thru 2421 can be written more concisely as 

```
if (this.error != null) {
     return true;
}
if (this.errorIterator.hasNext()) {
     return true;
}
populate()
```

balamurugana: No. This will make loose of `error`  not addressed in every `populate()` especially first run i.e. `populate()` sets error.

api/src/main/java/io/minio/MinioClient.java
balamurugana: No.  My previous comment explains why

poornas:
@balamurugana , looks like appveyor and travis are failing.

balamurugana:
@poornas all CI ran successfully

PR #611, opened 2017-09-13
Support releasing an minio-java artifact from maven.

harshavardhana:
Earlier we used to rely on Nexus repo manager to
release minio-java jar's, but with this change we
can now just `release` the repository after
`uploadArchives`.

Additionally MAINTAINERS.md is updated with relevant
instructions.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -4,45 +4,58 @@ Minio Java SDK uses [gradle](https://gradle.org/) build system.
 ## Responsibilities
 Go through [Maintainer Responsibility Guide](https://gist.github.com/abperiasamy/f4d9b31d3186bbd26522).
 
-### Setup your minio-java Github Repository
-Fork [minio-java](https://github.com/minio/minio-java/fork) source repository to your own personal repository.
-```bash
-$ git clone https://github.com/$USER_ID/minio-java
+## Setup your minio-java Github Repository
+Clone [minio-java](https://github.com/minio/minio-java/) source repository locally.
+```sh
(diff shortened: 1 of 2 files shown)


MAINTAINERS.md
```diff
@@ -4,45 +4,58 @@ Minio Java SDK uses [gradle](https://gradle.org/) build system.
 ## Responsibilities
 Go through [Maintainer Responsibility Guide](https://gist.github.com/abperiasamy/f4d9b31d3186bbd26522).
 
-### Setup your minio-java Github Repository
+## Setup your minio-java Github Repository
 Fork [minio-java](https://github.com/minio/minio-java/fork) source repository to your own personal repository.
-```bash
+```sh
```
balamurugana: `bash` and `sh` do not make any difference in markdown

build.gradle
```diff
@@ -19,89 +19,90 @@
 /*************************/
 
 buildscript {
-    repositories {
-        jcenter()
-    }
-    dependencies {
-        classpath 'com.github.jengelman.gradle.plugins:shadow:1.2.3'
-    }
+  repositories {
```
balamurugana: I strongly recommend to avoid reformatting.  Unable to review the changes exactly.

harshavardhana: That's a mistake reverting..

MAINTAINERS.md
harshavardhana: sh is preferred over `bash` will stick to `sh`

deekoder:
```sh is fine.

PR #616, opened 2017-09-26
api: add new method to get presigned object url for given method.

### LINKED ISSUE #605: Allow presigned DELETE url's
Currently, the Java client does not support the generation of presigned URL's for deleting a file, only for PUT, POST and GET operations. 

In a scenario with a javascript frontend app and a java backend service, the service may choose to let frontend handle file operations directly with the minio server by issuing presigned URLs for upload and download. However, there is currently no way to let the frontend handle file deletions. 

It would be nice to have a presignedDeleteObject or a generic presignedRequest method to help support this scenario.

balamurugana:
new method.
```
public String getPresignedObjectUrl(Method method, String bucketName, String objectName,
                                    Integer expires, Map<String, String> reqParams)
```

Fixes #605

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -2102,6 +2102,51 @@ public class MinioClient {
   }
 
   /**
+   * Returns a presigned URL string with given HTTP method, expiry time and custom request params for a specific object
+   * in the bucket.
+   *
+   * </p><b>Example:</b><br>
+   * <pre>{@code String url = minioClient.getPresignedObjectUrl(Method.DELETE, "my-bucketname", "my-objectname",
+   * 60 * 60 * 24, reqParams);
+   * System.out.println(url); }</pre>
+   *
+   * @param method      HTTP {@link Method}.
+   * @param bucketName  Bucket name.
+   * @param objectName  Object name in the bucket.
+   * @param expires     Expiration time in seconds of presigned URL.
+   * @param reqParams   Override values for set of response headers. Currently supported request parameters are
+   *                    [response-expires, response-content-type, response-cache-control, response-content-disposition]
+   *
+   * @return string contains URL to download the object.
+   *
+   * @throws InvalidBucketNameException   upon an invalid bucket name
+   * @throws InvalidKeyException          upon an invalid access key or secret key
+   * @throws IOException                  upon signature calculation failure
+   * @throws NoSuchAlgorithmException     upon requested algorithm was not found during signature calculation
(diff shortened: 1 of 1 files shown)


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2102,12 +2102,15 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
   }
 
   /**
-   * Returns an presigned URL to download the object in the bucket with given expiry time with custom request params.
+   * Returns an presigned URL string for given method of given object in the bucket with given expiry time with custom
+   * request params.
```
poornas: suggestion for readability:
Generates a presigned URL string with given HTTP method, expiry time and custom request params for a specific object in the bucket

balamurugana: slightly modified

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2122,8 +2125,8 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
    * @throws NoSuchAlgorithmException     upon requested algorithm was not found during signature calculation
    * @throws InvalidExpiresRangeException upon input expires is out of range
    */
-  public String presignedGetObject(String bucketName, String objectName, Integer expires,
```
ebozduman: I think it would be better if we keep the old `presignedGetObject` method in place for backward compatibility reasons and have it call the newly introduced method `getPresignedObjectUrl` with `method='GET'` internally.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2132,14 +2135,47 @@ public String presignedGetObject(String bucketName, String objectName, Integer e
...
+   *
+   * @return string contains URL to download the object.
+   *
+   * @throws InvalidBucketNameException   upon an invalid bucket name
+   * @throws InvalidKeyException          upon an invalid access key or secret key
+   * @throws IOException                  upon signature calculation failure
+   * @throws NoSuchAlgorithmException     upon requested algorithm was not found during signature calculation
+   * @throws InvalidExpiresRangeException upon input expires is out of range
+   */
+  public String presignedGetObject(String bucketName, String objectName, Integer expires,
```
ebozduman: We can have `getPresignedObjectUrl` method called within `presignedPutObject` exactly the same way  `presignedGetObject` is modified.

balamurugana:
Currently `API.md` and `README.md` contain commonly used API documentations. Rest are covered in Javadoc https://minio.github.io/minio-java/

PR #618, opened 2017-09-27
tests: honor mint mode

balamurugana:
When functional tests are run in mint, mint passes mode (full or core)
to indicate what set of tests to be run.  This patch handles mint mode
and executes tests accordingly.

### DIFF
--- a/functional/FunctionalTest.java
+++ b/functional/FunctionalTest.java
@@ -49,9 +49,9 @@ public class FunctionalTest {
   private static final String IGNORED = "NA";
   private static final int MB = 1024 * 1024;
   private static final Random random = new Random(new SecureRandom().nextLong());
-  private static final String bucketName = getRandomName();
   private static final String customContentType = "application/javascript";
   private static final String nullContentType = null;
+  private static String bucketName = getRandomName();
   private static boolean mintEnv = false;
   private static Path dataFile1Mb;
   private static Path dataFile65Mb;
@@ -2488,9 +2488,9 @@ public class FunctionalTest {
   }
 
   /**
-   * runFastTests: runs a fast set of tests.
+   * runQuickTests: runs tests those completely quicker.
    */
-  public static void runFastTests() throws Exception {
+  public static void runQuickTests() throws Exception {
     makeBucket_test1();
     listBuckets_test();
     bucketExists_test();
@@ -2535,6 +2535,11 @@ public class FunctionalTest {
       dataFile65Mb = Paths.get(dataDir, "datafile-65-MB");
     }
 
+    String mintMode = null;
+    if (mintEnv) {
+      mintMode = System.getenv("MINT_MODE");
+    }
+
     endpoint = args[0];
     accessKey = args[1];
     secretKey = args[2];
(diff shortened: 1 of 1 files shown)


functional/FunctionalTest.java
```diff
@@ -2473,12 +2478,23 @@ public static void main(String[] args) {
...
       // Enable trace for debugging.
       // client.traceOn(System.out);
-      FunctionalTest.runTests();
-
-      // Run fast test with region parameter passed to the constructor
-      client = new MinioClient(endpoint, accessKey, secretKey, region);
-      FunctionalTest.runFastTests();
 
+      // For mint environment, run tests based on mint mode
+      if (mintEnv) {
```
harshavardhana: This would mean that functional tests won't be run in full by default.. I thought its the other way around if needed for quick tests MINT_MODE is set to `quick` if not we run full mode for both mint and regular SDK tests.  

This also simplifies the condition we have below.

functional/FunctionalTest.java
```diff
@@ -2473,12 +2478,23 @@ public static void main(String[] args) {
...
+        if (mintMode != null && mintMode.equals("full")) {
+          FunctionalTest.runTests();
+        } else {
+          FunctionalTest.runQuickTests();
+        }
+      } else {
+        FunctionalTest.runTests();
+
+        // Get new bucket name to avoid minio azure gateway failure.
+        bucketName = getRandomName();
```
vadmeste: Looks like this line doesn't have an effect.

functional/FunctionalTest.java
```diff
@@ -2473,12 +2478,23 @@ public static void main(String[] args) {
...
+        // Quick tests with passed region.
+        client = new MinioClient(endpoint, accessKey, secretKey, region);
```
vadmeste: Also this line, what is this for?

functional/FunctionalTest.java
balamurugana: No.  Accordingly to mint, mint's default mode is `quick`

functional/FunctionalTest.java
balamurugana: Unable to understand

functional/FunctionalTest.java
harshavardhana: It won't be correct for quick to be default, the tests are not sufficient to validate the server.  Most of us will happily think server is working fine.  MINT_MODE=quick is only needed on special situations i.e CI/CD envs . So i don't see why special situations should be default behavior.

functional/FunctionalTest.java
vadmeste: > I am unable to understand your question

I mean isn't this line of code useless ? client seems like an unused variable for me.

balamurugana: Nope. You missed `client` is `private static` to `public class FunctionalTest` and all tests use `client`

vadmeste: > Nope. You missed client is private static to public class FunctionalTest and all tests use client

okay.. also, it is little weird that the client is being instantiated many times in main(). I guess we can see that later.

functional/FunctionalTest.java
harshavardhana: Understood after discussing with @balamurugana ,we have to properly pick the tests for sanity. Right now minio-java has chosen enough set of tests for sanity (there is room to add more). But this is what we need to repeat across all our libraries.

functional/FunctionalTest.java
nitisht: Why do you need to initialize `client` again here?

balamurugana: Because of `// Quick tests with passed region.`.  Full tests are run without passed region.

nitisht: I see.

PR #621, opened 2017-09-30
fix: set default connection timeout to 15 minutes

balamurugana:
The default timeout 10 seconds of OkHttp is not sufficient for some
server response.  This patch fixes this by setting it to 15 minutes
like how Amazon AWS treat time difference in the signature.


nitisht:
Build seems to be failing for Travis, is this expected?

balamurugana:
@poornas added the description

deekoder:
I am not sure what happened @poornas  was dropped by github as a reviewer. So readded.

PR #622, opened 2017-10-02
tests: add more checks in list buckets functional test

balamurugana:


functional/FunctionalTest.java
```diff
@@ -300,8 +300,29 @@ public static void listBuckets_test() throws Exception {
...
     long startTime = System.currentTimeMillis();
     try {
+      Date expectedTime = new Date();
+      String bucketName = getRandomName();
+      boolean found = false;
+      client.makeBucket(bucketName);
       for (Bucket bucket : client.listBuckets()) {
-        ignore(bucket);
+        if (bucket.name().equals(bucketName)) {
+          if (found) {
```
kannappanr: When would we have a duplicate entry? In what failure scenario, do you think this will happen?

balamurugana: It should not happen.  A buggy server may return duplicate entry so this check is added.

PR #625, opened 2017-10-03
fix: add method to return http headers in ObjectStat

### LINKED ISSUE #624: How to get custom meta data that save using putObject method ?
Hi,
   i have checked the put object method for saving the custom meta data along with file upload . Is there any other way to save custom meta data . Also how can i retrieve this . not found one with getObject and statObject method . please suggest a solution for this .

My example is given below
            
           ` Map<String, String> headerMap = new HashMap();
            headerMap.put("x-amz-meta-name", "pmv3");
            headerMap.put("x-amz-meta-size", "1025");
            
            minioClient.putObject("test1", "pmv3",
            stream,stream.available(),headerMap);           
            ObjectStat objectStat = minioClient.statObject("test1", "pmv3");      
            
            System.out.println("objectStat "+objectStat);`

 Output doesnt contain the headerMap values . Please suggest the way to find this .

balamurugana:
Previously ObjectStat did not carry HTTP headers returned from S3
server.  This patch fixes the issue by adding a new method

`Map<String,List<String>> ObjectStat.httpHeaders()`

which returns received headers from the server.

Fixes #624

### DIFF
--- a/api/src/main/java/io/minio/HttpResponse.java
+++ b/api/src/main/java/io/minio/HttpResponse.java
@@ -16,6 +16,9 @@
 
 package io.minio;
 
+import java.util.List;
+import java.util.Map;
+
 import okhttp3.Response;
 import okhttp3.ResponseBody;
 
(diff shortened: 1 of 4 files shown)


kannappanr:
Agree with @nitisht on adding a new functional test. Otherwise, this looks good to me. Do you want to handle that in a separate PR?

balamurugana:
Yep. I am waiting on @Prasanthmv for confirming the solution working.

Prasanthmv:
@balamurugana  Let me check. I will inform once its working fine

PR #633, opened 2017-10-23
fix: do not initialize httpClient in MinioClient class level.

### LINKED ISSUE #628: Issue in the way OkHttpClient is used in the Minio Java Client
These seems to be an issue in the way OkHttpClient is used in the Minio Java Client

*Context:*
We are trying to connect to Minio using the Mimio-java client from a SpringBoot application.
+ JDK 1.8.0_112
+ Minio Client - 3.0.7

The SpringBoot application is using a custom PKI provider.

Issue faced:
It looks like Minio Java client is using the OKHTTPClient. And the problem appears to be how Minio Java client is using the OkHttPClient.

https://github.com/minio/minio-java/blob/master/api/src/main/java/io/minio/MinioClient.java  

So even though Minio Java Client provides a constructor to pass the OkHttpClient (see below), it is already initializing the OkHttpClient at the Class level in line 202:

This is the constructor that we want to use
`MinioClient(String endpoint, int port, String accessKey, String secretKey, String region, boolean secure, OkHttpClient httpClient)`

line 202
`private OkHttpClient httpClient = new OkHttpClient();`

So if this httpClient initialization is moved to the relevant Constructors (& not at the class level), we think that the issue can be resolved.

The problem happens when the default OkHttpClient tries to set the default TrustManager:

Within the OkHttpClient
https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/OkHttpClient.java 
Line 284
`trustManagerFactory.init((KeyStore) null);`
Not all the custom PKI providers handle this null gracefully.

*Work around (used for now):*
We switched from using the Minio Java Client to AmazonS3 Java client and everything works as expected.

balamurugana:
Previously httpClient was initialized at class level which causes
error with custom PKI implementation. In this situation any java
program using MinioClient (import io.minio.MinioClient;) would cause
failure.

This patch fixes the issue by initalizing httpClient in constructor
level

Fixes #628

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -199,7 +199,7 @@ public class MinioClient {
 
   private String userAgent = DEFAULT_USER_AGENT;
 
-  private OkHttpClient httpClient = new OkHttpClient();
+  private OkHttpClient httpClient;
 
 
   /**
@@ -635,6 +635,7 @@ public class MinioClient {
     if (httpClient != null) {
       this.httpClient = httpClient;
     } else {
+      this.httpClient = new OkHttpClient();
       this.httpClient = this.httpClient.newBuilder()
         .connectTimeout(DEFAULT_CONNECTION_TIMEOUT, TimeUnit.SECONDS)
         .writeTimeout(DEFAULT_CONNECTION_TIMEOUT, TimeUnit.SECONDS)


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -199,7 +199,7 @@
 
   private String userAgent = DEFAULT_USER_AGENT;
 
-  private OkHttpClient httpClient = new OkHttpClient();
+  private OkHttpClient httpClient;
```
nitisht: Shouldn't `httpClient` be initialized in all other constructors (other than `MinioClient(String endpoint, int port, String accessKey, String secretKey, String region, boolean secure, OkHttpClient httpClient)` as well?

balamurugana: Not required. Because all constructors call this constructor only.

PR #638, opened 2017-10-30
api: getObject should set range offset when its > 0

### LINKED ISSUE #636: Get empty File
Hi,

I uploaded an empty file (length = 0 Bytes) to minio.
If I try to download it with the getObject Method I get the following error:
```
ErrorResponse(code=InvalidRange, message=The requested range is not satisfiable, bucketName=null, objectName=null, resource=/my-bucket/test, requestId=3L137, hostId=3L137)
request={method=GET, url=http://127.0.0.1:9000/my-bucket/test, headers=Range: bytes=0-
Host: 127.0.0.1:9000
```
I have created a minimal example that reproduces the issue here:
https://gist.github.com/JNKielmann/cd7283b0d7ec9cc57b7fea36c46395b0

The problem seems to be the Range header: `bytes=0-` because the file has no bytes. If you remove this entry from the request it succeeds.

Is this a bug or are empty files just not supported? If it's a bug I can make a PR if you want.

harshavardhana:
AWS S3 and Minio both return ErrInvalidRange if the
object is 0 bytes and the Range header is '0-'

Fixes #636

### DIFF
--- a/functional/FunctionalTest.java
+++ b/functional/FunctionalTest.java
@@ -104,7 +104,7 @@ public class FunctionalTest {
 
     return createFile(1 * MB);
   }
-    
+
   /**
    * Create 65 MB temporary file.
    */
@@ -117,7 +117,7 @@ public class FunctionalTest {
 
     return createFile(65 * MB);
   }
-    
+
   /**
    * Generate random name.
    */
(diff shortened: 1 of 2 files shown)


functional/FunctionalTest.java
```diff
@@ -990,6 +990,32 @@ public static void getObject_test7() throws Exception {
   }
 
   /**
+   * Test: getObject(String bucketName, String objectName) - download 0 bytes.
+   */
+  public static void getObject_test8() throws Exception {
+    if (!mintEnv) {
+      System.out.println("Test: getObject(String bucketName, String objectName)");
```
balamurugana: please add "zero size object"  at the end of the string

harshavardhana:
Fixed @balamurugana

PR #650, opened 2017-12-01
fix: preserve trailing '/' in GetBucketLocation request

### LINKED ISSUE #648: SignatureDoesNotMatch if using nginx route based on bucket names
Accoding to the [Setup Nginx proxy with Minio Server](https://docs.minio.io/docs/setup-nginx-proxy-with-minio), we can using nginx route based on bucket names.
If I config nginx like the doc said：
```
location /somebucketName/ {
   proxy_buffering off;
   proxy_set_header Host $http_host;
   proxy_pass http://localhost:9000;
 }
```
The java client would throw SignatureDoesNotMatch exception,but the mc client works fine.
I did some debugging,and find out why.
This is the nginx log
```
"GET /somebucketName?location= HTTP/1.1" 301 185 "
"GET /somebucketName/?location= HTTP/1.1" 403 348 "
```
Because the nginx location is `/somebucketName/`,the `/` suffix did the trick,the java client use `/somebucketName` as the url (no `/` suffix ),but nginx rewrite the url to `/somebucketName/` (with the `/` as suffix),so the client signature does not match the server signature.If I config the location like `/somebucketName` (with no `/` suffix),everything works fine.
This may not be a bug of java client,but is not consistent with mc and the minio docs.
Thanks!

nitisht:
Fixes #648

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -903,6 +903,10 @@ public class MinioClient {
       for (String pathSegment : objectName.split("/")) {
         urlBuilder.addEncodedPathSegment(S3Escaper.encode(pathSegment));
       }
+      if (objectName.endsWith("/")) {
+        // Fix issue #648: preserve trailing '/' to work with proxies like nginx.
+        urlBuilder.addPathSegment("");
+      }
     }
 
     if (queryParamMap != null) {
@@ -912,7 +916,7 @@ public class MinioClient {
     }
 
     HttpUrl url = urlBuilder.build();
-
+    
     Request.Builder requestBuilder = new Request.Builder();
     requestBuilder.url(url);
     if (headerMap != null) {
@@ -1209,11 +1213,11 @@ public class MinioClient {
            InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
            InternalException {
     if (bucketName != null && this.accessKey != null && this.secretKey != null
-        && !BucketRegionCache.INSTANCE.exists(bucketName)) {
+          && !BucketRegionCache.INSTANCE.exists(bucketName)) {
       Map<String,String> queryParamMap = new HashMap<>();
       queryParamMap.put("location", null);
 
-      HttpResponse response = execute(Method.GET, US_EAST_1, bucketName, null,
+      HttpResponse response = execute(Method.GET, US_EAST_1, bucketName, "/",
           null, queryParamMap, null, 0);
 
       // existing XmlEntity does not work, so fallback to regular parsing.


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1209,11 +1214,11 @@ private void updateRegionCache(String bucketName)
            InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
            InternalException {
     if (bucketName != null && this.accessKey != null && this.secretKey != null
-        && !BucketRegionCache.INSTANCE.exists(bucketName)) {
+          && !BucketRegionCache.INSTANCE.exists(bucketName)) {
       Map<String,String> queryParamMap = new HashMap<>();
       queryParamMap.put("location", null);
 
-      HttpResponse response = execute(Method.GET, US_EAST_1, bucketName, null,
+      HttpResponse response = execute(Method.GET, US_EAST_1, bucketName, "/",
```
balamurugana: With little thought, adding `/` doesn't give a value.  Two cases here.
- path style.  there is no meaning to adding `/`.  its not http specification (I need to verify)
- url style. no need to add in this case.

I think nginix is mis-configured.  Please verify.

nitisht: yes, setting nginx location to `/bucketName` instead of `/bucketName/` will make it everything work as expected. 

However, I verified with other SDKs (minio-go, aws-sdk-go) and they work with both configurations (`/bucketName` and `/bucketName/`), and minio-java should be compatible as well.

harshavardhana: The problem is perhaps that minio-java is not redirecting correctly, when a `/bucketName` request is made it re-directs to `/bucketName/` but doesn't reAuth the request.

This type of change is not needed if the re-authentication is enabled.

nitisht: > The problem is perhaps that minio-java is not redirecting correctly, when a /bucketName request is made it re-directs to /bucketName/ but doesn't reAuth the request.

minio-java sends the request to `/bucketName` while other SDKs send it to `/bucketName/`. 

When nginx location is set to `/bucketName/` , nginx, irrespective of income request location adds the trailing `/` (if not present), i.e .nginx does the redirection.

Since minio-java calculated signature based on `/bucketname`, it fails due to signature mismatch. Since other SDKs signed based on `/bucketName/`, it works fine.

balamurugana: How does aws-sdk-java handle get bucket location query? with `/` or without `/`?

nitisht: let me check and confirm, I had tried with aws-sdk-go .

balamurugana: You could try aws-sdk-java with nginix configured with/without `/` on AWS S3 and play.minio.io

nitisht: > aws-sdk-java with ngnix configured with/without / on AWS S3 and play.minio.io

Both upstreams you mentioned are SSL enabled, and sending secure traffic to upstream servers needs detailed configuration and CA certificate (ref: https://www.nginx.com/resources/admin-guide/nginx-tcp-ssl-upstreams/) 

So, I tested with both Nginx configurations `/bucketName` and `/bucketName/` with local upstream Minio server built from latest master, in both the cases `aws-sdk-java` `getBucketLocation` API sends out requests to `bucketName/`, this is Minio server trace

```sh
[REQUEST (objectAPIHandlers).GetBucketLocationHandler-fm] [151223897.018844] [2017-12-02 23:52:50 +0530]
GET /photos/?location
Host: domain.abc
Connection: close
Content-Length: 0
X-Amz-Date: 20171202T182250Z
Amz-Sdk-Retry: 0/0/500
X-Amz-Content-Sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
Authorization: AWS4-HMAC-SHA256 Credential=minio/20171202//s3/aws4_request, SignedHeaders=amz-sdk-invocation-id;amz-sdk-retry;content-type;host;user-agent;x-amz-content-sha256;x-amz-date, Signature=a47038e0a9611345c9a745d419951aef434eb54a899c8735289a53a0b5776d44
User-Agent: aws-sdk-java/1.11.240 Linux/4.13.0-16-lowlatency OpenJDK_64-Bit_Server_VM/25.151-b12 java/1.8.0_151
Amz-Sdk-Invocation-Id: 4b598be3-0530-b9b5-222b-7626ca651776
Content-Type: application/octet-stream


[RESPONSE] [151223897.018844] [2017-12-02 23:52:50 +0530]
200 OK
X-Amz-Request-Id: 14FC8D54C99A0977
Server: Minio/DEVELOPMENT.GOGET (linux; amd64)
Accept-Ranges: bytes
Content-Type: application/xml
Vary: Origin
```

balamurugana: This means URI ending with `/` comes with a meaning i.e. `createRequest()` should preserve `/`.  This fix needs to be done accordingly.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -899,9 +899,14 @@ private Request createRequest(Method method, String bucketName, String objectNam
     }
 
     if (objectName != null) {
-      // Limitation: OkHttp does not allow to add '.' and '..' as path segment.
-      for (String pathSegment : objectName.split("/")) {
-        urlBuilder.addEncodedPathSegment(S3Escaper.encode(pathSegment));
+      if (!objectName.equals("/")) {
```
balamurugana: this `if` is not required.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -899,9 +899,14 @@ private Request createRequest(Method method, String bucketName, String objectNam
...
+        // Limitation: OkHttp does not allow to add '.' and '..' as path segment.
+        for (String pathSegment : objectName.split("/")) {
+          urlBuilder.addEncodedPathSegment(S3Escaper.encode(pathSegment));
+        }
+      } else if (objectName.equals("/")) {
```
balamurugana: Have this `if` checking `.endsWith("/")` and add comment about this fix and issue#

nitisht:
updated @balamurugana

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -903,6 +903,12 @@ private Request createRequest(Method method, String bucketName, String objectNam
       for (String pathSegment : objectName.split("/")) {
         urlBuilder.addEncodedPathSegment(S3Escaper.encode(pathSegment));
       }
+      if (objectName.endsWith("/")) {
+        // When objectName ends with "/", we add a path segment / at the end of URL. 
+        // This is to comply with aws-sdk-java behavior 
+        // See https://github.com/minio/minio-java/issues/648 for more details
```
balamurugana: Below comment is good enough

```
// Fix issue #648: preserve tailing '/' to work with proxies like nginx.
```

nitisht: :+1:

poornas: trailing '/' is better

nitisht: @poornas I have updated the comment as `// Fix issue #648: preserve tailing '/' to work with proxies like nginx.`

nitisht: @poornas I got your point now :)

nitisht:
updated the commit message @balamurugana

PR #652, opened 2017-12-05
fix: Add parameter validation for removeObject and getObject

### LINKED ISSUE #651: Argument validation should be more strict.
Hi!
In this method:
``` java
public void removeObject(String bucketName, String objectName)
```
If objectName is null or empty(due to some bad coding),this method will delete the whole bucket,which is very dangerous,even removeBucket would not recursively delete objects.

The api description should be more precise,and do the argument validation strictly,if there is a  invalid argument , an IllegalArgumentException would be better.

There is other  method like `getObject(String bucketName, String objectName)`,if objectName is null or empty,this method would act as ListBucketResult with no prefix.

### 3.0.8

nitisht:
fixes #651

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -1578,6 +1578,14 @@ public class MinioClient {
            InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
            InternalException,
            InvalidArgumentException {
+    if ((bucketName == null) || (bucketName.isEmpty())) {
+      throw new InvalidArgumentException("bucket name cannot be empty");
+    }
+  
+    if ((objectName == null) || (objectName.isEmpty())) {
+      throw new InvalidArgumentException("object name cannot be empty");
+    }
+      
     if (offset < 0) {
       throw new InvalidArgumentException("offset should be zero or greater");
     }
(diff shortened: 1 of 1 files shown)


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1505,8 +1505,15 @@ public String getObjectUrl(String bucketName, String objectName)
...
-           InternalException,
-           InvalidArgumentException {
+           InternalException, InvalidArgumentException {
+    if ((bucketName == null) || (bucketName.isEmpty())) {
+      throw new InvalidArgumentException("bucket name cannot be empty");
+    }
+    
+    if ((objectName == null) || (objectName.isEmpty())) {
+      throw new InvalidArgumentException("object name cannot be empty");
+    }
```
balamurugana: Add this check at line #1849 and #1581

nitisht: updated @balamurugana

balamurugana: If so, why this check is here?

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1586,6 +1593,14 @@ public InputStream getObject(String bucketName, String objectName, long offset,
       throw new InvalidArgumentException("length should be greater than zero");
     }
 
+    if ((bucketName == null) || (bucketName.isEmpty())) {
+      throw new InvalidArgumentException("bucket name cannot be empty");
+    }
+    
+    if ((objectName == null) || (objectName.isEmpty())) {
+      throw new InvalidArgumentException("object name cannot be empty");
+    }
```
balamurugana: Do this validation first.  See the order of method arguments.

nitisht:
I get your point now @balamurugana . Updated the PR to add checks only for final `getObject` and `removeObject` methods.  Removed the checks from wrapper methods

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1576,16 +1575,23 @@ public InputStream getObject(String bucketName, String objectName, long offset)
...
+      
     if (offset < 0) {
       throw new InvalidArgumentException("offset should be zero or greater");
     }
 
     if (length != null && length <= 0) {
       throw new InvalidArgumentException("length should be greater than zero");
     }
-
+
```
balamurugana: fix this whitespace

nitisht: done

PR #664, opened 2018-03-10
Remove client side encryption

nitisht:
As Minio server now supports SSE, it was planned to remove support
for client side encryption.


deekoder:
Closing this so that we can have 1 pr where the removal and addition is available.

nitisht:
This is ready for review again

api/src/main/java/io/minio/ErrorCode.java
```diff
@@ -54,7 +54,6 @@
   INVALID_BUCKET_NAME("InvalidBucketName", "Bucket name is not valid"),
   INVALID_BUCKET_STATE("InvalidBucketState", "The request is not valid with the current state of the bucket"),
   INVALID_DIGEST("InvalidDigest", "Specified Content-MD5 is not valid"),
-  INVALID_ENCRYPTION_ALGORITHM_ERROR("InvalidEncryptionAlgorithmError", "Invalid encryption algorithm error"),
```
balamurugana: Capturing more server error code is better than removing them.  This can not be removed.

functional/FunctionalTest.java
```diff
@@ -714,8 +643,7 @@ public static void putObject_test11() throws Exception {
    * Test: putObject(String bucketName, String objectName, InputStream stream, long size,
    *                 Map&lt;String, String&gt; headerMap) with Storage Class REDUCED_REDUNDANCY.
    */
-  @SuppressFBWarnings("UCF")
```
balamurugana: This should not be removed?

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -86,13 +84,9 @@
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
 import java.nio.file.StandardOpenOption;
-import java.security.InvalidAlgorithmParameterException;
 import java.security.InvalidKeyException;
```
balamurugana: This is not required.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -86,13 +84,9 @@
...
-import java.security.Key;
-import java.security.KeyPair;
 import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
```
balamurugana: Remove the above two imports.

api/src/main/java/io/minio/ErrorCode.java
nitisht: restored.

functional/FunctionalTest.java
nitisht: yes, restored

api/src/main/java/io/minio/MinioClient.java
nitisht: `InvalidKeyException` is used at other places as well, like https://github.com/minio/minio-java/blob/master/api/src/main/java/io/minio/Signer.java#L369

api/src/main/java/io/minio/MinioClient.java
nitisht: `NoSuchAlgorithmException` is used at other places as well, like https://github.com/minio/minio-java/blob/master/api/src/main/java/io/minio/Signer.java#L369

PR #671, opened 2018-04-05
Ask findbugs to be more picky.

tobias-:
This commit basically asks findbugs to be more picky. Without this, e.g. string comparison with == isn't found.

The following is the other warnings that popped up in the API project when this is changed. Without looking into them, all of these except the performance ones and maybe the serialVersionUID ones are potentially bad.

### Bad practice Warnings

Code | Warning
-- | --
ES | Comparison of String parameter using == or != in io.minio.policy.Resources.matched(String, String)
Se | Class io.minio.errors.ErrorResponseException defines non-transient non-serializable instance field response
Se | Class io.minio.errors.NoResponseException defines non-transient non-serializable instance field request
Se | Class io.minio.errors.NoSuchBucketPolicyException defines non-transient non-serializable instance field bucketPolicy
SnVI | io.minio.policy.ConditionKeyMap is Serializable; consider declaring a serialVersionUID
SnVI | io.minio.policy.ConditionMap is Serializable; consider declaring a serialVersionUID
SnVI | io.minio.policy.Resources is Serializable; consider declaring a serialVersionUID

### Internationalization Warnings

Code | Warning
-- | --
Dm | Use of non-localized String.toUpperCase() or String.toLowerCase() in io.minio.Digest.sha256Hash(Object, int)
Dm | Use of non-localized String.toUpperCase() or String.toLowerCase() in io.minio.Digest.sha256Hash(String)
Dm | Use of non-localized String.toUpperCase() or String.toLowerCase() in io.minio.Digest.sha256Md5Hashes(Object, int)
Dm | Use of non-localized String.toUpperCase() or String.toLowerCase() in io.minio.http.HeaderParser.set(Headers, Object)
Dm | Use of non-localized String.toUpperCase() or String.toLowerCase() in  io.minio.MinioClient.execute(Method, String, String, String, Map, Map,  Object, int)
Dm | Use of non-localized String.toUpperCase() or String.toLowerCase() in io.minio.Signer.setCanonicalHeaders()
Dm | Use of non-localized String.toUpperCase() or String.toLowerCase() in io.minio.Signer.setSignature()

### Performance Warnings

Code | Warning
-- | --
SIC | The class io.minio.MinioClient$1 could be refactored into a named _static_ inner class
SIC | The class io.minio.MinioClient$2 could be refactored into a named _static_ inner class
SIC | The class io.minio.MinioClient$3 could be refactored into a named _static_ inner class

### Dodgy code Warnings

Code | Warning
-- | --
NP | Load of known null value in io.minio.MinioClient.updateRegionCache(String)
PZLA | Should io.minio.org.apache.commons.validator.routines.RegexValidator.match(String) return a zero length array rather than null?
UwF | ResponseHeader.date not initialized in constructor and dereferenced in io.minio.ResponseHeader.date()
UwF | ResponseHeader.lastModified not initialized in constructor and dereferenced in io.minio.ResponseHeader.lastModified()


balamurugana:
@tobias- Yep. Having strict `findbugs` is better to have.  It would be great if this PR comes with fixes of those errors.  Otherwise someone has to fix it before merging this PR else all PR fails in CI (like this PR)

tobias-:
@balamurugana It became quite a few changes, but now the PR builds cleanly with very picky FindBugs settings. Most changes seems to be in the tests, and most were just bad input stream closing

api/src/main/java/io/minio/MinioRequestBody.java
```diff
@@ -0,0 +1,66 @@
+package io.minio;
```
harshavardhana: Add the license header just like other files... @tobias-

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1178,19 +1178,18 @@ private void updateRegionCache(String bucketName)
...
+          }
           xpp.next();
-          location = getText(xpp, location);
-          break;
         }
-        xpp.next();
       }
 
-      // close response body.
-      response.body().close();
```
harshavardhana: Is the body being closed automatically here?

examples/PutObjectUiProgressBar.java
```diff
@@ -86,17 +86,14 @@ private void uploadFile(String fileName) throws IOException, NoSuchAlgorithmExce
...
-      bis = new BufferedInputStream(new FileInputStream(file));
+    try (BufferedInputStream bis = new BufferedInputStream(new FileInputStream(file))) {
       ProgressMonitorInputStream pmis = new ProgressMonitorInputStream(
           this,
           "Uploading... " + file.getAbsolutePath(),
           bis);
 
       pmis.getProgressMonitor().setMillisToPopup(10);
       minioClient.putObject("bank", "my-objectname", pmis, bis.available(), "application/octet-stream");
-      pmis.close();
```
harshavardhana: So assuming if this is the behavior in Java 7 close is automatically called.

api/src/main/java/io/minio/MinioClient.java
tobias-: Yes, it's in the enclosed try-using block. Please see https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

examples/PutObjectUiProgressBar.java
tobias-: Not java 7 behaviour, but as the link above describes, it closes what's declared in the try header

api/src/main/java/io/minio/MinioRequestBody.java
tobias-: Done.

harshavardhana:
@tobias- can you fix the styling issues? 

```
[ant:checkstyle] /home/travis/build/minio/minio-java/api/src/main/java/io/minio/MinioClient.java:2078: 'try' child have incorrect indentation level 8, expected level should be 6.
[ant:checkstyle] /home/travis/build/minio/minio-java/api/src/main/java/io/minio/MinioClient.java:2078:1: Line contains a tab character.
[ant:checkstyle] /home/travis/build/minio/minio-java/api/src/main/java/io/minio/MinioRequestBody.java:29: First sentence should be present.
:api:checkstyleMain FAILED
```

tobias-:
Ok, I think I fixed all of the problems in the merge conflict as well. I also removed the unused private method that became disused in the merge from master (which made findbugs complain)

api/src/main/java/io/minio/MinioRequestBody.java
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * RequestBody that wraps a single data object.
+ */
+class MinioRequestBody extends RequestBody {
+  private final String contentType;
+  private final Object data;
+  private final int len;
+
+  MinioRequestBody(final String contentType, final Object data, final int len) {
```
balamurugana: Could you change class name to `HttpRequestBody` like `HttpResponse`?

api/src/main/java/io/minio/errors/BucketPolicyTooLargeException.java
```diff
@@ -21,8 +21,9 @@
  */
 @SuppressWarnings({"WeakerAccess", "unused"})
 public class BucketPolicyTooLargeException extends MinioException {
-  private final String bucketName;
+  private static final long serialVersionUID = 6435208839560441168L;
```
balamurugana: could you tell me about his declaration?

tobias-: Exceptions are all Serializable, so they should include an explicit version. Actual findbugs error is  "SnVI".

References:
http://findbugs.sourceforge.net/bugDescriptions.html#SE_NO_SERIALVERSIONID
http://lmgtfy.com/?q=serialVersionUID
http://lmgtfy.com/?q=serialVersionUID+findbugs

balamurugana: With the reference https://stackoverflow.com/questions/285793/what-is-a-serialversionuid-and-why-should-i-use-it, I think we should ignore the warning.  Similarly it looks odd to define only to this exception.

deekoder:
Hi @tobias- are you able to address @balamurugana 's comments?

deekoder:
We would love to merge your PRs after these change requests are discussed.

api/src/main/java/io/minio/MinioRequestBody.java
tobias-: Done

api/src/main/java/io/minio/errors/BucketPolicyTooLargeException.java
tobias-: Ok, I definitely agree that they should be everywhere or nowhere. 

It's not a extremely important to use manually set serialVersionUID as it's more likely that they won't be updated when a serializable class changes than that they cause unnecessary. For some reason I don't get findbugs errors on this one anymore, and I never got on the other classes.

tobias-:
@deekoder @balamurugana 
I believe this should solve the issues the PR had.

Issue #680, opened 2018-06-01
Wrong signature calculation when 80 port is used with https


kullanici0606:
Hi,
We have a weird requirement that we have to run Minio server on 80 port with https enabled. However with this configuration, getPresignedObjectUrl calculates the signature wrong and clients fail to upload objects to Minio. It seems that there is a minor bug in the host header calculation:
https://github.com/minio/minio-java/blob/486c8b202befa20a1266951119dfdc504e26a35a/api/src/main/java/io/minio/MinioClient.java#L973-L977

First if condition should be 
```java
    if ((url.scheme().equals("http") && url.port() == 80) || (url.scheme().equals("https") && url.port() == 443)) {
```
or
```java
if ((!url.isHttps() && url.port() == 80) || (url.isHttps() && url.port() == 443)) {
```
since port is omitted from header only when http is used with 80 or https is used with 443 port. I mean port and protocol must match.

How to generate:
Generation of presigned url (keys and domain name are dummy, only generated for issue):
```java
MinioClient client = new MinioClient("https://test.example.com:80", "OHE23YP31RGLNYS0I7H2", "4scDD2tcNsDXkW3ab0cdsPlCd/1Xtz6evwcCE2ST");
Map<String, String> params = new HashMap<>();
params.put("Content-Type", "application/octet-stream");
String url = client.getPresignedObjectUrl(Method.PUT, "test", "123123", 60 * 60, params);
System.out.println(url);
```

Generated url:
```
https://test.example.com:80/test/123123?Content-Type=application%2Foctet-stream&X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=OHE23YP31RGLNYS0I7H2%2F20180601%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20180601T183640Z&X-Amz-Expires=3600&X-Amz-SignedHeaders=host&X-Amz-Signature=5cb9a5406aa9d5e237e6be6223043aa6dfd65e618808b4aaafc66e7d95151996
```

Curl command to run:
```
[...]

PR #681, opened 2018-06-01
Fix issue #680: Port from host header should be removed only protocol…

kullanici0606:
… matches

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -970,7 +970,7 @@ public class MinioClient {
     if (md5Hash != null) {
       requestBuilder.header("Content-MD5", md5Hash);
     }
-    if (url.port() == 80 || url.port() == 443) {
+    if (this.shouldOmitPortInHostHeader(url)) {
       requestBuilder.header("Host", url.host());
     } else {
       requestBuilder.header("Host", url.host() + ":" + url.port());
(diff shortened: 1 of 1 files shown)


kannappanr:
@kullanici0606 Can you please fix the formatting issues?

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1049,6 +1049,17 @@ public void writeTo(BufferedSink sink) throws IOException {
...
+  
+  /**
+   * Checks whether port should be omitted in Host header
+   * 
+   * HTTP Spec (rfc2616) defines that port should be omitted in Host header
+   * when port and service matches (i.e HTTP -> 80, HTTPS -> 443)
+   * @param url Url object
+   */
+  private boolean shouldOmitPortInHostHeader(HttpUrl url) {
+	  return (url.scheme().equals("http") && url.port() == 80) || (url.scheme().equals("https") && url.port() == 443);
```
balamurugana: Wondering someone is using non-standard port for TLS and non-TLS i.e. for TLS port 80 is used and for non-TLS port 443 is used.

Is this your problem or just making spec compatible?

kullanici0606: I also have this problem but the change is making the code spec compatible.

kullanici0606:
I fixed the formatting issues, could you please review it?

api/src/main/java/io/minio/MinioClient.java
balamurugana: Got it.

PR #698, opened 2018-08-19
Updated minio-java documents with examples for SSE.

aarushiarya:
Added full examples for server side encryption in minio-java documents.


deekoder:
@aarushiarya will you be able to fix this?

PR #707, opened 2018-10-02
fix: allow range headers to be set in getObject request when offset is 0 (#706)

### LINKED ISSUE #706: Unable to get first X bytes of object
I am unable to get the first X bytes of object via getObject(String bucketName, String objectName, long offset, Long length). Specifying a 0 offset and X length would return an entire file rather than bytes from 0 - X-1

Directly Quoting RFC 2616 Page 138, Regarding Byte range.

> The first 500 bytes (byte offsets 0-499, inclusive):  bytes=0-499

Further inspection of the source code, I found that if offset is less than or equal to 0, range header would get ignored.
Screenshot of source code from `MinioClient.java` Version 5.0.1
screen shot 2018-10-01 at 02 17 46

christopher-fredregill:
fix: allow range headers to be set in getObject request when offset is 0 (#706)

fixes #706

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -1620,12 +1620,10 @@ public class MinioClient {
     }
 
     Map<String,String> headerMap = new HashMap<>();
-    if (offset > 0) {
-      if (length != null) {
-        headerMap.put("Range", "bytes=" + offset + "-" + (offset + length - 1));
-      } else {
-        headerMap.put("Range", "bytes=" + offset + "-");
-      }
+    if (length != null) {
+      headerMap.put("Range", "bytes=" + offset + "-" + (offset + length - 1));
+    } else if (offset > 0) {
+      headerMap.put("Range", "bytes=" + offset + "-");
     }
 
     HttpResponse response = executeGet(bucketName, objectName, headerMap, null);
--- a/functional/FunctionalTest.java
+++ b/functional/FunctionalTest.java
@@ -753,6 +753,8 @@ public class FunctionalTest {
         }
       }
 
+      client.removeObject(bucketName, objectName);
+
       mintSuccessLog("putObject(String bucketName, String objectName, InputStream stream, "
                    + "long size, Map<String, String> headerMap)",
                   "size: 13 MB", startTime);
@@ -1183,6 +1185,65 @@ public class FunctionalTest {
       mintFailedLog("getObject(String bucketName, String objectName, ServerSideEncryption sse)"
           + " using SSE_C.",null, startTime, null,
                     e.toString() + " >>> " + Arrays.toString(e.getStackTrace()));
+      throw e;
+    }
+  }
+
+  /**
+   * Test: getObject(String bucketName, String objectName, long offset, Long length) with offset=0.
+   */
+  public static void getObject_test8() throws Exception {
+    if (!mintEnv) {
+      System.out.println(
+          "Test: getObject(String bucketName, String objectName, long offset, Long length) with offset=0"
+      );
+    }
+
+    final long startTime = System.currentTimeMillis();
+    final int fullLength = 1024;
+    final int partialLength = 256;
+    final long offset = 0L;
+    final String objectName = getRandomName();
+    try {
+      try (final InputStream is = new ContentInputStream(fullLength)) {
+        client.putObject(bucketName, objectName, is, fullLength, nullContentType);
+      }
+
+      try (
+          final InputStream partialObjectStream = client.getObject(
+              bucketName,
+              objectName,
+              offset,
+              Long.valueOf(partialLength)
+          )
+      ) {
+        byte[] result = new byte[fullLength];
+        final int read = partialObjectStream.read(result);
+        result = Arrays.copyOf(result, read);
+        if (result.length != partialLength) {
+          throw new Exception(
+              String.format(
+                  "Expecting only the first %d bytes from partial getObject request; received %d bytes instead.",
+                  partialLength,
+                  read
+              )
+          );
+        }
+      }
+      client.removeObject(bucketName, objectName);
+      mintSuccessLog(
+          "getObject(String bucketName, String objectName, long offset, Long length) with offset=0",
+          String.format("offset: %d, length: %d bytes", offset, partialLength),
+          startTime
+      );
+    } catch (final Exception e) {
+      mintFailedLog(
+          "getObject(String bucketName, String objectName, long offset, Long length) with offset=0",
+          String.format("offset: %d, length: %d bytes", offset, partialLength),
+          startTime,
+          null,
+          e.toString() + " >>> " + Arrays.toString(e.getStackTrace())
+      );
       throw e;
     }
   }
@@ -2708,6 +2769,7 @@ public class FunctionalTest {
     getObject_test4();
     getObject_test5();
     getObject_test6();
+    getObject_test8();
 
     listObject_test1();
     listObject_test2();


christopher-fredregill:
Some notes for anyone reviewing this:

1. The test that's failing in Travis does not fail when run against a minio host on RELEASE.2018-08-02T23-11-36Z; it seems like there is some sort of regression in whatever version of minio is currently backing the https://play.minio.io:9000 endpoint.

Here's a docker-compose snippet you can use to spin up a local test endpoint against which the full functional test suite (on my feature branch) will pass:

```
version: '3.7'
services:
  minio:
    image: minio/minio:RELEASE.2018-08-02T23-11-36Z
    command: server /data
    ports:
      - 9000:9000
    volumes:
      - ./data/:/data/
```

2. Please see https://github.com/minio/minio-java/blob/master/api/src/main/java/io/minio/MinioClient.java#L1627 in the original implementation. When the length is `null`, but the offset is valid, something like `Range: bytes=X-` (where `X` is the user-supplied offset) gets added to the request headers. This always results in a `416/The requested range is not satisfiable` error response, in the test suite. You'll notice that, as a workaround, I've placed the failing test in a try/catch block, but I am wondering if there is any reason why we should not simply require that a valid length be given. Please advise as to how best to handle this.

3. None of the test cases I found when I started looking into this problem were actually capable of detecting the error, so I modified one of the functional test cases to make some stronger assertions about the response from a partial getObject request. I'm open to other suggestions around how to verify the functionality in question.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1620,10 +1620,11 @@ public InputStream getObject(String bucketName, String objectName, long offset,
     }
 
     Map<String,String> headerMap = new HashMap<>();
-    if (offset > 0) {
+    if (offset >= 0) {
       if (length != null) {
         headerMap.put("Range", "bytes=" + offset + "-" + (offset + length - 1));
       } else {
+        // results in a 416 with response message "The requested range is not satisfiable"
```
balamurugana: This comment is not needed

functional/FunctionalTest.java
```diff
@@ -1033,12 +1033,24 @@ public static void getObject_test3() throws Exception {
     long startTime = System.currentTimeMillis();
     try {
       String objectName = getRandomName();
-      try (final InputStream is = new ContentInputStream(3 * MB)) {
-        client.putObject(bucketName, objectName, is, 3 * MB, nullContentType);
+      try (final InputStream is = new ContentInputStream(1024)) {
```
balamurugana: Add new tests than changing existing ones.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1620,10 +1620,11 @@ public InputStream getObject(String bucketName, String objectName, long offset,
     }
 
     Map<String,String> headerMap = new HashMap<>();
```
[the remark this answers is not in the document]
balamurugana: getObject() in minio-java does accept offset and length not range i.e. not start/end values.  With this existing fix is correct.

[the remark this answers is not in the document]
harshavardhana: There are three allowed behaviors

- bytes=N-M
- bytes=N-
- bytes=-M 

In minio-java the last one is not supported this is what I meant by complete implementation @balamurugana

functional/FunctionalTest.java
christopher-fredregill: I've moved the new test case (with the assertions about the bytes that should be received from the partial getObject request) into `getObject_test8` in the functional suite. I had to make minor updates to two other functional tests that were leaving behind objects in the shared bucket and causing `listObject_test5` to fail in error; those other tests should now be cleaning up after themselves.

christopher-fredregill:
@harshavardhana @balamurugana can you please take another look?

functional/FunctionalTest.java
```diff
@@ -1125,8 +1127,14 @@ public static void getObject_test6() throws Exception {
         client.putObject(bucketName, objectName, is, 0, nullContentType);
       }
 
-      client.getObject(bucketName, objectName)
-          .close();
+      try {
+        client.getObject(bucketName, objectName).close();
+      } catch (final ErrorResponseException e) {
+        // current implementation results in request headers="Range: bytes=0-"; response is a 416 with the given message
```
harshavardhana: We can remove this comment, this is an expected response.

christopher-fredregill: removed

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1620,12 +1620,10 @@ public InputStream getObject(String bucketName, String objectName, long offset,
...
-    if (offset > 0) {
-      if (length != null) {
-        headerMap.put("Range", "bytes=" + offset + "-" + (offset + length - 1));
-      } else {
-        headerMap.put("Range", "bytes=" + offset + "-");
-      }
+    if (length != null) {
```
balamurugana: `if (offset > 0)` has to be there else wrong range will be sent for negative values.

functional/FunctionalTest.java
```diff
@@ -1125,8 +1127,13 @@ public static void getObject_test6() throws Exception {
...
 
-      client.getObject(bucketName, objectName)
-          .close();
+      try {
+        client.getObject(bucketName, objectName).close();
+      } catch (final ErrorResponseException e) {
+        if (!e.toString().contains("message=The requested range is not satisfiable")) {
+          throw e;
+        }
+      }
```
balamurugana: Unable to understand this change.  Why is it required for this test?

functional/FunctionalTest.java
```diff
@@ -1187,6 +1194,63 @@ public static void getObject_test7() throws Exception {
     }
   }
 
+  /**
+   * Test: getObject(String bucketName, String objectName, long offset, Long length).
+   */
+  public static void getObject_test8() throws Exception {
```
balamurugana: You would need to add specific to `offset=0`.  `getObject_test2()` and getObject_test3()` are covering other cases.  Additionally keep only one test in test method.

api/src/main/java/io/minio/MinioClient.java
christopher-fredregill: The linter was complaining in CI because this check is actually unnecessary; if you look a few lines up, there is already a check for the case when offset < 0. I had to remove the duplicate check here in order to satisfy the linter.

functional/FunctionalTest.java
christopher-fredregill: The original test catches and re-throws any exception it encounters.

I modified the behavior of the client, such that a known 416 exception will always be encountered here.

The test now has to know in advance that it will see this kind of exception and accept it, or else it will simply fail.

This is why I included comments explaining this in both the client code and in the fictional test code.

Some sort of change is necessary here... maybe you can propose an alternative solution, if this doesn’t suit you.

harshavardhana: The problem is we send offset requests by default now - after this PR change even for empty objects which is not allowed in S3 

This exception check is needed, or we make sure in this test Range is not set for empty objects.

api/src/main/java/io/minio/MinioClient.java
balamurugana: got it

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1620,12 +1620,10 @@ public InputStream getObject(String bucketName, String objectName, long offset,
...
+      headerMap.put("Range", "bytes=" + offset + "-" + (offset + length - 1));
+    } else {
+      headerMap.put("Range", "bytes=" + offset + "-");
```
balamurugana: To be clean you would need to set `Range` for non-zero offset

christopher-fredregill: Not sure what has to change, specifically; could you clarify what you'd like the `Range` to look like, in each of the scenarios under consideration? (`offset`=0 vs. `offset` > 0, with and without positive `length` argument, etc.).

balamurugana: You would need to do below;
```java
if (length != null) {
  headerMap.put("Range", "bytes=" + offset + "-" + (offset + length -1));
  // Here offset is set anyway to satisfy range i.e. `Range: bytes=0-100` or `Range: bytes=10-100`
} else if offset > 0 {
  headerMap.put("Range", "bytes=" + offset + "-");
  // Here `Range` header is set only if offset > 0 i.e. for offset=10; `Range: bytes=10-`; for offset=0; nothing is set.
}
```

christopher-fredregill: Updated the client and test code as per ^. Thanks.

harshavardhana:
We could add a test to generate the 416 range not satisfiable with offset=1 for a zero byte object @christopher-fredregill upto you though. This PR is okay to be merged.

PR #709, opened 2018-10-03
GetObject to N bytes when offset is 0

sinhaashish:
GetObject was returning the entire file when Offset was 0.  Specifying a 0 offset and X length would return an entire file rather than bytes from 0 - X-1. It was working correctly when the Offset was greater than 0. Issue #706


harshavardhana:
@sinhaashish there is a more comprehensive PR already submitted https://github.com/minio/minio-java/pull/707 , this PR can be closed.

sinhaashish:
Closing

PR #722, opened 2018-11-02
Enabled API for encrypted putObject with metadata

### LINKED ISSUE #705: Support for metadata in putObject
One of the requests that we got at the conference here was to support metadata for putObject (just like how we support it in minio-go)

Praveenrajmani:
minio-java already had provision for metadata support for PUT calls. Which were private.

Fixes #705

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -3294,6 +3294,91 @@ public class MinioClient {
 
 
   /**
+   * Uploads data from given stream as object to given bucket with specified meta data 
+   * and encrypt the stream with a sse key.
+
+   * <p>
+   * If the object is larger than 5MB, the client will automatically use a multipart session. 
+   * </p>
+   * <p>
+   * If the session fails, the user may attempt to re-upload the object by attempting to create
+   * the exact same object again. The client will examine all parts of any current upload session
+   * and attempt to reuse the session automatically. If a mismatch is discovered, the upload will fail
+   * before uploading any more data. Otherwise, it will resume uploading where the session left off.
+   * </p>
+   * <p>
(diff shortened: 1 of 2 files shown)


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3293,6 +3293,89 @@ public void putObject(String bucketName, String objectName, InputStream stream,
   }
 
 
+  /**
+   * Uploads data from given stream as object to given bucket where the stream size is unknown.
```
balamurugana: Description is wrong to this method.

deekoder:
ping @Praveenrajmani

api/src/main/java/io/minio/MinioClient.java
Praveenrajmani: description modified :+1:

balamurugana: Is this description "Uploads data from given stream as object to given bucket where the stream size is unknown." correct for this version of put object?

Praveenrajmani: No .. changed the `unknown` to `known` :+1:

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3293,6 +3293,92 @@ public void putObject(String bucketName, String objectName, InputStream stream,
   }
 
 
+  /**
+   * Uploads data from given stream as object to given bucket where the stream size is known.
```
balamurugana: Use http://minio.github.io/minio-java/io/minio/MinioClient.html#putObject-java.lang.String-java.lang.String-java.io.InputStream-long-java.util.Map- description as base and modify as per this method

PR #727, opened 2018-11-13
Add support of listen bucket notification

vadmeste:
This PR adds support of listen bucket notification

```java
  try {
    class TestBucketListener implements BucketEventListener {
      @Override
      public void updateEvent(NotificationInfo info) {
        System.out.println(info.records[0].s3.bucket.name + "/"
           + info.records[0].s3.object.key + " has been created");
      }
    }
     minioClient.listenBucketNotification("testbucket", "prefix", "suffix",
        new String[]{"s3:ObjectCreated:*", "s3:ObjectAccessed:*"}, new TestBucketListener());
  } catch (Exception e) {
    System.out.println("Error occurred: " + e);
  }
```


vadmeste:
Waiting for API approval

balamurugana:
> Waiting for API approval

Marking as `Blocked` till its resolved.

deekoder:
Why is this blocked? Not clear.

vadmeste:
> Why is this blocked? Please, can you add exactly what we need to discuss?

Since this is a new API, I was seeking AB to approve it

kannappanr:
The API has been approved, removing the blocked label. This PR can now be reviewed.

functional/FunctionalTest.java
```diff
@@ -2779,6 +2780,74 @@ public static void removeAllBucketNotification_test1() throws Exception {
...
+        public void run() {
+          String[] events = {"s3:ObjectCreated:*", "s3:ObjectAccessed:*"};
+          try {
+            client.listenBucketNotification(bucketName, "prefix", "suffix", events, bl);
+          } catch (Exception e) {
+            System.out.println(e);
+          }
+        }}).start();
+
+      String file1kb = createFile(1024);
```
sinhaashish: ./gradlew build
Indentation issue at line 2821 `String file1kb = createFile(1024);` 
> Task :api:checkstyleMain FAILED
[ant:checkstyle] /home/ashish/go/src/github.com/minio/minio-java/functional/FunctionalTest.java:2822: 'member def type' have incorrect indentation level 7, expected level should be 6.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -815,8 +819,8 @@ public boolean verify(String hostname, SSLSession session) {
    * @param length         Length of HTTP request body.
    */
   private Request createRequest(Method method, String bucketName, String objectName,
-                                String region, Map<String,String> headerMap,
-                                Map<String,String> queryParamMap, final String contentType,
+                                String region, Map<String,String[]> headerMap,
+                                Map<String,String[]> queryParamMap, final String contentType,
```
balamurugana: Can we use `Multimap` here than `Map<String,String[]>` ?
```java
import com.google.common.collect.TreeMultimap;
import com.google.common.collect.Multimap;

public class Hello {
  public static void main(String[] args) throws Exception {
    Multimap<String, String> map = TreeMultimap.create();
    map.put("key1", "value3");
    map.put("key1", "value1");
    map.put("key1", "value2");
    // prints [value1, value2, value3]
    System.out.println(map.get("key1"));
  }
}
```

examples/ListenBucketNotification.java
```diff
@@ -0,0 +1,58 @@
...
+    try {
+      /* play.minio.io for test and development. */
+      MinioClient minioClient = new MinioClient("https://play.minio.io:9000", "Q3AM3UQ867SPQQA43P2F",
+                                                "zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG");
+
+      /* Amazon S3: */
+      // MinioClient minioClient = new MinioClient("https://s3.amazonaws.com", "YOUR-ACCESSKEYID",
+      //                                           "YOUR-SECRETACCESSKEY");
+
+      class TestBucketListener implements BucketEventListener {
```
sinhaashish: Class 'TestBucketListener' must either be declared abstract or implement abstract method 'updateError(Exception)' in 'BucketEventListener'.

api/src/main/java/io/minio/Signer.java
```diff
@@ -144,7 +148,7 @@ private void setCanonicalHeaders() {
 
 
   private void setCanonicalQueryString() {
-    Map<String,String> signedQueryParams = new TreeMap<>();
+    Map<String,ArrayList<String>> signedQueryParams = new TreeMap<>();
```
balamurugana: We can use `Multimap` here.
```java
    String encodedQuery = this.url.encodedQuery();
    if (encodedQuery == null) {
      this.canonicalQueryString = "";
      return;
    }

    Multimap<String, String> signedQueryParams = TreeMultimap.create();

    for (String queryParam : encodedQuery.split("&")) {
      String[] tokens = queryParam.split("=");
      if (tokens.length > 1) {
        signedQueryParams.put(tokens[0], tokens[1]);
      } else {
        signedQueryParams.put(tokens[0], "");
      }
    }

    this.canonicalQueryString = Joiner.on("&").withKeyValueSeparator("=").join(signedQueryParams.entries());
```

docs/API.md
```diff
@@ -553,6 +554,50 @@ try {
 }
 ```
 
+<a name="listenBucketNotification"></a>
+### listenBucketNotification(String bucketName, String prefix, String suffix, String[] events, BucketEventListener listener)
+`public void listenBucketNotification(String bucketName, String prefix, String suffix, String[] events, BucketEventListener listener)`
```
balamurugana: My choice here is like `public Iterable<Result<Event>> listenBucketNotification(String bucketName, String prefix, String suffix, EventType[] events)` .  Its just a thought.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -993,14 +1001,40 @@ private boolean shouldOmitPortInHostHeader(HttpUrl url) {
...
+
+    Map<String, String[]> headerMap = null;
+    if (simpleHeaderMap != null) {
+      headerMap = new HashMap<>();
+      for (Map.Entry<String, String> h: simpleHeaderMap.entrySet()) {
+        headerMap.put(h.getKey(), new String[]{h.getValue()});
+      }
+    }
+
+    return executeReq(method, region, bucketName, objectName, headerMap, queryParamMap, body, length);
```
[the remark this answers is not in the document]
vadmeste: > Can we use method overloading here?

```
minio-java/api/src/main/java/io/minio/MinioClient.java:1029: error: name clash: execute(Method,String,String,String,Map<String,String[]>,Map<String,String[]>,Object,int) and execute(Method,String,String,String,Map<String,String>,Map<String,String>,Object,int) have the same erasure
  private HttpResponse execute(Method method, String region, String bucketName, String objectName,
```

functional/FunctionalTest.java
vadmeste: > ./gradlew build

Fixed.

api/src/main/java/io/minio/MinioClient.java
vadmeste: Fixed, but I decided to use HashMultimap instead of TreeMultimap (since we don't need to order keys and values)

api/src/main/java/io/minio/Signer.java
vadmeste: Good idea! Fixed

docs/API.md
vadmeste: > My choice here is like `public Iterable<Result<Event>> listenBucketNotification(String bucketName, String prefix, String suffix, EventType[] events)` . Its just a thought.

One problem with Iterable is that it doesn't have close/free resources concept which means we cannot close the underlying resources (opened connection to the server)

vadmeste:
All comments addressed/responded @balamurugana @sinhaashish, would you please continue to review ?

api/src/main/java/io/minio/MinioClient.java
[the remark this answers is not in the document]
balamurugana: As `Multimap` is used now, do you still see the error for method overloading?

api/src/main/java/io/minio/Signer.java
```diff
@@ -142,16 +145,15 @@ private void setCanonicalHeaders() {
...
   private void setCanonicalQueryString() {
-    Map<String,String> signedQueryParams = new TreeMap<>();
-
     String encodedQuery = this.url.encodedQuery();
     if (encodedQuery == null) {
       this.canonicalQueryString = "";
       return;
     }
 
+    Multimap<String, String> signedQueryParams = MultimapBuilder.treeKeys().arrayListValues().build();
```
balamurugana: Just curious to know the difference with `Multimap<String, String> signedQueryParams = TreeMultimap.create();`

api/src/main/java/io/minio/MinioClient.java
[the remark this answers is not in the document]
vadmeste: > As `Multimap` is used now, do you still see the error for method overloading?

Yes, this is the new error:
```
/home/vadmeste/work/java/minio-java/api/src/main/java/io/minio/MinioClient.java:1267: error: reference to execute is ambiguous
    HttpResponse response = execute(Method.HEAD, getRegion(bucketName), bucketName, objectName, null,
                            ^
  both method execute(Method,String,String,String,Map<String,String>,Map<String,String>,Object,int) in MinioClient and method execute(Method,String,String,String,Multimap<String,String>,Multimap<String,String>,Object,int) in MinioClient match
1 error
```

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -993,14 +999,40 @@ private boolean shouldOmitPortInHostHeader(HttpUrl url) {
...
+                               Map<String,String> simpleHeaderMap, Map<String,String> simpleQueryParamMap,
+                               Object body, int length)
+    throws InvalidBucketNameException, NoSuchAlgorithmException, InsufficientDataException, IOException,
+           InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
+           InternalException {
+
+    Multimap<String, String> queryParamMap = null;
+    if (simpleQueryParamMap != null) {
+      queryParamMap = HashMultimap.create();
+      for (Map.Entry<String, String> m: simpleQueryParamMap.entrySet()) {
```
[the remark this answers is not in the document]
vadmeste: > You could use [Multimaps.forMap(map)](https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/Multimaps.html#forMap(java.util.Map)) than having own `for` loop.

Done

api/src/main/java/io/minio/Signer.java
vadmeste: > Just curious to know the difference with `Multimap<String, String> signedQueryParams = TreeMultimap.create();` Even though it works like this, `TreeMultimap.create()` is quite straight forward.

TreeMultimap orders keys & values, meaning if one key has several values, values will also be ordered.

Though this what S3 spec says, Minio server doesn't support it, so I need to send other PRs in Minio server/minio-go/minio-java and merge them at the same time and release all of them together to fix this.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -993,14 +999,40 @@ private boolean shouldOmitPortInHostHeader(HttpUrl url) {
...
+                               Multimap<String,String> headerMap, Multimap<String,String> queryParamMap,
                                Object body, int length)
     throws InvalidBucketNameException, NoSuchAlgorithmException, InsufficientDataException, IOException,
            InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
            InternalException {
     String contentType = null;
-    if (headerMap != null) {
-      contentType = headerMap.get("Content-Type");
+    if (headerMap != null && headerMap.get("Content-Type") != null) {
+      contentType = String.join(" ", headerMap.get("Content-Type"));
```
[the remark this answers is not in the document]
vadmeste: > As this is an `private` method, it would be clean to have `HashMultimap` to avoid `null` value check like here.

I would prefer this additional null check in case when someone does more tweaks in the code later.

api/src/main/java/io/minio/Signer.java
balamurugana: This means it will fail working with AWS S3 till it's fixed correctly.  IMO minio server issue needs to be handled separately than fixing minio-java to work with minio server's issue.

api/src/main/java/io/minio/MinioClient.java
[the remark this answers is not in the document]
balamurugana: IMO `private` methods and their usage need to be well controlled.  Its just a thought.

api/src/main/java/io/minio/Signer.java
vadmeste: > This means it will fail working with AWS S3 till it's fixed correctly. IMO minio server issue needs to be handled separately than fixing minio-java to work with minio server's issue.

This is a specific Minio feature, AWS S3 doesn't support it. Then fixing Minio server will make codes using minio-go deployment fail, that's why I wanted more coordination about this

balamurugana: My point here is, it is Signer where the bug is introduced.  Sending single or multiple values to query parameters are higher level behaviours, but Signer needs to work correct for requests.

vadmeste: > My point here is, it is Signer where the bug is introduced. Sending single or multiple values to query parameters are higher level behaviours, but Signer needs to work correct for requests.

yes, I agree. I am just saying it is a TODO thing (we are actually not making any regressions here) and I prefer if this is done with some coordination to avoid having users complaining.

PR #735, opened 2018-12-06
Uploads file as object and encrypt with sse key

sinhaashish:
New Method introduced 
```putObject(String bucketName, String objectName, ServerSideEncryption sse, String fileName)```
which uploads given file as object in given bucket and encrypt with a sse key.

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -3413,7 +3413,75 @@ public class MinioClient {
     putObject(bucketName, objectName, size, new BufferedInputStream(stream), headerMap, sse);
   }
 
-  
+
+
+  /**
+   * Uploads given file as object in given bucket and encrypt with a sse key.
+   * <p>
+   * If the object is larger than 5MB, the client will automatically use a multipart session.
+   * </p>
+   * <p>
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: api/src/main/java/io/minio/MinioClient.java [hunk 1]
```diff
@@ -3412,7 +3412,75 @@ public void putObject(String bucketName, String objectName, InputStream stream,
...
+   */
+
+  public void putObject(String bucketName, String objectName, ServerSideEncryption sse, String fileName)
+          throws InvalidBucketNameException, NoSuchAlgorithmException, IOException,
+          InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
+          InternalException, InvalidArgumentException, InsufficientDataException {
+
+    if ((sse.getType() == (ServerSideEncryption.Type.SSE_C)) && (!this.baseUrl.isHttps())) {
+      throw new InvalidArgumentException("SSE_C operations must be performed over a secure connection.");
+    } else if ((sse.getType() == (ServerSideEncryption.Type.SSE_KMS)) && (!this.baseUrl.isHttps())) {
```
balamurugana: Remove all unnecessary `()`

sinhaashish: Done.

PR #737, opened 2018-12-11
tests: Ignore NotImplemented error in listenBucketNotification

vadmeste:
Functional tests were just infinitely waiting for a put object
notification event, this commit will just ignore NotImplemented
error to support gateways which are not supporting bucket
notifications.

### DIFF
--- a/functional/FunctionalTest.java
+++ b/functional/FunctionalTest.java
@@ -2857,13 +2857,20 @@ public class FunctionalTest {
     }
 
     long startTime = System.currentTimeMillis();
+    String file = createFile1Mb();
+    String bucketName = getRandomName();
+
     try {
-      String bucketName = getRandomName();
       client.makeBucket(bucketName, region);
 
       class TestBucketListener implements BucketEventListener {
         int eventsReceived = 0;
+        Exception lastException = null;
         NotificationEvent firstEvent = null;
+
+        public void updateError(Exception e) {
+          lastException = e;
+        }
 
         @Override
         public void updateEvent(NotificationInfo info) {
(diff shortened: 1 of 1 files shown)


vadmeste:
@harshavardhana your comment is addressed, the test ignores not implemented error now.

(by the way, listen bucket notification is supported in gateway NAS)

harshavardhana:
I don't know which release you are using it should be disabled as per https://github.com/minio/minio/pull/6920

poornas:
it is defined incorrectly - should have been
```
func (n *nasObjects) IsListenBucketSupported() bool {
	return false
}
```

harshavardhana:
> it is defined incorrectly - should have been
> 
> ```
> func (n *nasObjects) IsListenBucketSupported() bool {
> 	return false
> }
> ```

Please fix it @poornas it is not expected to support ListenBucket with NAS gateway.

functional/FunctionalTest.java
```diff
@@ -2856,19 +2861,43 @@ public void run() {
...
 
+      // Fail for any exception but ignore NotImplemented error
+      if (bl.lastException != null) {
+        ErrorResponse errorResponse = null;
+        if (bl.lastException instanceof ErrorResponseException) {
+          ErrorResponseException exp = (ErrorResponseException) bl.lastException;
+          errorResponse = exp.errorResponse();
+        }
+        if (errorResponse != null && errorResponse.errorCode() == ErrorCode.NOT_IMPLEMENTED) {
+          mintIgnoredLog("listenBucketNotification(String bucketName)", null, startTime);
```
kannappanr: what about cleanup here?

functional/FunctionalTest.java
```diff
@@ -2856,19 +2861,43 @@ public void run() {
...
+          return;
+        } else {
+          mintFailedLog("listenBucketNotification(String bucketName)", null, startTime,
+              null, bl.lastException.toString() + " >>> " + Arrays.toString(bl.lastException.getStackTrace()));
```
kannappanr: Here also, created bucket and object will not be removed.

functional/FunctionalTest.java
```diff
@@ -2856,19 +2861,43 @@ public void run() {
...
+      for (int i = 0; i < 20; i++) {
         client.putObject(bucketName, "prefix-random-suffix", file1kb);
-        Thread.sleep(300);
+        Thread.sleep(500);
+        if (bl.eventsReceived > 0 || bl.lastException != null) {
+          break;
+        }
       }
 
+      // Fail for any exception but ignore NotImplemented error
```
kannappanr: Maybe the cleanup code of removing the bucket and object should be moved to this part before you start checking for errors.

vadmeste: > Maybe the cleanup code of removing the bucket and object should be moved here before you start checking for errors.

Nice catch! fixed

poornas:
@vadmeste, this needs rebase and build failures checked.

PR #739, opened 2018-12-18
Fix: Do not base64 encode md5Hash

### LINKED ISSUE #738: I call putObject() API, when there are incomplete parts, it still putObject() and gets the same etag again. I found my md5Hash and  the part.etag() look very different. Is there anything wrong?
I just use JAVA main function for testing(JAVA sdk).

Minio version: 5.0.4
image

The steps are:
1. I called "putObject()" to upload a 1G size file.
image
2. Interrupt the process after few seconds
3. Check my minio docker volume, I found the **/data/.minio.sys/multipart/** folder appear some files and the size about 500MB.
4.  Repeat the step1, I called "putObject()" to upload the same 1G size file again.
5.  Debug and trace the code, it gets `existingParts` successfully(https://github.com/minio/minio-java/blob/master/api/src/main/java/io/minio/MinioClient.java#L3693). But when the process goes to `if (md5Hash.equals(part.etag()))`(https://github.com/minio/minio-java/blob/master/api/src/main/java/io/minio/MinioClient.java#L3736), it always gets the negative result, so it uploads all of the parts again.

     I printed my **md5Hash** and the **part.etag()**, these two formats look very different(refer to the screenshot below). 
image

Please help, which step did I do wrong?

kannappanr:
Should not have to base64 encode the calculated md5Hash.

Fixes #738

### DIFF
--- a/api/src/main/java/io/minio/Digest.java
+++ b/api/src/main/java/io/minio/Digest.java
@@ -119,7 +119,17 @@ class Digest {
                                   + "please report this issue at https://github.com/minio/minio-java/issues");
     }
 
-    return BaseEncoding.base64().encode(md5Digest.digest());
+    byte[] hash = md5Digest.digest();
+    StringBuffer hexString = new StringBuffer();
+    for (int i = 0; i < hash.length; i++) {
+      if ((0xff & hash[i]) < 0x10) {
+        hexString.append("0"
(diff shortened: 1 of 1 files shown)


kannappanr:
Closing this PR as we have another PR(#740) to fix this issue.

PR #740, opened 2018-12-19
ResumeMultipart

sinhaashish:
For Resuming the multipart the Etag was compared with md5, which were never equal and
thus every time new upload was done. The check has been removed so that upload can
skip the already uploaded part

Resolves the [issue](https://github.com/minio/minio-java/issues/738) 

For S3
```The ETag value returned by S3 for objects uploaded using the multipart upload API is computed differently than for objects uploaded with PUT object, and does not represent the MD5 of the object data, though it still uniquely represents the object.```
[Refer](https://forums.aws.amazon.com/thread.jspa?messageID=293789)


sinhaashish:
Closing this PR as a consolidated [PR](https://github.com/minio/minio-java/pull/741) is raised.

PR #742, opened 2018-12-27
Add API to  Compose Object

sinhaashish:
Creates a new object by combining different Source Objects .

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -42,6 +42,7 @@ import io.minio.http.Scheme;
 import io.minio.messages.Bucket;
 import io.minio.messages.CompleteMultipartUpload;
 import io.minio.messages.CopyObjectResult;
+import io.minio.messages.CopyPartResult;
 import io.minio.messages.CreateBucketConfiguration;
 import io.minio.messages.DeleteError;
 import io.minio.messages.DeleteObject;
@@ -104,6 +105,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -2325,6 +2327,218 @@ public class MinioClient {
     }
   }
 
+
+  /**
+   * Create an object by concatenating a list of source objects using server-side copying.
+   * </p>
+   * <b>Example:</b><br>   *
+   * <pre>
+   * {@code minioClient.composeObject(String bucketName, String objectName, 
+   * List<ComposeSource> composeSource, Map<String,String> userMetaData,
+   * ServiceConfigurationError sse );}
+   * </pre>
+   *
+   * @param bucketName
+   *          Destination Bucket to be created upon compose.
+   * @param objectName
+   *          Destination Object to be created upon compose.
+   * @param sseTarget
+   *          Server Side Encryption.
+   * @param userMetaData
+   *          User Meta data.
+   * @param composeSources
+   *          List of Source Objects used to compose Object
+   *
+   * @throws InvalidBucketNameException  upon invalid bucket name is given
+   * @throws NoSuchAlgorithmException
+   *           upon requested algorithm was not found during signature calculation
+   * @throws InsufficientDataException  upon getting EOFException while reading given
+   *           InputStream even before reading given length
+   * @throws IOException                 upon connection error
+   * @throws InvalidKeyException
+   *           upon an invalid access key or secret key
+   * @throws NoResponseException         upon no response from server
+   * @throws XmlPullParserException      upon parsing response xml
+   * @throws ErrorResponseException      upon unsuccessful execution
+   * @throws InternalException           upon internal library error
+   * @throws InvalidArgumentException    upon invalid value is passed to a method.
+   * @throws InvalidResponseException    upon a non-xml response from server
+   */
+  public void  composeObject(String bucketName, String objectName, 
+      List<ComposeSource> composeSources, Map<String,String> userMetaData, 
+      ServerSideEncryption sseTarget)
+   throws InvalidBucketNameException, NoSuchAlgorithmException,
+    InsufficientDataException, IOException, InvalidKeyException, NoResponseException,
+    XmlPullParserException, ErrorResponseException, InternalException, 
+    InvalidArgumentException, InvalidResponseException {
+    if (bucketName == null) {
+      throw new InvalidArgumentException("Destination bucket name cannot be empty");
+    }
+    if (objectName == null) {
+      throw new InvalidArgumentException("Destination object name cannot be empty");
+    }
+
+    // Validate encryption for destination
+    if ( sseTarget != null ) {
+      if ((sseTarget.getType() == ServerSideEncryption.Type.SSE_C)
+          && (!this.baseUrl.isHttps())) {
+        throw new InvalidArgumentException(
+          "SSE_C operations must be performed over a secure connection.");
+      } else if ((sseTarget.getType() == (ServerSideEncryption.Type.SSE_KMS)) 
+          && (!this.baseUrl.isHttps())) {
+        throw new InvalidArgumentException(
+          "SSE_KMS operations must be performed over a secure connection.");
+      }
+    }
+
+    if (composeSources.isEmpty() || composeSources.size() > MAX_MULTIPART_COUNT ) {
+      throw new InvalidArgumentException("There must be as least one and up to " 
+       + "MAX_MULTIPART_COUNT source objects.");
+    }
+
+    int i = 0;
+    int totalParts = 0;
+    long totalSize = 0;
+    long[]  sourceSize =  new long[composeSources.size()];
+    for (ComposeSource source : composeSources) {
+
+      Map<String, String> sourceHeader = source.headers();
+      // Error out if client side encryption is used in this source object when
+      // more than one source objects are given.
+      if ((composeSources.size() > 1 ) 
+          && sourceHeader.containsKey("x-amz-meta-x-amz-key")) {
+        throw new InvalidArgumentException("Client side encryption is used in source" 
+        + " object" + source.bucketName() + source.objectName());
+      }
+      ObjectStat os = null;
+      //Validate encryption for source object
+      if (source.sseSource() != null ) {
+        os = statObject(source.bucketName(), source.objectName(), 
+        source.sseSource());
+      } else {
+        os = statObject(source.bucketName(), source.objectName());
+      }
+
+      long offset = source.offset() == null ? 0L :  source.offset();
+      long length = source.length() == null ? 0L :  source.length();
+      long sizeOfCopy = 0L;
+      if (offset == 0 && length == 0) {
+        sizeOfCopy = os.length();
+      } else if (offset != 0 && length == 0) {
+        sizeOfCopy = os.length() - offset;
+      } else if (length != 0) {
+        sizeOfCopy = length;
+      }
+      // Check if a segment is specified is within object bounds
+      if ( length > os.length()) {
+        throw new InvalidArgumentException("Source specified length " + length
+         + " is more than " + "size of Object " + os.length());
+      }
+      if ( sizeOfCopy < MIN_MULTIPART_SIZE && i < composeSources.size() - 1 ) {
+        throw new InvalidArgumentException(" Source Object " + i +  " is too small"
+        +  " size :  " + sizeOfCopy + " and it is not the last part");
+      }
+
+      totalSize = totalSize + sizeOfCopy;
+      if ( totalSize > MAX_OBJECT_SIZE ) {
+        throw new InvalidArgumentException(" Cannot compose an object of size"
+        + " more than " + MAX_OBJECT_SIZE );
+      }
+      totalParts += partsRequired(sizeOfCopy);
+      sourceSize[i] = sizeOfCopy;
+      if ( totalParts > MAX_MULTIPART_COUNT ) {
+        throw new InvalidArgumentException(" Your proposed compose object requires"
+        + " more than " + MAX_MULTIPART_COUNT + " parts.");
+      }
+
+      // Ensure that the object has not been changed while
+      // we are copying data. Add x-amz-copy-source-if-match
+      // if it is not provided as a part of CopyCondition
+      if (!sourceHeader.containsKey("x-amz-copy-source-if-match")) {
+        sourceHeader.put("x-amz-copy-source-if-match",os.etag());
+      }
+      i++;
+    }
+
+    // Initiate a new multipart upload.
+    // Set user-metadata on the destination object. If no
+    // userMetaData is specified, and there is only one source,
+    // (only) then metadata from source is copied.
+    if ( userMetaData == null  && composeSources.size() == 1  ) {
+      userMetaData = new HashMap<String,String>();
+      userMetaData.putAll(composeSources.get(0).headers());
+    }
+
+    // Now, handle multipart-copy cases.
+    Map<String, String> headerMap = new HashMap<>();
+    if (sseTarget != null) {
+      if ( userMetaData != null ) {
+        sseTarget.marshal(userMetaData);
+      } else {
+        userMetaData = new HashMap<>();
+        sseTarget.marshal(userMetaData);
+      }
+    }
+    if (userMetaData != null ) {
+      headerMap.putAll(userMetaData);
+    }
+
+    String uploadId = initMultipartUpload(bucketName, objectName, headerMap);
+
+    int partNumber = 1;
+    Part[] totalPart = new Part[totalParts];
+    int counter = 0;
+
+    for (ComposeSource sourceObject : composeSources) {
+      headerMap.putAll(sourceObject.headers());
+      List<String> rangeList = getByteRange(sourceObject, sourceSize[counter]);
+      for (int k = 0; k < rangeList.size(); k++) {
+        headerMap.put("x-amz-copy-source-range", rangeList.get(k));
+        String eTag = uploadPartCopy(bucketName, objectName, uploadId, partNumber,
+            headerMap);
+        totalPart[partNumber - 1] = new Part(partNumber, eTag);
+        partNumber++;
+      }
+      counter++;
+    }
+    completeMultipart(bucketName, objectName, uploadId, totalPart);
+  }
+
+  private int partsRequired(Long length ) {
+    /* maxPartSize will always be an int as 5 *1024 *1024* 1024 *1024 /10000-1
+    * = 549810794.96749675
+    which is within int range
+    */
+    long maxPartSize = MAX_OBJECT_SIZE / ( MAX_MULTIPART_COUNT - 1 );
+    long partCount = length / maxPartSize;
+    if ( length % maxPartSize > 0L ) {
+      partCount ++;
+    }
+    return (int)partCount;
+  }
+
+  private List<String> getByteRange( ComposeSource source, long size ) {
+    List<String> list = new LinkedList<String>();
+    int requiredParts = partsRequired(size);
+    long offset = source.offset() == null ? 0L : source.offset();
+    long quot = size / requiredParts;
+    long rem = size % requiredParts;
+    long nextStart = offset;
+    //long curPartSize = quot;
+    for ( int j = 0; j < requiredParts; j++) {
+      long curPartSize = quot;
+      if ( j < rem ) {
+        curPartSize++;
+      }
+      long cStart = nextStart;
+      long  cEnd = cStart + curPartSize - 1;
+      nextStart = cEnd + 1;
+      list.add("bytes=" + cStart + "-" + cEnd);
+    }
+    return list;
+  }
+
+
   /**
    * Returns a presigned URL string with given HTTP method, expiry time and custom request params for a 
    * specific object in the bucket.
@@ -4259,6 +4473,8 @@ public class MinioClient {
   }
 
 
+
+
   /**
    * Executes put object. If size of object data is <= 5MiB, single put object is used
    * else multipart put object is used.
@@ -5021,6 +5237,37 @@ public class MinioClient {
     result.parseXml(response.body().charStream());
     response.body().close();
     return result.uploadId();
+  }
+
+  /**
+   * Executes uploadPartCopy multipart upload of given bucket name, 
+   * object name, upload ID and parts.
+   */
+  private String uploadPartCopy(String bucketName, String objectName,
+      String uploadId, int partNumber, Map<String, String> headerMap)
+    throws InvalidBucketNameException, NoSuchAlgorithmException, 
+      InsufficientDataException, IOException, InvalidKeyException, 
+      NoResponseException, XmlPullParserException, ErrorResponseException,
+      InternalException, InvalidResponseException {
+
+    HttpResponse response = null;
+    Map<String,String> queryParamMap = null;
+    if (partNumber > 0 && uploadId != null && !"".equals(uploadId)) {
+      queryParamMap = new LinkedHashMap<>();
+      queryParamMap.put("partNumber", Integer.toString(partNumber));
+      queryParamMap.put("uploadId", uploadId);
+
+    }
+    response = executePut(bucketName, objectName, headerMap, queryParamMap,
+       getRegion(bucketName) ,"", 0);
+
+    CopyPartResult result = new CopyPartResult();
+    try (ResponseBody body = response.body()) {
+      result.parseXml(body.charStream());
+    }
+    response.body().close();
+    return result.etag();
+
   }
 
 
--- a/api/src/main/java/io/minio/ComposeSource.java
+++ b/api/src/main/java/io/minio/ComposeSource.java
@@ -1,0 +1,149 @@
+/*
+ * Minio Java SDK for Amazon S3 Compatible Cloud Storage, (C) 2019 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.minio;
+
+import java.util.HashMap;
+import java.util.Map;
+import io.minio.errors.InvalidArgumentException;
+import io.minio.CopyConditions;
+
+public class ComposeSource {
+  private String bucketName;
+  private String objectName;
+  private Long offset;
+  private Long length;
+  private ServerSideEncryption sseSource;
+  private Map<String, String> headerMap ;
+
+  /**
+   * Create new ComposeSource for given bucket and object.
+   */
+  public ComposeSource(String bucketName, String objectName ) 
+    throws InvalidArgumentException {
+    this(bucketName, objectName, null, null, null, null, null);
+  }
+  /**
+   * Create new ComposeSource for given bucket, object, offset and length.
+   */
+  public ComposeSource(String bucketName, String objectName, Long offset, 
+      Long length)
+    throws InvalidArgumentException {
+    this(bucketName, objectName, offset, length, null, null, null);
+  }
+  /**
+   * Create new ComposeSource for given bucket, object, offset, length 
+   * and headerMap.
+   */
+  public ComposeSource(String bucketName, String objectName, Long offset,
+      Long length,Map<String, String> headerMap) 
+    throws InvalidArgumentException {
+    this(bucketName, objectName, offset, length, headerMap, null, null);
+  }
+
+  /**
+   * Create new ComposeSource for given bucket, object, offset, length, 
+   * headerMap and CopyConditions.
+   */
+  public ComposeSource(String bucketName, String objectName, Long offset, 
+      Long length,Map<String, String> headerMap,
+      CopyConditions copyConditions)
+    throws InvalidArgumentException {
+    this(bucketName, objectName, offset, length, headerMap, 
+      copyConditions, null);
+  }
+
+  /**
+   * Creates new ComposeSource for given bucket, object, offset, length,
+   * headerMap, CopyConditions and server side encryption.
+   *
+   * @throws InvalidArgumentException
+   *          When offset or lenghth is negative
+   */
+  public ComposeSource(String bucketName, String objectName, Long offset,
+       Long length, Map<String, String> headerMap, 
+       CopyConditions copyConditions, ServerSideEncryption sseSource)
+    throws InvalidArgumentException {
+    if (bucketName == null) {
+      throw new InvalidArgumentException("Source bucket name cannot be empty");
+    }
+    if (objectName == null) {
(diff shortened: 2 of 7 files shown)


api/src/main/java/io/minio/DestinationInfo.java
```diff
@@ -0,0 +1,73 @@
+package io.minio;
+
+
+import java.util.Map;
+
+
+/** DestinationInfo creates the DestinationObject
+ * used for Compose Object.
+*/
+public class DestinationInfo {
```
balamurugana: I am not favour of this class in java supporting polymorphism.  `composeObject()` may have below versions
```java
composeObject(bucketName String, objectName String, ComposeSource []sources)
composeObject(bucketName String, objectName String, ServerSideEncryption sse, Map<String,String> userMetaData, ComposeSource []sources)
```

api/src/main/java/io/minio/SourceInfo.java
```diff
@@ -0,0 +1,99 @@
...
+
+import io.minio.errors.InvalidArgumentException;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/** SourceInfo creates the SourceObject
+ * used for Compose Object.
+ */
+public class SourceInfo {
```
balamurugana: My choice is `ComposeSource`

api/src/main/java/io/minio/SourceInfo.java
```diff
@@ -0,0 +1,99 @@
...
+  private String bucketName;
+  private String objectName;
+  private int start;
+  private int end;
```
balamurugana: use `offset`/`length` instead of `start`/`end`

api/src/main/java/io/minio/SourceInfo.java
```diff
@@ -0,0 +1,99 @@
...
+  public SourceInfo(String bucketName, String objectName, ServerSideEncryption sseSource) {
+    this.bucketName = bucketName;
+    this.objectName = objectName;
+    this.sseSource = sseSource;
+    this.start = -1;
+    headerMap.put("x-amz-copy-source", S3Escaper.encodePath(bucketName + "/"
+            + objectName));
+  }
+
+  public String getBucketName() {
```
balamurugana: You could have 
1. `source()` which returns `/source_bucket/sourceObject`
2. `headers()` which returns all headers including SSE and below;
```
x-amz-copy-source-range:bytes=first-last
x-amz-copy-source-if-match: etag
x-amz-copy-source-if-none-match: etag
x-amz-copy-source-if-unmodified-since: time_stamp
x-amz-copy-source-if-modified-since: time_stamp
```

All other getters are not useful.

api/src/main/java/io/minio/SourceInfo.java
```diff
@@ -0,0 +1,99 @@
...
+    return bucketName;
+  }
+
+  public void setBucketName(String bucketName) {
```
balamurugana: You could have below constructors
```
CopySource(String bucketName, String objectName)
CopySource(String bucketName, String objectName, int offset, int length)
CopySource(String bucketName, String objectName, Integer offset, Integer length, Map<String, String> headerMap)
CopySource(String bucketName, String objectName, Integer offset, Integer length, ServerSideEncryption sse, Map<String, String> headerMap)
```
and no setters are required.  here `headerMap` carries all headers below where user sets
```
x-amz-copy-source-range:bytes=first-last
x-amz-copy-source-if-match: etag
x-amz-copy-source-if-none-match: etag
x-amz-copy-source-if-unmodified-since: time_stamp
x-amz-copy-source-if-modified-since: time_stamp
```
This way there is no need to have condition methods.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2198,6 +2198,149 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
...
+
+    if (listSourceObjects.size() < 1 || listSourceObjects.size() > MAX_MULTIPART_COUNT) {
+      throw new InvalidArgumentException("There must be as least one and up to 10000 source objects.");
+    }
+    int i = 0;
+    long totalSize = 0;
+    int totalParts = 0;
+    Object data = new Object();
+    Iterator<SourceInfo> sourceObjects = listSourceObjects.iterator();
+    while (sourceObjects.hasNext()) {
```
balamurugana: Use `for` instead of `while`

api/src/main/java/io/minio/CopySource.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.Map;
+
+/** SourceInfo creates the SourceObject
+ * used for Compose Object.
+ */
+public class CopySource {
+  private String bucketName;
+  private String objectName;
+  private int offset;
+  private int length;
```
balamurugana: You would need to use `Integer` than `int` so that they can be set to `null`.

* `offset = null && length == null`, return `x-amz-copy-source-range` header is not set.
* `offset != null && length != null`, return header `x-amz-copy-source-range=byte=<offset>-<offset+length>`
* `offset != null && length == null`, return header `x-amz-copy-source-range=byte=<offset>-`
* `offset == null && length != null`, return header `x-amz-copy-source-range=byte=0-<length>`

Praveenrajmani:
@sinhaashish Is this PR ready for review ?

deekoder:
Should this be closed as a stale PR?

sinhaashish:
> Should this be closed as a stale PR?

I am fixing the code. Will ask for review once done.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -98,16 +79,7 @@
...
-import java.util.HashSet;
-import java.util.Set;
-import java.util.Iterator;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.NoSuchElementException;
-import java.util.Scanner;
+import java.util.*;
```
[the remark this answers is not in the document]
sinhaashish: done

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -38,26 +38,7 @@
...
-import io.minio.messages.ListAllMyBucketsResult;
-import io.minio.messages.ListBucketResult;
-import io.minio.messages.ListBucketResultV1;
-import io.minio.messages.ListMultipartUploadsResult;
-import io.minio.messages.ListPartsResult;
-import io.minio.messages.Part;
-import io.minio.messages.Prefix;
-import io.minio.messages.Upload;
-import io.minio.messages.NotificationConfiguration;
+import io.minio.messages.*;
```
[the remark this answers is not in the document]
sinhaashish: done

api/src/main/java/io/minio/CopySource.java
sinhaashish: done.

sinhaashish:
@balamurugana @Praveenrajmani  PTAL

api/src/main/java/io/minio/CopySource.java
```diff
@@ -0,0 +1,136 @@
+package io.minio;
+
+import java.util.HashMap;
+import java.util.Map;
+import io.minio.errors.InvalidArgumentException;
+
+/** SourceInfo creates the SourceObject
+ * used for Compose Object.
```
balamurugana: Is this comment correct?

api/src/main/java/io/minio/CopySource.java
```diff
@@ -0,0 +1,136 @@
...
+public class CopySource {
+  private String bucketName;
+  private String objectName;
+  private Integer offset;
+  private Integer length;
+  private ServerSideEncryption sseSource;
+  private Map<String, String> headerMap = new HashMap<>();
+
+  /**
+   * Constucts the CopySource with bucket name and object name.
```
balamurugana: Use same language as https://github.com/minio/minio-java/blob/master/api/src/main/java/io/minio/ChunkedInputStream.java#L73

api/src/main/java/io/minio/CopySource.java
```diff
@@ -0,0 +1,136 @@
...
+    return length - offset;
+  }
+
+  /**
+   * Returns the headerMap after setting copy-source, copy-source-range .
+   */
+  public Map<String, String> headers() throws InvalidArgumentException {
+/*    if (offset != -1) {
+      headerMap.put("x-amz-copy-source-range", "bytes=" + offset + "-" + length);
+    }*/
```
balamurugana: remove stale comments

api/src/main/java/io/minio/CopySource.java
```diff
@@ -0,0 +1,136 @@
...
+
+
+    if (offset == null) {
+      if ( length != null) {
+        headerMap.put("x-amz-copy-source-range", "bytes="  + "0-" + length);
+      } else {
+        throw new InvalidArgumentException("x-amz-copy-source-range header is not set.");
```
balamurugana: Is it required to set `x-amz-copy-source-range` for put object copy?

api/src/main/java/io/minio/CopySource.java
```diff
@@ -0,0 +1,136 @@
...
+      }
+    } else {
+      if ( length != null) {
+        if ( offset < 0 || length < 0) {
```
balamurugana: this check needs to be outside of above `if`

api/src/main/java/io/minio/CopySource.java
```diff
@@ -0,0 +1,136 @@
...
+          throw new InvalidArgumentException("Offset/Length can not be negative");
+        }
+        headerMap.put("x-amz-copy-source-range", "bytes=" + offset + "-" + length);
+      } else {
+        headerMap.put("x-amz-copy-source-range", "bytes=" + offset + "-" );
+      }
+    }
+
+
+    headerMap.put("x-amz-copy-source",S3Escaper.encodePath("/" + this.bucketName + "/"
```
balamurugana: is it required to encode full path or bucket name and object name?

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2294,6 +2297,182 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
     }
   }
 
+
+  /**
+   * ComposeObject - It creates an object using server-side copying of
```
balamurugana: Is this javadoc style?

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2294,6 +2297,182 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
...
+   * @throws InvalidKeyException
+   *           upon an invalid access key or secret key
+   * @throws NoResponseException         upon no response from server
+   * @throws XmlPullParserException      upon parsing response xml
+   * @throws ErrorResponseException      upon unsuccessful execution
+   * @throws InternalException           upon internal library error
+   * @throws InvalidArgumentException    upon invalid value is passed to a method.
+   */
+
+  public void  composeObject(String destinationBucketName, String destintionObjectName,
```
balamurugana: call `bucketName` and `objectName` like other methods e.g. `putObject()`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2294,6 +2297,182 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
...
+                             List<CopySource> listSourceObjects)
```
balamurugana: `copySources` is much better than `listSourceObjects`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2294,6 +2297,182 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
...
+   */
+
+
+  public void  composeObject(String destinationBucketName, String destintionObjectName, ServerSideEncryption sse,
+                             Map<String,String> userMetaData, List<CopySource> listSourceObjects)
+          throws InvalidBucketNameException, NoSuchAlgorithmException, InsufficientDataException, IOException,
+          InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
+          InternalException, InvalidArgumentException {
+
+    if (listSourceObjects.size() < 1 || listSourceObjects.size() > MAX_MULTIPART_COUNT ) {
```
balamurugana: use `listSourceObjects.size() == 0`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2294,6 +2297,182 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
...
+      throw new InvalidArgumentException("There must be as least one and up to 10000 source objects.");
```
balamurugana: you could use `MAX_MULTIPART_COUNT` than `10000`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2294,6 +2297,182 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
...
+    }
+    int i = 0;
+    long totalSize = 0;
+    for (CopySource sourceObject : listSourceObjects) {
+      i++;
+      ObjectStat os = statObject(sourceObject.bucketName(), sourceObject.objectName());
+      // Error out if client side encryption is used in this source object when
+      // more than one source objects are given.
+      if ((listSourceObjects.size() > 1 )  && sourceObject.headers().get("x-amz-meta-x-amz-key") !=  null) {
```
balamurugana: use `Map.containsKey()` than `Map.get()`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2294,6 +2297,182 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
...
+                + sourceObject.bucketName() + sourceObject.objectName());
+      }
+
+      // Check if a segment is specified is within object bounds
+      if ( sourceObject.length() > os.length()) {
+        throw new InvalidArgumentException("Source specified length " + sourceObject.length() + " is more than  "
+                +  "size of Object " + os.length());
+      }
+
+      if ( os.length() < MIN_MULTIPART_SIZE && i < listSourceObjects.size() - 1 ) {
```
balamurugana: `i` is incremented before this, hence this check doesn't work.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2294,6 +2297,182 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
...
+    if (sse != null) {
+      sse.marshal(headerMap);
+    }
+    String uploadId = initMultipartUpload(destinationBucketName,destintionObjectName , headerMap);
+    int partNumber = 1;
+    Part[] totalPart = new Part[listSourceObjects.size()];
+
+    for (CopySource source : listSourceObjects) {
+      Map<String, String> sourceMap  = source.headers();
+      ObjectStat os = statObject(source.bucketName(), source.objectName());
```
balamurugana: Why is it done once again?  You could merge or do checks/constructions in only one loop.

api/src/main/java/io/minio/CopySource.java
sinhaashish: Modified.

api/src/main/java/io/minio/CopySource.java
sinhaashish: done.

api/src/main/java/io/minio/CopySource.java
sinhaashish: removed

api/src/main/java/io/minio/CopySource.java
sinhaashish: yes.

api/src/main/java/io/minio/CopySource.java
sinhaashish: done.

api/src/main/java/io/minio/CopySource.java
sinhaashish: edited

api/src/main/java/io/minio/MinioClient.java
sinhaashish: edited.

api/src/main/java/io/minio/MinioClient.java
sinhaashish: done.

api/src/main/java/io/minio/MinioClient.java
sinhaashish: done.

api/src/main/java/io/minio/MinioClient.java
sinhaashish: done.

api/src/main/java/io/minio/MinioClient.java
sinhaashish: done.

api/src/main/java/io/minio/MinioClient.java
sinhaashish: done.

api/src/main/java/io/minio/MinioClient.java
sinhaashish: done.

api/src/main/java/io/minio/MinioClient.java
sinhaashish: All the validations are done in the first for loop. Lets say we have to compose object from 800 objects. the validations fails for 799th object so all the parts copied are useless. So we first validate and then compose object. The same in done in `minio-go` sdk

donatello: Yeah, unfortunately in the compose API, even if the request appears to be valid, the operation may not be feasible - some source objects may be larger than 5GB and may need to be broken up into multiple part uploads - this may in turn require more than 10K parts. For this reason a separate loop performs the validation and a second loop performs the upload. To ensure that the source objects do not change between the first and second loop, the second loop uses the Src-If-Match condition in the go-implementation, this should be done here too. This feature and all these implementation details were discussed with @abperiasamy when it was first implemented in Go (some customer had requested this feature at that time).

sinhaashish: @balamurugana  PTAL

api/src/main/java/io/minio/CopySource.java
```diff
@@ -0,0 +1,139 @@
...
+    if ( sseSource != null) {
+      sseSource.marshal(headerMap);
+    }
+    return  headerMap;
+  }
+
+  /**
+   * Sets the range for copy .
+   */
+  public Map<String,String> setCopySourceRange() {
```
balamurugana: Few correction in this class.

1. offset and length needs to be `Long`sseSource.marshal(headerMap);
2. all constructors need to call `public CopySource(String bucketName, String objectName, Integer offset, Integer length, ServerSideEncryption sseSource, Map<String, String> headerMap)` to avoid repeated validations.
3. rename `getHeaders()` to `headers()`
4. `sseSource.marshal(headerMap);` needs to be set in constructor `public CopySource(String bucketName, String objectName, Integer offset, Integer length, ServerSideEncryption sseSource, Map<String, String> headerMap)` to make `headers()` to be simple and straight forward.
5. `x-amz-copy-range-source-range` needs to be set in constructor `public CopySource(String bucketName, String objectName, Integer offset, Integer length, ServerSideEncryption sseSource, Map<String, String> headerMap)` and `setCopySourceRange()` is unnecessary.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2294,6 +2297,200 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
...
+   * @throws NoResponseException         upon no response from server
+   * @throws XmlPullParserException      upon parsing response xml
+   * @throws ErrorResponseException      upon unsuccessful execution
+   * @throws InternalException           upon internal library error
+   * @throws InvalidArgumentException    upon invalid value is passed to a method.
+   */
+
+
+  public void  composeObject(String bucketName, String objectName, ServerSideEncryption sse,
+                             Map<String,String> userMetaData, List<CopySource> copySources)
```
balamurugana: 1. move optional arguments to end i.e. `public void  composeObject(String bucketName, String objectName, List<CopySource> copySources, Map<String,String> userMetaData, ServerSideEncryption sse)`
2. you could remove another version `  public void  composeObject(String bucketName, String objectName, List<CopySource> copySources)`.  This is not necessary to overload.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2294,6 +2297,200 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
...
+          throws InvalidBucketNameException, NoSuchAlgorithmException, InsufficientDataException, IOException,
+          InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
+          InternalException, InvalidArgumentException {
+
+    if (copySources.size() == 0 || copySources.size() > MAX_MULTIPART_COUNT ) {
```
balamurugana: use `copySources.isEmpty()`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2294,6 +2297,200 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
...
+      throw new InvalidArgumentException("There must be as least one and up to MAX_MULTIPART_COUNT source objects.");
+    }
+    int i = 1;
+    int totalParts = 0;
+    long totalSize = 0;
+    for (CopySource sourceObject : copySources) {
```
balamurugana: you could call it as just `source` than `sourceObject`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2294,6 +2297,200 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
...
+
+      ObjectStat os = statObject(sourceObject.bucketName(), sourceObject.objectName());
+      // Error out if client side encryption is used in this source object when
+      // more than one source objects are given.
+      if ((copySources.size() > 1 )  && sourceObject.getHeaders().containsKey("x-amz-meta-x-amz-key")) {
```
balamurugana: `sourceObject.geHeaders().containsKey("x-amz-meta-x-amz-key")` check should be done in `CopySource` constructor.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2294,6 +2297,200 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
...
+
+      totalSize = totalSize + os.length();
+      if ( totalSize > MAX_OBJECT_SIZE ) {
+        throw new InvalidArgumentException(" Cannot compose an object of size more than " + MAX_OBJECT_SIZE );
+      }
+      if ( sourceObject.length() != null ) {
+        totalParts += calculateMultipartSize(sourceObject.length())[1];
+      } else {
+        totalParts += calculateMultipartSize((int)os.length())[1];
+      }
```
balamurugana: you use `os.length()` and `sourceObject.length()` based on conditions.  This makes logical error here and above.  Its better to define `length` variable and use `length` variable appropriately.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2294,6 +2297,200 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
...
+
+      if ( totalParts > MAX_MULTIPART_COUNT ) {
+        throw new InvalidArgumentException(" Your proposed compose object requires more than " + MAX_MULTIPART_COUNT
+         + " parts.");
+      }
+
+      //  Ensure that the object has not been changed while
+      //  we are copying data.
+      if (!sourceObject.getHeaders().containsKey("x-amz-copy-source-if-match")) {
+        sourceObject.getHeaders().put("x-amz-copy-source-if-match",os.etag());
```
balamurugana: 1. it is bad idea to mutate headers in `sourceObject`.  Get headers `getHeaders()` once and mutate it wherever needed.
2. This check and addition needs to be done after `statObject()`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2294,6 +2297,200 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
...
+      }
+      long end = os.length() - 1;
+      sourceObject.getHeaders().put("x-amz-copy-source-range","bytes="  + "0-" + end);
```
balamurugana: Why is it needed to copy entire object?

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2294,6 +2297,200 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
...
+      i++;
+    }
+
+    // Single source object case (i.e. when only one source is
+    // involved, it is being copied wholly and at most 5GiB in
+    // size, emptyfiles are also supported).
+    if ( copySources.size() == 1  || totalSize == 0 ) {
```
balamurugana: doing `totalSize` check is unnecessary for `copyObject()`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2294,6 +2297,200 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
...
+      CopySource  sourceObject = copySources.get(0);
+      copyObject(sourceObject.bucketName(), sourceObject.objectName(),
+              bucketName, objectName);
```
balamurugana: Why aren't you using `headers()` from `sourceObject`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2294,6 +2297,200 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
...
+    }
+
+    //  Initiate a new multipart upload.
+    // Set user-metadata on the destination object. If no
+    // userMetaData is specified, and there is only one source,
+    // (only) then metadata from source is copied.
+    if ( userMetaData == null  && copySources.size() == 1  ) {
+      userMetaData = new HashMap<String,String>();
+      userMetaData.putAll(copySources.get(0).getHeaders());
+    }
```
balamurugana: There are two cases and two sub-cases each to handle correctly.
1. When single copy source is given.
  a. When size is <= 5 GiB, you could use `PUT Object - Copy` to complete it. (OR)
  b. When size is > 5 GiB, you would need to use multipart to do copy object.
2. When more than one copy sources are given
  a. When each size of all of copy sources are <= 5 GiB, you would need to use multipart to do copy object. (OR)
  b. When one (or more) of the size of copy sources is > 5 GiB, you would need to use multipart to split > 5 GiB copy source into multiple parts of 5 GiB and upload them.

I suspect you are not handling all the cases properly.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -4726,6 +4925,36 @@ private String initMultipartUpload(String bucketName, String objectName, Map<Str
   }
 
 
+  /**
+   * Executes uploadPartCopy multipart upload of given bucket name, object name, upload ID and parts.
+   */
+  private String uploadPartCopy(String bucketName, String objectName,String uploadId, int partNumber,
+                                Map<String, String> headerMap)
```
balamurugana: I see [PUT Object](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUT.html) and [PUT Object - Copy](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectCOPY.html) differs only with headers. This means existing `putObject()` can be used than duplicating.
2. Similarly [Upload Part](https://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPart.html) and [Upload Part - Copy](https://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPartCopy.html) differs only with headers. This means existing `putObject()` can be used than duplicating.

sinhaashish: `Upload Part` and `Upload Part - Copy` differs in response. In `Upload Part ` etag is a part of header while in `Upload Part - Copy` the etag is a part of body.  `Initaiate multipart` and `Complete multipart` were same so used it as is. Removed 'Put Object' .

sinhaashish:
@balamurugana  @Praveenrajmani  PTAL

api/src/main/java/io/minio/CopySource.java
```diff
@@ -0,0 +1,135 @@
...
+public class CopySource {
+  private String bucketName;
+  private String objectName;
+  private Long offset;
+  private Long length;
+  private ServerSideEncryption sseSource;
+  private Map<String, String> headerMap = new HashMap<>();
+  private CopyConditions copyConditions;
+
+  public ServerSideEncryption getSseSource() {
```
balamurugana: Have this function name as `sseSource()`

api/src/main/java/io/minio/CopySource.java
balamurugana: This comment is not useful or not clear.

api/src/main/java/io/minio/CopySource.java
```diff
@@ -0,0 +1,135 @@
...
+   * Returns headerMap.
+   */
+
+  public Map<String, String> headers() {
+    if ( headerMap ==   null ) {
+      headerMap = new HashMap<String,String>();
+    }
+    // If no conditions available, skip addition else add the conditions to the header
+    if (copyConditions != null) {
+      headerMap.putAll(copyConditions.getConditions());
```
balamurugana: You are adding to `headerMap` every time this method is called. Similarly initialization happens too. This method should just return `headerMap` and updating `headerMap` should be done in `public CopySource(String bucketName, String objectName, Long offset, Long length, Map<String, String> headerMap,
      CopyConditions copyConditions, ServerSideEncryption sseSource)`

api/src/main/java/io/minio/CopySource.java
sinhaashish: 1. Done
2. Done
3. done
4. Done
5.private method `getByteRange` introduced to calculate the value for `x-amz-copy-range-source-range` so it is not in constructor.

api/src/main/java/io/minio/MinioClient.java
sinhaashish: Done

api/src/main/java/io/minio/CopySource.java
sinhaashish: done.

api/src/main/java/io/minio/CopySource.java
sinhaashish: done.

api/src/main/java/io/minio/MinioClient.java
sinhaashish: Removed

api/src/main/java/io/minio/MinioClient.java
sinhaashish: Removed

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2294,6 +2297,200 @@ public void copyObject(String bucketName, String objectName, String destBucketNa
...
+
+    if (copySources.size() == 0 || copySources.size() > MAX_MULTIPART_COUNT ) {
+      throw new InvalidArgumentException("There must be as least one and up to MAX_MULTIPART_COUNT source objects.");
+    }
+    int i = 1;
+    int totalParts = 0;
+    long totalSize = 0;
+    for (CopySource sourceObject : copySources) {
+
+      ObjectStat os = statObject(sourceObject.bucketName(), sourceObject.objectName());
```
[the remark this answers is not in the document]
sinhaashish: Done

api/src/main/java/io/minio/MinioClient.java
sinhaashish: Done

api/src/main/java/io/minio/MinioClient.java
sinhaashish: Done

kannappanr:
blocked until 776, 777 and 778 are merged

PR #750, opened 2019-02-14
Docs updated with Set/Get BucketNotification

sinhaashish:
Set/Get Bucket Notification added in docs. The lik to java docs is also updated for some APIs.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -24,7 +24,8 @@ MinioClient s3Client = new MinioClient("https://s3.amazonaws.com", "YOUR-ACCESSK
 | [`listObjects`](#listObjects)  | [`removeObject`](#removeObject) |   |  [`deleteBucketLifeCycle`](#deleteBucketLifeCycle) |
 | [`listIncompleteUploads`](#listIncompleteUploads)  | [`removeIncompleteUpload`](#removeIncompleteUpload) |   |   |
 | [`listenBucketNotification`](#listenBucketNotification) |  |   |   |
-
+| [`setBucketNotification`](#setBucketNotification) |  |   |   |
+| [`getBucketNotification`](#getBucketNotification) |  |   |   |
 
 ## 1. Constructors
 
@@ -651,9 +652,9 @@ try {
 ### setBucketLifeCycle(String bucketName, String lifeCycle)
 `public void setBucketLifeCycle(String bucketName, String lifeCycle)`
 
-Set a life cydle on bucket.
+Set a life cycle on bucket.
 
-[View Javadoc](http://minio.github.io/minio-java/io/minio/MinioClient.html#setBucketLifeCycle-java.lang.String-java.lang.String-io.minio.BucketLifeCycle-)
+[View Javadoc](http://minio.github.io/minio-java/io/minio/MinioClient.html#setBucketLifeCycle-java.lang.String-java.lang.String-)
 
 __Parameters__
 
@@ -744,7 +745,7 @@ try {
 
 Delete the lifecycle of the bucket.
 
-[View Javadoc](http://minio.github.io/minio-java/io/minio/MinioClient.html#getBucketLifeCycle-java.lang.String-)
+[View Javadoc](http://minio.github.io/minio-java/io/minio/MinioClient.html#deleteBucketLifeCycle-java.lang.String-)
 
 __Parameters__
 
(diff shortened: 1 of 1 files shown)


docs/API.md
```diff
@@ -796,11 +797,20 @@ __Parameters__
 | ``prefix`` | _String_ | Only listen for objects with the given prefix. |
 | ``suffix`` | _String_ | Only listen for objects with the given suffix. |
 | ``events`` | _String[]_ | Only listen for the specified events, such as s3:ObjectCreated:*, s3:ObjectAccessed:*, s3:ObjectRemoved:*, ..  |
+| ``listener`` | _ BucketEventListener_ | Interface with updateEvent method |
```
ebozduman: You have an extra space in between `_`  and `BucketEventListener`
in `_ BucketEventListener_`

docs/API.md
```diff
@@ -796,11 +797,20 @@ __Parameters__
...
 
 | Return Type	  | Exceptions	  |
 |:--- |:--- |
 |  None  | Listed Exceptions: |
-|        |  ``RuntimeException`` : runtime exception. |
+|        |  ``InvalidBucketNameException`` : upon invalid bucket name. |
+|        | ``NoSuchAlgorithmException`` : upon requested algorithm was not found during signature calculation.  |
+|        | ``InsufficientDataException`` : Thrown to indicate that reading given InputStream gets EOFException before reading given length. |
```
ebozduman: I guess you need a space in between `EOF` and `Exception`

docs/API.md
```diff
@@ -824,6 +834,115 @@ __Example__
...
+| ``bucketName``  | _String_  | Name of the bucket.  |
+| ``notificationConfiguration`` | _NotificationConfiguration_ | Notification configuration to be set. |
+
+| Return Type	  | Exceptions	  |
+|:--- |:--- |
+|  None  | Listed Exceptions: |
+|        | ``InvalidBucketNameException`` : upon invalid bucket name. |
+|        | ``InvalidObjectPrefixException`` : upon invalid object prefix. |
+|        | ``NoSuchAlgorithmException`` : upon requested algorithm was not found during signature calculation.  |
+|        | ``InsufficientDataException`` : Thrown to indicate that reading given InputStream gets EOFException before reading given length. |
```
ebozduman: The same as the previous comment.
```
This is documentation, it needs to be meaningful text imho.
So, what is `given InputStream` from the perspective of the end-user? 
Why do we say `given`? This is not an input parameter end-user provides.
We should not reveal any internal parameters when we explain the meaning of return values.

You also need a space in between `EOF` and `Exception`. Even it would be better to call it something like `"End of file exception"`.```

File: docs/API.md [hunk 4]
```diff
@@ -824,6 +834,115 @@ __Example__
...
+|:--- |:--- |:--- |
+| ``bucketName``  | _String_  | Name of the bucket.  |
+
+| Return Type	  | Exceptions	  |
+|:--- |:--- |
+|  ``NotificationConfiguration``:  NotificationConfiguration    | Listed Exceptions: |
+|        | ``InvalidBucketNameException`` : upon invalid bucket name. |
+|        | ``InvalidObjectPrefixException`` : upon invalid object prefix. |
+|        | ``NoSuchAlgorithmException`` : upon requested algorithm was not found during signature calculation.  |
+|        | ``InsufficientDataException`` : Thrown to indicate that reading given InputStream gets EOFException before reading given length. |
```
ebozduman: The same comment here.

This is documentation, it needs to be meaningful text, imho.
So, what is `given InputStream` from the perspective of the end-user? 
Why do we say `given`? This is not an input parameter end-user provides.
We should not reveal any internal parameters when we explain the meaning of return values.

You also need a space in between `EOF` and `Exception`. Even it would be better to call it something like `"End of file exception"`.

sinhaashish: modified

docs/API.md
sinhaashish: modified

docs/API.md
sinhaashish: modified

docs/API.md
sinhaashish: Corerected

docs/API.md
```diff
@@ -653,7 +654,7 @@ try {
 
 Set a life cydle on bucket.
```
Praveenrajmani: Typo here

sinhaashish: corrected

ebozduman:
@sinhaashish,
Please look into the Travis failure...

sinhaashish:
> @sinhaashish,
> Please look into the Travis failure...

I ran the `./gradlew runFunctionalTest` locally and its passing. The error shows ```Test: listIncompleteUploads(String bucketName)
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.``` and is no where related to the docs changes in this PR.

PR #758, opened 2019-03-07
Fix hangs with TLS and HTTP 2.0

harshavardhana:


harshavardhana:
This PR should be merged https://github.com/minio/minio/pull/7334 before this PR builds properly.

api/src/main/java/io/minio/HttpRequestBody.java
```diff
@@ -71,7 +71,7 @@ public long contentLength() {
   public void writeTo(BufferedSink sink) throws IOException {
     if (data instanceof InputStream) {
       InputStream stream = (InputStream) data;
-      sink.write(Okio.source(stream), len);
+      sink.writeAll(Okio.source(stream));
```
balamurugana: Could you share why `sink.Write()` is a problem?

api/src/main/java/io/minio/HttpResponse.java
```diff
@@ -52,6 +52,13 @@ public ResponseBody body() {
     return this.response.body();
   }
 
+  /**
+   * closes the response.
+   */
+  public void close() {
```
balamurugana: Keep this clean up as seprate PR.

functional/ContentInputStream.java
```diff
@@ -171,7 +172,6 @@ public int read() throws IOException {
     }
 
     if (available <= 0) {
-      // Indicate EOF
```
balamurugana: Is this comment inappropriate?

functional/ContentInputStream.java
```diff
@@ -192,7 +192,7 @@ public long skip(long n) throws IOException {
     }
 
     if (available <= 0) {
-      throw new IOException("no more data");
+      throw new EOFException("no more data available");
```
balamurugana: `EOFException` is not used anywhere.  Why can't we retain `IOException`?

api/src/main/java/io/minio/HttpRequestBody.java
harshavardhana: That is the cause of the hang @balamurugana

functional/ContentInputStream.java
harshavardhana: that needs to be added back, it was a mistake to remove it.

api/src/main/java/io/minio/HttpResponse.java
harshavardhana: this is also needed otherwise we had bugs where the body is not closed properly.

functional/ContentInputStream.java
harshavardhana: EOF is more appropriate I believe we can keep IOException as well.

api/src/main/java/io/minio/HttpResponse.java
balamurugana: Where is the code the body wasn't closed?

functional/ContentInputStream.java
balamurugana: `EOFException` is subclass of `IOException` i.e. someone catching `IOException` is also valid

api/src/main/java/io/minio/HttpResponse.java
harshavardhana: Error was shown by okhttp that we are leaking response and it is recommended that we close response.close() instead of response.body().close()

api/src/main/java/io/minio/HttpRequestBody.java
harshavardhana: For now will keep this in the PR until we get clarity upstream.

kannappanr:
ping @balamurugana @sinhaashish

api/src/main/java/io/minio/HttpRequestBody.java
balamurugana: Using `sink.WriteAll()` is incorrect i.e. all data of `stream` is written, only `len` needs to be written.

We would need to find root cause of `sink.write()` failure.

harshavardhana: This is acceptable for now @balamurugana - root cause for write() is perhaps in the HTTP 2.0 implementation of reading RequestBody... I haven't found the root cause yet.

But this fixes everything for now. I suggest we take this in and move forward.

balamurugana: Reading from input stream beyond given size leads to data loss (for user).  e.g. [putObject()](http://minio.github.io/minio-java/io/minio/MinioClient.html#putObject-java.lang.String-java.lang.String-java.io.InputStream-long-java.util.Map-) can use partial stream upload and rest of data for other purposes.

I am OK for workarounds.  Unfortunately we don't know the problem yet to solve here.

harshavardhana: I don't know why you think it's data loss, content is set and server honors it. If more data is sent server fails. So either way it's a moot argument. What we are trying to test is does write return an exception which is marginal..

sinhaashish:
@harshavardhana  :  Why we don't we need this unit test case `testPutOversizedSmallPut` now ?

harshavardhana:
Mock tests should be removed, also we have real tests for these situationa.. these tests don't add much value.

harshavardhana:
Unblocking @abperiasamy agreed that we can take in this for now.

harshavardhana:
We will open another issue where we will dig deeper into okhttp source to figure out what might be going wrong.

PR #766, opened 2019-04-16
 Protocol set to 1.1

sinhaashish:
`okhttp3.OkHttpClient` uses default http2.0  by default, which has been changed to 1.1 as  web server behaved incorrectly when HTTP/2 is enabled. 
With http2.0 errors like `okhttp3.internal.http2.StreamResetException: stream was reset: PROTOCOL_ERROR`  and `okhttp3.internal.http2.StreamResetException: stream was reset: NO_ERROR` arises

The resolution to this issue as suggested in community is to change the protocol to 1.1 
[Refer](https://github.com/square/okhttp/issues/3955)  and thus this PR restricts the protocol to 1.1

[Fixes](https://github.com/minio/minio/issues/7501)
[Refer](https://square.github.io/okhttp/3.x/okhttp/okhttp3/OkHttpClient.Builder.html)  

```
 public OkHttpClient.Builder protocols(List<Protocol> protocols)
Configure the protocols used by this client to communicate with remote servers. 
By default this client will prefer the most efficient transport available, falling back to more
 ubiquitous protocols. Applications should only call this method to avoid specific 
compatibility problems, such as web servers that behave incorrectly when HTTP/2 is enabled.
```

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -68,6 +68,7 @@ import okhttp3.Request;
 import okhttp3.RequestBody;
 import okhttp3.Response;
 import okhttp3.ResponseBody;
+import okhttp3.Protocol;
 
 import org.joda.time.DateTime;
 import org.xmlpull.v1.XmlPullParser;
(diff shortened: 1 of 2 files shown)


harshavardhana:
The description is not elaborate, can you provide more details of the fix ?

sinhaashish:
> The description is not elaborate, can you provide more details of the fix ?

updated.

harshavardhana:
I am still not sure if this is the right fix @sinhaashish - turning off http2.0 is perhaps not the solution here.

deekoder:
@harshavardhana is there a recommendation for @sinhaashish

harshavardhana:
> @harshavardhana is there a recommendation for @sinhaashish

My recommendation is we talk to upstream Go first and report this issue to them and get their guidance on what might be happening. Since this issue is sporadic as well, we shouldn't default to HTTP 1.1 right away without much information.

sinhaashish:
Go UpSTream [Issue](https://github.com/golang/go/issues/31534)

PR #767, opened 2019-04-16
"gzip" value for "Content-Encoding" header specified by client is overriden when putting object

dnldaniel:
When using MinioClient.putObject with InputStream of a gzipped file body, "Content-Encoding: gzip" header specified by client is always overriden by "Content-Encoding: aws-chunked" (link to the line in code [here](https://github.com/minio/minio-java/blob/6.0.4/api/src/main/java/io/minio/MinioClient.java#L931)). "aws-chunked" value is correct but it can also be linked with client specified encoding (gzip is just one of possible values), e.g. ending up with "Content-Encoding: aws-chunked,gzip" (which is correct according to [AWS specification](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html)).

Here is an example client code:
```java
String filePath = "/home/daniel/random-file.gz";
FileInputStream fileInputStream = new FileInputStream(filePath);
Map<String, String> headers = new HashMap<>(Collections.singletonMap("Content-Encoding", "gzip"));
minioClient.putObject("my-bucket", "random-file", fileInputStream, 0, headers);
```

Because "gzip" header is overriden, thus not propagated, when reading object back (MinioClient.getObject), http layer has no chance to unzip the stream on the fly (because of lack of "Content-Encoding: gzip"). In such case, client has to check the first bytes of returned InputStream manually which is cumbersome and introduces redundant complexity.

My proposal is to always concatenate client specified encodings in such way:
```java
if (headerMap != null) {
  requestBuilder.header("Content-Encoding", Stream.concat(Stream.of("aws-chunked"),
    headerMap.get("Content-Encoding").stream()).distinct().filter(encoding -> !encoding.isEmpty())
    .collect(Collectors.joining(",")));
} else {
  requestBuilder.header("Content-Encoding", "aws-chunked");
}
```

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -100,7 +100,6 @@ import java.security.cert.X509Certificate;
 
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Set;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
@@ -108,8 +107,11 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Scanner;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLSession;
@@ -928,7 +930,15 @@ public class MinioClient {
       // Handle putobject specially to use chunked upload.
       if (method == Method.PUT && objectName != null && body != null && body instanceof InputStream && length > 0) {
         sha256Hash = "STREAMING-AWS4-HMAC-SHA256-PAYLOAD";
-        requestBuilder.header("Content-Encoding", "aws-chunked");
+
+        String contentEncoding = "aws-chunked";
+        if (headerMap != null) {
+          contentEncoding = Stream.concat(Stream.of("aws-chunked"),
+              headerMap.get("Content-Encoding").stream()).distinct().filter(encoding -> !encoding.isEmpty())
+              .collect(Collectors.joining(","));
+        }
+        requestBuilder.header("Content-Encoding", contentEncoding);
+
         requestBuilder.header("x-amz-decoded-content-length", Integer.toString(length));
         chunkedUpload = true;
       } else if (url.isHttps()) {
--- a/api/src/test/java/io/minio/MinioClientTest.java
+++ b/api/src/test/java/io/minio/MinioClientTest.java
@@ -32,9 +32,24 @@ import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.Arrays;
 import java.util.Calendar;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Iterator;
+import java.util.Map;
 import java.util.TimeZone;
 
+import io.minio.errors.ErrorResponseException;
+import io.minio.errors.InsufficientDataException;
+import io.minio.errors.InternalException;
+import io.minio.errors.InvalidArgumentException;
+import io.minio.errors.InvalidBucketNameException;
+import io.minio.errors.InvalidEndpointException;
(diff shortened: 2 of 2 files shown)


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -98,18 +98,11 @@
...
-import java.util.HashSet;
-import java.util.Set;
-import java.util.Iterator;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.NoSuchElementException;
-import java.util.Scanner;
+import java.util.*;
```
balamurugana: Its required to import specifically than all.  It is done by some code analysis tools guidelines.

dnldaniel: I didnt find any linter hint in the CONTRIBUTING. But sure, I can change that.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -928,7 +921,15 @@ private Request createRequest(Method method, String bucketName, String objectNam
...
       if (method == Method.PUT && objectName != null && body != null && body instanceof InputStream && length > 0) {
         sha256Hash = "STREAMING-AWS4-HMAC-SHA256-PAYLOAD";
-        requestBuilder.header("Content-Encoding", "aws-chunked");
+
+        if (headerMap != null) {
+          requestBuilder.header("Content-Encoding", Stream.concat(Stream.of("aws-chunked"),
+              headerMap.get("Content-Encoding").stream()).distinct().filter(encoding -> !encoding.isEmpty())
+              .collect(Collectors.joining(",")));
+        } else {
+          requestBuilder.header("Content-Encoding", "aws-chunked");
```
balamurugana: You could do 
```java
String contentEncoding = "aws-chunked";
if (headerMap != null) {
  contentEncoding = Stream.concat(Stream.of("aws-chunked"),
              headerMap.get("Content-Encoding").stream()).distinct().filter(encoding -> !encoding.isEmpty())
              .collect(Collectors.joining(","));
}
requestBuilder.header("Content-Encoding", contentEncoding)
```

api/src/test/java/io/minio/MinioClientTest.java
```diff
@@ -30,11 +30,10 @@
 import java.security.NoSuchAlgorithmException;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
-import java.util.Arrays;
-import java.util.Calendar;
-import java.util.Iterator;
-import java.util.TimeZone;
+import java.util.*;
```
balamurugana: Its required to import specifically than all. It is done by some code analysis tools guidelines.

api/src/main/java/io/minio/MinioClient.java
dnldaniel: Yeah, looks good. I suppose I will wait for all the comments to incorporate the changes.

balamurugana: Rest of things look good.

api/src/main/java/io/minio/MinioClient.java
dnldaniel: Done (both for MinioClient and MinioClientTest)

api/src/main/java/io/minio/MinioClient.java
dnldaniel: Done

dnldaniel:
@balamurugana it looks like the tests that failed are flaky and the failure reason is not related with the change. Is the travis tests passing required to have the change merged - in such case build retrigger is needed?

balamurugana:
@dnldaniel Yep. Due to some recent changes, travis-ci fails sporadically.  We could restart those jobs one by one till getting success.

dnldaniel:
@balamurugana still failing. Could you try again?

dnldaniel:
> @balamurugana still failing. Could you try again?

@balamurugana @Praveenrajmani The change represents great value for my team and we are awaiting for it to be merged. Is there a way to merge it - e.g. when the flaky tests are supposed to be fixed, or can those be disabled? The state is that review has approval, passes all the tests but unreliable testing pipe blocks merge.

Issue #718, opened 2018-10-25
XMLPullParserException on non XML error ResponseBody


NathanAtKadaster:
(This continues on #354 but I cannot re-open that issue so to not hijack that thread I made opened a new issue. )

So to clarify this a bit more. I stumbled upon this when trying to make a new bucket and the lack of logging is a bit of a pain since only after I found this thread (#354), I was able to find out what happened.

The isssue is that the client does not do any type of content negotiation and just tries to parse anything it gets back as XML.

So after enabling the client object to `#traceOn` `System.out` I found the following request and reponse:

```---------START-HTTP---------
PUT /*REDACTED* HTTP/1.1
Host: localhost:9000
User-Agent: Minio (amd64; amd64) minio-java/dev
x-amz-content-sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
x-amz-date: 20181025T094012Z
Authorization: AWS4-HMAC-SHA256 Credential=*REDACTED*/20181025/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=*REDACTED*
```

Response:
 ``` HTTP/1.1 404
Cache-Control: max-age=31536000
Content-Type: text/plain; charset=utf-8
X-Content-Type-Options: nosniff
Date: Thu, 25 Oct 2018 09:40:13 GMT
Content-Length: 19
```

Now the exception is talking about this `4` thing which is most likely the `4` of `404`. I find this really interesting. The response clearly states that it is just plain text. So why is the client trying to parse this as XML? 

Digging in the code I found `MinioClient#execute` which always tries to parse the response body as XML. Except on a `HEAD` method (because no body).:
```
    // HEAD returns no body, and fails on parseXml
    if (!method.equals(Method.HEAD)) {
      Scanner scanner = new Scanner(response.body().charStream());
      try {
        scanner.useDelimiter("\\A");
        String errorXml = "";

        // read entire body stream to string.
        if (scanner.hasNext()) {
          errorXml = scanner.next();
        }

        errorResponse = new ErrorResponse(new StringReader(errorXml));
        if (this.traceStream != null) {
          this.traceStream.println(errorXml);
        }

      } finally {
        response.body().close();
        scanner.close();
      }
    }
```

Which could be rewritten to something like: 
```
    // HEAD returns no body, and fails on parseXml
    if (!method.equals(Method.HEAD)) {
      try (String boy = response.body().string()) {
[...]

PR #771, opened 2019-05-27
Add content negotiation

sinhaashish:
Client only parses when the content-type of  response is application/xml
Fixes https://github.com/minio/minio-java/issues/718

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -33,6 +33,7 @@ import io.minio.errors.InvalidEndpointException;
 import io.minio.errors.InvalidExpiresRangeException;
 import io.minio.errors.InvalidObjectPrefixException;
 import io.minio.errors.InvalidPortException;
+import io.minio.errors.InvalidResponseException;
 import io.minio.errors.NoResponseException;
 import io.minio.errors.RegionConflictException;
 import io.minio.http.HeaderParser;
@@ -1051,7 +1052,7 @@ public class MinioClient {
                                Object body, int length)
     throws InvalidBucketNameException, NoSuchAlgorithmException, InsufficientDataException, IOException,
            InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
-           InternalException {
+           InternalException, InvalidResponseException {
 
     if (headerMap != null) {
       headerMap = normalizeHeaders(headerMap);
@@ -1075,7 +1076,7 @@ public class MinioClient {
                                Object body, int length)
     throws InvalidBucketNameException, NoSuchAlgorithmException, InsufficientDataException, IOException,
            InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
-           InternalException {
+           InternalException, InvalidResponseException {
     String contentType = null;
     if (headerMap != null && headerMap.get("Content-Type") != null) {
       contentType = String.join(" ", headerMap.get("Content-Type"));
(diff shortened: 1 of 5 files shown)


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1138,7 +1138,10 @@ private HttpResponse executeReq(Method method, String region, String bucketName,
           errorXml = scanner.next();
         }
 
-        errorResponse = new ErrorResponse(new StringReader(errorXml));
+        if ("application/xml".equals(response.headers().get("content-type"))) {
+          errorResponse = new ErrorResponse(new StringReader(errorXml));
+        }
```
balamurugana: now you are swallowing non-xml response.  you would need to throw an exception about non-xml response received.

sinhaashish: Done.

sinhaashish:
@balamurugana @ebozduman  PTAL

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1138,7 +1138,12 @@ private HttpResponse executeReq(Method method, String region, String bucketName,
           errorXml = scanner.next();
         }
 
-        errorResponse = new ErrorResponse(new StringReader(errorXml));
+        if ("application/xml".equals(response.headers().get("content-type"))) {
+          errorResponse = new ErrorResponse(new StringReader(errorXml));
+        } else {
+          errorResponse = new ErrorResponse("Non-Xml Response Received");
```
balamurugana: This is incorrect usage or addition. Each class and error has its own purpose. You would need a new exception `InvalidResponseException` like `NoResponseException`.

Go through javadoc http://minio.github.io/minio-java/ completely.

api/src/main/java/io/minio/messages/ErrorResponse.java
```diff
@@ -64,6 +64,15 @@ public ErrorResponse(Reader reader) throws IOException, XmlPullParserException {
     this.parseXml(reader);
   }
 
+  /**
+   * Constructs a new ErrorResponse object in case of non-xml response.
+   */
+  public ErrorResponse(String message) throws IOException, XmlPullParserException {
```
balamurugana: This constructor is not used and not required.

sinhaashish: Removed.

api/src/main/java/io/minio/messages/ErrorResponse.java
```diff
@@ -64,7 +64,6 @@ public ErrorResponse(Reader reader) throws IOException, XmlPullParserException {
     this.parseXml(reader);
   }
 
-
```
balamurugana: Make zero diff in this file.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1138,7 +1139,12 @@ private HttpResponse executeReq(Method method, String region, String bucketName,
           errorXml = scanner.next();
         }
 
-        errorResponse = new ErrorResponse(new StringReader(errorXml));
+        if ("application/xml".equals(response.headers().get("content-type"))) {
+          errorResponse = new ErrorResponse(new StringReader(errorXml));
+        } else {
+          throw new InvalidResponseException();
```
balamurugana: Do a negative logic and you don't need `else` part.

api/src/main/java/io/minio/messages/ErrorResponse.java
sinhaashish: done.

api/src/main/java/io/minio/MinioClient.java
sinhaashish: Modified.

Issue #772, opened 2019-05-27
Dependency conflicts in project.


ZawadaZSE:
Hi,

I've just noticed that there are some dependency conflicts in this project resolved automatically by gradle. It would be nice to resolve those conflicts which are internal (okio), force external version conflicts (Jackson, JSR) and so it won't happen again accidentally fail the build in case of version mismatch.

```
compile - Dependencies for source set 'main' (deprecated, use 'implementation ' instead).
+--- com.google.http-client:google-http-client-xml:1.20.0
|    +--- com.google.http-client:google-http-client:1.20.0
|    |    +--- com.google.code.findbugs:jsr305:1.3.9 -> 3.0.2
|    |    \--- org.apache.httpcomponents:httpclient:4.0.1
|    |         +--- org.apache.httpcomponents:httpcore:4.0.1
|    |         +--- commons-logging:commons-logging:1.1.1
|    |         \--- commons-codec:commons-codec:1.3
|    \--- xpp3:xpp3:1.1.4c
+--- com.google.guava:guava:25.1-jre
|    +--- com.google.code.findbugs:jsr305:3.0.2
|    +--- org.checkerframework:checker-qual:2.0.0
|    +--- com.google.errorprone:error_prone_annotations:2.1.3
|    +--- com.google.j2objc:j2objc-annotations:1.1
|    \--- org.codehaus.mojo:animal-sniffer-annotations:1.14
+--- com.squareup.okhttp3:okhttp:3.13.1
|    \--- com.squareup.okio:okio:1.17.2 -> 1.17.3
+--- com.squareup.okio:okio:1.17.3
+--- joda-time:joda-time:2.7
+--- com.fasterxml.jackson.core:jackson-annotations:2.9.6
+--- com.fasterxml.jackson.core:jackson-core:2.9.6
+--- com.fasterxml.jackson.core:jackson-databind:2.9.6
|    +--- com.fasterxml.jackson.core:jackson-annotations:2.9.0 -> 2.9.6
|    \--- com.fasterxml.jackson.core:jackson-core:2.9.6
+--- com.google.code.findbugs:annotations:3.0.1
|    +--- net.jcip:jcip-annotations:1.0
|    \--- com.google.code.findbugs:jsr305:3.0.1 -> 3.0.2
\--- com.google.code.findbugs:jsr305:3.0.1 -> 3.0.2
```

harshavardhana:
feel free to send a PR @ZawadaZSE

[...]

PR #773, opened 2019-05-27
Fix dependency mismatch. Enforce failing on version confilict.

ZawadaZSE:
Fix for: #772

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,18 +52,18 @@ subprojects {
     }
 
     dependencies {
-        compile "com.google.http-client:google-http-client-xml:1.20.0"
+        compile "com.google.http-client:google-http-client-xml:1.24.1"
         compile "com.google.guava:guava:25.1-jre"
         compile "com.squareup.okhttp3:okhttp:3.13.1"
-        compile "com.squareup.okio:okio:1.17.3"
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -52,18 +52,18 @@ subprojects {
     }
 
     dependencies {
-        compile "com.google.http-client:google-http-client-xml:1.20.0"
+        compile "com.google.http-client:google-http-client-xml:1.24.1"
         compile "com.google.guava:guava:25.1-jre"
         compile "com.squareup.okhttp3:okhttp:3.13.1"
-        compile "com.squareup.okio:okio:1.17.3"
+        compile "com.squareup.okio:okio:1.17.2"
```
balamurugana: Any reason to downgrade `okio`?

ZawadaZSE: Well, okhttp is using okio 1.17.2 I thought that it would be better to downgrade than have newer version forced.

balamurugana: ok. if possible we could move to okhttp 3.14.2

ZawadaZSE: I've checked it also with that version and tests were green but I didn't want to change too much in this commit. Moreover, 3.14.2 also uses 1.17.2 version of okio.

PR #777, opened 2019-07-15
fix: handle SSE in copyObject() properly and have one source code.

balamurugana:
Added new version of copyObject(). Note: this method makes source and
destination argument swapping correctly to denote on which bucket and
object the request is sent.
    
```
public void copyObject(String bucketName, String objectName, Map<String,String> headerMap, ServerSideEncryption sse,
                       String srcBucketName, String srcObjectName, ServerSideEncryption srcSse,
                       CopyConditions copyConditions)
```

PR #776, opened 2019-07-15
fix: handle SSE in getObject() properly and have one source code.

balamurugana:
Added below version of getObject() for completion.
```
public InputStream getObject(String bucketName, String objectName, Long offset, Long length,
                             ServerSideEncryption sse)
```

PR #775, opened 2019-07-15
fix: handle SSE in statObject() properly and have one source code.

balamurugana:

PR #778, opened 2019-07-16
fix: rearrange, cleanup, deprecation and fixes in ServerSideEncryption

balamurugana:
1. Cleanup formatting issues.
2. Arrange code usage together.
3. Deprecate `getType()` and `marshal()`, `ServerSideEncryptionCopyWithCustomerKey` and
   `copyWithCustomerKey()`. Use `type()`, `headers()`, `copySourceHeaders()` and `withCustomerKey()`.

### DIFF
--- a/api/src/main/java/io/minio/ServerSideEncryption.java
+++ b/api/src/main/java/io/minio/ServerSideEncryption.java
@@ -20,6 +20,7 @@ import java.io.UnsupportedEncodingException;
 import java.security.InvalidKeyException;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
@@ -28,7 +29,7 @@ import javax.crypto.SecretKey;
 import javax.security.auth.DestroyFailedException;
 import javax.security.auth.Destroyable;
 
-import  com.google.common.io.BaseEncoding;
+import com.google.common.io.BaseEncoding;
 
 import io.minio.errors.InvalidArgumentException;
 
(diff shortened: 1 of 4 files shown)


api/src/main/java/io/minio/ServerSideEncryption.java
```diff
@@ -37,209 +38,203 @@
...
+    protected final MessageDigest md5;
+
     public ServerSideEncryptionWithCustomerKey(SecretKey key, MessageDigest md5) {
       this.secretKey = key;
       this.md5 = md5;
     }
 
     @Override
-    public final Type getType() { 
+    public final Type type() {
```
aead: Java uses getter / setter for properties. Renaming is not idiomatic here.

api/src/main/java/io/minio/ServerSideEncryption.java
```diff
@@ -37,209 +38,203 @@
...
     @Override
-    public final Type getType() { 
+    public final Type type() { 
       return Type.SSE_C; 
     }
-   
+
     @Override
-    public void marshal(Map<String, String> headers) {
+    public Map<String, String> headers() {
```
aead: Why do we return a new HashMap instead of adding the headers to an existing one? The typical use case is to add the SSE-C headers to the HTTP header map. What do we gain here?

api/src/main/java/io/minio/ServerSideEncryption.java
```diff
@@ -37,209 +38,203 @@
...
+      throw new InvalidKeyException("The secret key is not a 256 bit AES key");
     }
+    return new ServerSideEncryptionCopyWithCustomerKey(key, MessageDigest.getInstance(("MD5")));
   }
 
   static final class ServerSideEncryptionS3 extends ServerSideEncryption {
     @Override
-    public final Type getType() { 
-      return Type.SSE_S3; 
+    public final Type type() {
```
aead: Same

api/src/main/java/io/minio/ServerSideEncryption.java
```diff
@@ -37,209 +38,203 @@
...
+      return Type.SSE_S3;
     }
 
     @Override
-    public final void marshal(Map<String, String> headers) { 
-      headers.put("X-Amz-Server-Side-Encryption", "AES256"); 
+    public final Map<String, String> headers() {
```
aead: Same

api/src/main/java/io/minio/ServerSideEncryption.java
```diff
@@ -248,8 +243,16 @@ public final void destroy() throws DestroyFailedException {
     }
   }
 
-  static final class ServerSideEncryptionKms extends ServerSideEncryption {
+  /**
+   * Create a new server-side-encryption object for encryption at rest (a.k.a. SSE-S3).
+   *
+   * @return an instance of ServerSideEncryption implementing SSE-S3
+   */
+  public static ServerSideEncryption atRest() {
```
aead: We should keep AWS S3 naming conventions.

api/src/main/java/io/minio/ServerSideEncryption.java
aead: Should this be `final`?

api/src/main/java/io/minio/ServerSideEncryption.java
balamurugana: It is old style and minio-java follows this style inspired from okhttp.

api/src/main/java/io/minio/ServerSideEncryption.java
balamurugana: Returning `Map` gives full control to users how they will be manipulated, `marshal(Map)` mutates passed values which we need to avoid for future failures.

api/src/main/java/io/minio/ServerSideEncryption.java
balamurugana: I haven't changed any name here. Whatever renames I made have `deprecated` annotation.

api/src/main/java/io/minio/ServerSideEncryption.java
balamurugana: Can be. As `Map` is returned and not owned by this object, it is OK without `final`.

aead: Not sure I understand what you mean. How is
```
Map<String, String> headers = new Map<>();
marshal(headers);
```
different from:
```
Map<String, String> headers = sse.headers();
```

apart from the situation that the the former case avoids a new map allocation when we serialize the SSE object into the HTTP headers - which is the entire purpose of this function. In the later case you crate a new map and "force" an explicit copy. So, in general both options are equally powerful just that we usually want to serialize to the HTTP header map of the request and not create a new map which we have to copy explicitly - or am I missing something.

Anyway minor.

api/src/main/java/io/minio/ServerSideEncryption.java
aead: Okay, can/have we document that? Because a **lot** of Java code follows the "style" I mentioned...

api/src/main/java/io/minio/ServerSideEncryption.java
aead: Oh sry, my bad.

api/src/main/java/io/minio/ServerSideEncryption.java
aead: Just curious because all other `public Map<String, String> headers()` are `final`.

balamurugana: Fixed

api/src/main/java/io/minio/ServerSideEncryption.java
balamurugana: There is no coding standard document we have. Modern style is followed from the beginning of minio-java. There is little different with getter style where getters may return reference of an class member but this style returns copy of a value.

api/src/main/java/io/minio/ServerSideEncryption.java
balamurugana: There is a little difference. `marshal()` mutates header map passed, there could be a possible value loss if header map contains any such header. Whereas `headers()` returns a new copy of SSE headers every time and caller has a control when/how to use returned headers.  As this is public class/method, it is not a good idea to mutate a passed header map.

api/src/main/java/io/minio/ServerSideEncryption.java
aead: Okay

api/src/main/java/io/minio/ServerSideEncryption.java
aead: > There is a little difference. marshal() mutates header map passed, there could be a possible value loss if header map contains any such header.

That I got. I still don't see how this is different from `headers`. Sure, you make a copy but callers can do that explicitly by passing a new map when they want to avoid overwriting with `marshal`.  So getting the same behavior as `headers` like shown above. Anyway, the only purpose of this function is  to serialize the object to the HTTP request. So you have a S3 API call:
```
putObject(...) {
   req = new Request();
   req.getHeaders().add("Content-MD5", ...);
   //...   
  sse.marshal(req.getHeaders()); // With new method: req.getHeaders().addAll(sse.headers)
}
```
> As this is public class/method, it is not a good idea to mutate a passed header map.

balamurugana:
Depends on https://github.com/minio/minio-java/pull/775 https://github.com/minio/minio-java/pull/776 https://github.com/minio/minio-java/pull/777

kannappanr:
ping @aead @sinhaashish

PR #780, opened 2019-07-26
Preserve all path segments appropriately for objectNames

harshavardhana:


functional/FunctionalTest.java
```diff
@@ -515,7 +515,7 @@ public static void putObject_test6() throws Exception {
 
     long startTime = System.currentTimeMillis();
     try {
-      String objectName = "/path/to/" + getRandomName();
```
balamurugana: '/path/to/` is also a valid input according to S3 specification. You would need to add a new test case.

harshavardhana: It is not valid for MinIO @balamurugana also we never preserved the "/" in the beginning. We happily trimmed all "/". So we shall keep the test case that works with MinIO as expected.

functional/FunctionalTest.java
```diff
@@ -1007,6 +1007,32 @@ public static void statObject_test2() throws Exception {
     }
   }
 
+  /**
+   * Test: statObject(String bucketName, '/').
+   */
+  public static void statObject_test3() throws Exception {
+    if (!mintEnv) {
+      System.out.println("Test: statObject(String bucketName, '/')");
```
balamurugana: `'/'` is a character, not string.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -902,14 +902,7 @@ private Request createRequest(Method method, String bucketName, String objectNam
...
     if (objectName != null) {
-      // Limitation: OkHttp does not allow to add '.' and '..' as path segment.
-      for (String pathSegment : objectName.split("/")) {
-        urlBuilder.addEncodedPathSegment(S3Escaper.encode(pathSegment));
-      }
-      if (objectName.endsWith("/")) {
-        // Fix issue #648: preserve trailing '/' to work with proxies like nginx.
-        urlBuilder.addPathSegment("");
-      }
+      urlBuilder.addEncodedPathSegments(S3Escaper.encodePath(objectName));
```
balamurugana: You would need to retain comment about '.' and '..` where they are valid accordingly to S3 specification.

functional/FunctionalTest.java
balamurugana: What I understand is, minio server does not support `/path/to/`. If true, my comment can be ignored.

harshavardhana: Your understanding is correct.

api/src/main/java/io/minio/MinioClient.java
harshavardhana: What would be the purpose of that comment? @balamurugana MinIO server rejects path segments with `.` and `..` - do we still need to keep it?

functional/FunctionalTest.java
harshavardhana: Fixing

api/src/main/java/io/minio/MinioClient.java
balamurugana: Because `.` and `..` are valid as per specification. As this library can be used with AWS S3 where it will fail and developer needs to know why it fails for such bug report.

harshavardhana: Done

harshavardhana:
ping @sinhaashish @ebozduman

PR #741, opened 2018-12-19
Remove resuming multipart uploads

harshavardhana:
ListParts is never fully complete in its output, there is always a
possibility that the List is partial where another
PutObjectPart() is being uploaded in parallel.

While it may seem like ListParts and ListMultipartUploads
provide the possibility of resuming an upload, it would
be a mistake as these APIs were inherently never meant do
these things. This is one of the reasons why Amazon
recommends that one should avoid using these API calls to
complete or add parts to an upload - they instead
recommend to remember the parts being uploaded at the client.

PR #786, opened 2019-08-06
Abort failed multipart uploads on RuntimeException

Dorthu:
If a file > 5MB is being uploaded to a bucket, this client will initiate
a multipart upload.  However, if in the course of uploading that file a
RuntimeException occurs, this library will not abort the multipart
upload.  This is especially possible when streaming data across a
network connection into a bucket.  This was introduced in #741 when
resuming failed multipart uploads was removed.

Reproduction steps:

* Use this script: https://gist.github.com/Dorthu/b75f2a4ad9e0e453a040412438d3fe32
* Replace the server URL and port, access key, secret key, and bucket name with
  valid values
* Compile/run the script
* Observe the state of multipart uploads in the bucket after running
  with `s3cmd multipart s3://bucket-name` (or another tool)

Subsequent runs of the script create new multipart uploads in the
bucket.

I specifically noticed this bug when using this library against Ceph, if
that's relevant.

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -4414,6 +4414,7 @@ public class MinioClient {
       // All parts have been uploaded, complete the multipart upload.
       completeMultipart(bucketName, objectName, uploadId, totalParts);
     } catch (RuntimeException e) {
+      abortMultipartUpload(bucketName, objectName, uploadId);
       throw e;
     } catch (Exception e) {
       abortMultipartUpload(bucketName, objectName, uploadId);


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -4414,6 +4414,7 @@ private void putObject(String bucketName, String objectName, Long size, Object d
       // All parts have been uploaded, complete the multipart upload.
       completeMultipart(bucketName, objectName, uploadId, totalParts);
     } catch (RuntimeException e) {
+      abortMultipartUpload(bucketName, objectName, uploadId);
```
balamurugana: Does `catch (Exception e)` catch `RuntimeException`?  If yes, you would need to remove `catch (RuntimeException e)`.

Dorthu: I thought it did, but I assumed they were caught separately for a reason.  I'll re-test with one catch block and update this if it does the same thing.

Dorthu: Ah, when building this without the `catch (RuntimeException e)` findbugs raises a `REC_CATCH_EXCEPTION` error:

> REC_CATCH_EXCEPTION: Exception is caught when Exception is not thrown
This method uses a try-catch block that catches Exception objects, but Exception is not thrown within the try block, and RuntimeException is not explicitly caught. It is a common bug pattern to say try { ... } catch (Exception e) { something } as a shorthand for catching a number of types of exception each of whose catch blocks is identical, but this construct also accidentally catches RuntimeException as well, masking potential bugs.
>
> A better approach is to either explicitly catch the specific exceptions that are thrown, or to explicitly catch RuntimeException exception, rethrow it, and then catch all non-Runtime Exceptions, as shown below:
>
>  try {
>    ...
>  } catch (RuntimeException e) {
>    throw e;
>  } catch (Exception e) {
>    ... deal with all non-runtime exceptions ...
>  }

I suspect this is why they were both caught before.  I could remove the `catch (Exception e)` but I'm not sure what that would let through.

balamurugana: Removing `catch (Exception e)` is not correct. But removing `catch (RuntimeException e)` is more appropriate, but I don't know why it was added. Its better to leave as is.

PR #787, opened 2019-08-19
Add function to compose object

sinhaashish:
Creates a new object by combining different Source Objects .

### DIFF
--- a/api/src/main/java/io/minio/ComposeSource.java
+++ b/api/src/main/java/io/minio/ComposeSource.java
@@ -1,0 +1,188 @@
+/*
+ * MinIO Java SDK for Amazon S3 Compatible Cloud Storage,
+ * (C) 2019 MinIO, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.minio;
+
+import java.util.Map;
+import java.util.TreeMap;
+import io.minio.errors.InvalidArgumentException;
+import io.minio.CopyConditions;
+
+public class ComposeSource {
+  private String bucketName;
+  private String objectName;
+  private Long offset;
+  private Long length;
+  private Map<String, String> headerMap;
+  private CopyConditions copyConditions;
+  private ServerSideEncryption sse;
+  private long objectSize;
+  private Map<String, String> headers;
+
+  /**
+   * Create new ComposeSource for given bucket and object.
+   */
+  public ComposeSource(String bucketName, String objectName) throws InvalidArgumentException {
+    this(bucketName, objectName, null, null, null, null, null);
+  }
+
+  /**
+   * Create new ComposeSource for given bucket, object, offset and length.
+   */
+  public ComposeSource(String bucketName, String objectName, Long offset, Long length)
+    throws InvalidArgumentException {
+    this(bucketName, objectName, offset, length, null, null, null);
+  }
+
+  /**
+   * Create new ComposeSource for given bucket, object, offset, length and headerMap.
+   */
+  public ComposeSource(String bucketName, String objectName, Long offset, Long length,
+      Map<String, String> headerMap) throws InvalidArgumentException {
+    this(bucketName, objectName, offset, length, headerMap, null, null);
+  }
+
+  /**
+   * Create new ComposeSource for given bucket, object, offset, length, headerMap and CopyConditions.
+   */
+  public ComposeSource(String bucketName, String objectName, Long offset, Long length,
+      Map<String, String> headerMap, CopyConditions copyConditions) throws InvalidArgumentException {
+    this(bucketName, objectName, offset, length, headerMap, copyConditions, null);
+  }
+
+  /**
+   * Creates new ComposeSource for given bucket, object, offset, length, headerMap, CopyConditions
+   * and server side encryption.
+   *
+   * @throws InvalidArgumentException    upon invalid value is passed to a method.
+   */
+  public ComposeSource(String bucketName, String objectName, Long offset, Long length, Map<String, String> headerMap,
+      CopyConditions copyConditions, ServerSideEncryption sse) throws InvalidArgumentException {
+    if (bucketName == null) {
+      throw new InvalidArgumentException("Source bucket name cannot be empty");
+    }
+
+    if (objectName == null) {
+      throw new InvalidArgumentException("Source object name cannot be empty");
+    }
+
+    if (offset != null && offset < 0) {
+      throw new InvalidArgumentException("Offset cannot be negative");
+    }
+
+    if (length != null && length < 0) {
+      throw new InvalidArgumentException("Length cannot be negative");
+    }
+
+    if (length != null && offset == null) {
+      offset = 0L;
+    }
+
+    this.bucketName = bucketName;
+    this.objectName = objectName;
+    this.offset = offset;
+    this.length = length;
+    this.headerMap = headerMap;
+    this.copyConditions = copyConditions;
+    this.sse = sse;
+  }
+
+  /**
+   * Constructs header  .
+   *
+   */
+  public void buildHeaders(long objectSize, String etag) throws InvalidArgumentException {
+    if (offset != null && offset >= objectSize) {
+      throw new InvalidArgumentException("source " + bucketName + "/" + objectName + ": offset " + offset
+        + " is beyond object size " + objectSize);
+    }
+
+    if (length != null) {
+      if (length > objectSize) {
+        throw new InvalidArgumentException("source " + bucketName + "/" + objectName + ": length " + length
+          + " is beyond object size " + objectSize);
+      }
+
+      if (offset + length > objectSize) {
+        throw new InvalidArgumentException("source " + bucketName + "/" + objectName + ": compose size "
+          + (offset + length) + " is beyond object size " + objectSize);
+      }
+    }
+
+    Map<String, String> headers = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+    headers.put("x-amz-copy-source", S3Escaper.encodePath(bucketName + "/" + objectName));
+    headers.put("x-amz-copy-source-if-match", etag);
+
+    if (headerMap != null) {
+      headers.putAll(headerMap);
+    }
+
+    if (copyConditions != null) {
+      headers.putAll(copyConditions.getConditions());
+    }
+
+    if (sse != null) {
+      headers.putAll(sse.copySourceHeaders());
+    }
+
+    this.objectSize = objectSize;
+    this.headers = headers;
+  }
+
+  public String bucketName() {
+    return bucketName;
+  }
+
+  public String objectName() {
+    return objectName;
+  }
+
+  public Long offset() {
+    return offset;
+  }
+
+  public Long length() {
+    return length;
+  }
+
+  public CopyConditions copyConditions() {
+    return copyConditions;
+  }
+
+  public ServerSideEncryption sse() {
+    return sse;
+  }
+
+  /**
+   * Returns header.
+   *
+   */
+  public Map<String, String> headers() {
+    Map<String, String> headers = null;
+
+    if (this.headers != null) {
+      headers = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+      headers.putAll(this.headers);
+    }
+
+    return headers;
+  }
+
+  public long objectSize() {
+    return objectSize;
+  }
+}
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -42,6 +42,7 @@ import io.minio.http.Scheme;
 import io.minio.messages.Bucket;
 import io.minio.messages.CompleteMultipartUpload;
 import io.minio.messages.CopyObjectResult;
+import io.minio.messages.CopyPartResult;
 import io.minio.messages.CreateBucketConfiguration;
 import io.minio.messages.DeleteError;
 import io.minio.messages.DeleteObject;
@@ -160,6 +161,8 @@ public class MinioClient {
   private static final long DEFAULT_CONNECTION_TIMEOUT = 15 * 60;
   // maximum allowed object size is 5TiB
   private static final long MAX_OBJECT_SIZE = 5L * 1024 * 1024 * 1024 * 1024;
+  // maxPartSize - maximum part size 5GiB for a single multipart upload operation
+  private static final long MAX_PART_SIZE = 5L * 1024 * 1024 * 1024;
   private static final int MAX_MULTIPART_COUNT = 10000;
   // minimum allowed multipart size is 5MiB
   private static final int MIN_MULTIPART_SIZE = 5 * 1024 * 1024;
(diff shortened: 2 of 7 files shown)


api/src/main/java/io/minio/ComposeSource.java
```diff
@@ -0,0 +1,189 @@
...
+    if (offset != null && offset < 0) {
+      throw new InvalidArgumentException("Offset cannot be negative");
+    }
+
+    if (length != null && length < 0) {
+      throw new InvalidArgumentException("Length cannot be negative");
+    }
+
+    if (length != null && offset == null) {
+      offset = 0L;//Long.valueOf(0);//new Long(0);
```
balamurugana: remove comment

api/src/main/java/io/minio/ComposeSource.java
```diff
@@ -0,0 +1,189 @@
...
+      headers.putAll(headerMap);
+    }
+
+    if (copyConditions != null) {
+      headers.putAll(copyConditions.getConditions());
+    }
+
+    if (sse != null) {
+//      sse.marshal(headers);
+      headers = sse.headers();
```
balamurugana: 1. remove unnecessary comments
2. you need to use `sse.copySourceHeaders()`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2383,6 +2386,226 @@ public void copyObject(String bucketName, String objectName, Map<String,String>
...
+    if (sources.isEmpty()) {
+      throw new InvalidArgumentException("compose sources cannot be empty");
+    }
+
+    if (sse != null && !this.baseUrl.isHttps()) {
+      if (sse.type() == ServerSideEncryption.Type.SSE_C) {
+        throw new InvalidArgumentException(
+          "SSE_C operations must be performed over a secure connection.");
+      } else if (sse.type() == ServerSideEncryption.Type.SSE_KMS ) {
+        throw new InvalidArgumentException("SSE_KMS operations must be performed over a secure connection.");
```
balamurugana: Use `checkReadRequestSse()` and `checkWriteRequestSse()`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2383,6 +2386,226 @@ public void copyObject(String bucketName, String objectName, Map<String,String>
...
+    if (headerMap == null) {
+      headerMap = new HashMap<>();
+    }
+
+    if (sse != null) {
+
+//
+//      sse.marshal(headerMap);
+      checkWriteRequestSse(sse);
+      headerMap.putAll(sse.headers());
```
balamurugana: No need to  `headerMap = new HashMap<>();`, you could do 
```java
if (sse != null) {
   headerMap = sse.headers();
}
```

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2383,6 +2386,226 @@ public void copyObject(String bucketName, String objectName, Map<String,String>
...
+
+          totalParts[partNumber - 1] = new Part(partNumber, eTag);
+
+          offset = startBytes;
+          size -= (endBytes - startBytes);
+        }
+      }
+
+      completeMultipart(bucketName, objectName, uploadId, totalParts);
+    } catch (RuntimeException e) {
```
balamurugana: add `abortMultipartUpload(bucketName, objectName, uploadId);`

api/src/main/java/io/minio/MinioClient.java
sinhaashish: In case sse is null  `headerMap` needs to be initialled as `initMultipartUpload` gets the Content-type  `headerMap.get("Content-Type") == null` . If the map is not initialised it causes a null pointer exception.

balamurugana: As per https://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadInitiate.html, `Content-Type` is an optional header. The check can be removed in `initMultipartUpload()` or you could deal with `null` `headerMap` in `initMultipartUpload()` itself which is clean.

api/src/main/java/io/minio/ComposeSource.java
sinhaashish: done.

api/src/main/java/io/minio/ComposeSource.java
sinhaashish: done.

api/src/main/java/io/minio/MinioClient.java
sinhaashish: done.

api/src/main/java/io/minio/MinioClient.java
sinhaashish: `headerMap = new HashMap<>();` is needed now as `uploadPartCopy` takes headerMap which contains encryption headers for source as well as destination.  In order to populate `headerMap` with source `headers` it has been initialised.

api/src/main/java/io/minio/MinioClient.java
sinhaashish: done.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2383,6 +2386,218 @@ public void copyObject(String bucketName, String objectName, Map<String,String>
...
+        if (size <= MAX_PART_SIZE) {
+          partNumber++;
+
+          Map<String, String> headers = src.headers();
+          if (src.length() != null) {
+            headers.put("x-amz-copy-source-range", "bytes=" + offset + "-" + (offset + src.length()));
+          } else if (src.offset() != null) {
+            headers.put("x-amz-copy-source-range", "bytes=" + offset + "-" + (offset + size - 1));
+          }
+          headerMap.putAll(headers);
```
balamurugana: `headerMap` is wrongly used and incrementally added up if `CopySource.headers()` varies in keys. You would need to have new headers on every iteration, and you don't need to set user headers for `uploadPartCopy()`

```java
          Map<String, String> headers = src.headers();
          if (src.length() != null) {
            headers.put("x-amz-copy-source-range", "bytes=" + offset + "-" + (offset + src.length()));
          } else if (src.offset() != null) {
            headers.put("x-amz-copy-source-range", "bytes=" + offset + "-" + (offset + size - 1));
          }
          headers.putAll(sseHeaders);
```

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2383,6 +2386,218 @@ public void copyObject(String bucketName, String objectName, Map<String,String>
...
+          partNumber++;
+
+          long startBytes = offset;
+          long endBytes = startBytes + MAX_PART_SIZE;
+          if (size < MAX_PART_SIZE) {
+            endBytes = startBytes + size;
+          }
+
+          Map<String, String> headers = src.headers();
+          headers.put("x-amz-copy-source-range", "bytes=" + startBytes + "-" + endBytes);
```
balamurugana: You need to add SSE headers here.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2383,6 +2386,218 @@ public void copyObject(String bucketName, String objectName, Map<String,String>
...
+    }
+
+    if (partsCount == 1) {
+      ComposeSource src = sources.get(0);
+      copyObject(bucketName, objectName, headerMap, sse, src.bucketName(), src.objectName(), src.sse(),
+          src.copyConditions());
+      return;
+    }
+    if (headerMap == null) {
+      headerMap = new HashMap<>();
```
balamurugana: This `if` block is not needed

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2383,6 +2386,218 @@ public void copyObject(String bucketName, String objectName, Map<String,String>
...
+    }
+
+    if (sse != null) {
+      headerMap = sse.headers();
+    }
```
balamurugana: Have sse headers defined
```java
Map<String, String> sseHeaders = null;
if (sse != null) {
  sseheaders = sse.headers();
  if (headerMap == null) {
    headerMap = new HashMap<>();
  }
  headerMap.putAll(sseHeaders);
}
```

api/src/main/java/io/minio/ComposeSource.java
```diff
@@ -0,0 +1,188 @@
...
+        + " is beyond object size " + objectSize);
+    }
+
+    if (length != null) {
+      if (length >= objectSize) {
+        throw new InvalidArgumentException("source " + bucketName + "/" + objectName + ": length " + length
+          + " is beyond object size " + objectSize);
+      }
+
+      if (offset + length > objectSize) {
```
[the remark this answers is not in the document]
balamurugana: Can you explain why we need to error out if `offset+length == objectSize`?

api/src/main/java/io/minio/ComposeSource.java
```diff
@@ -0,0 +1,189 @@
...
+        throw new InvalidArgumentException("source " + bucketName + "/" + objectName + ": compose size "
```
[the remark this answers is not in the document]
balamurugana: A size calculated by `(offset + length)` is a size of destination object or object part.  Its hard to understand `offset + length` text alone. For more detailed information, we would need to give `offset: $offset` and `length: $length` and destination size `compose size: $composeSize` which is too verbose IMO

functional/FunctionalTest.java
```diff
@@ -2561,6 +2561,248 @@ public static void copyObject_test12() throws Exception {
...
+      System.out.println("Test: composeObject(String bucketName, String objectName,List<ComposeSource> composeSources, "
+          + "Map <String,String > headerMap, ServerSideEncryption sseTarget) with offset and length.");
+    }
+    long startTime = System.currentTimeMillis();
+
+    try {
+      String destinationObjectName = getRandomName();
+      String filename1 = createFile6Mb();
+      String filename2 = createFile6Mb();
+      client.putObject(bucketName, filename1, filename1, null, null, null, null);
```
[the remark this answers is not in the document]
balamurugana: `putObject(String bucketName, String objectName, String fileName)` is deprecated.

functional/FunctionalTest.java
```diff
@@ -2561,6 +2561,248 @@ public static void copyObject_test12() throws Exception {
     }
   }
 
+  /**
+   * Test: composeObject(String bucketName, String objectName,
+   * List&lt;ComposeSource&gt; composeSources,Map &lt;String, String&gt; headerMap, ServerSideEncryption sseTarget).
+   */
+  public static void composeObject_test1() throws Exception {
+    if (!mintEnv) {
+      System.out.println("Test: composeObject(String bucketName, String objectName,List<ComposeSource> composeSources, "
```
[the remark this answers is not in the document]
balamurugana: This is out of scope of this PR IMO.  You could check how existing tests are handling it.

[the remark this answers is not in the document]
sinhaashish: You can find the difference in the test cases as all the test cases contains different information.
Example.
``` 
Test: composeObject(String bucketName, String objectName,List<ComposeSource> composeSources, Map <String,String > headerMap, ServerSideEncryption sseTarget).
Test: composeObject(String bucketName, String objectName,List<ComposeSource> composeSources, Map <String,String > headerMap, ServerSideEncryption sseTarget) with offset and length.
```

docs/API.md
```diff
@@ -2762,6 +2762,58 @@ try {
 }
 ```
 
+```
+ public void composeObject(String bucketName, String objectName, List<ComposeSource> sources,
```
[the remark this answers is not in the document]
sinhaashish: done

docs/API.md
```diff
@@ -2762,6 +2762,58 @@ try {
 }
 ```
 
+```
```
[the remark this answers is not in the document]
sinhaashish: done.

api/src/main/java/io/minio/MinioClient.java
sinhaashish: done.

api/src/main/java/io/minio/MinioClient.java
sinhaashish: removed.

api/src/main/java/io/minio/MinioClient.java
sinhaashish: done.

api/src/main/java/io/minio/MinioClient.java
sinhaashish: added

balamurugana: Nope. You haven't added SSE headers here.

functional/FunctionalTest.java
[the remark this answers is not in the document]
ebozduman: OK

functional/FunctionalTest.java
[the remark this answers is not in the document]
ebozduman: OK

api/src/main/java/io/minio/ComposeSource.java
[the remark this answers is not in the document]
balamurugana: Generally `offset` is an integer (including negative numbers) refers an index and `length` is a non-negative integer. If `size` is 10, `offset = 4` and `length = 2` refers two bytes from 4th index whereas `offset = -4` and `length = 2` refers two bytes from end of 4th index i.e. `offset = 6`. `length=0` is a legitimate value which refers zero bytes.

In `ComposeSource`
1. If `offset`/`length` is non-`null`, it must be >= 0 i.e. we do not support negative offset.

2. `offset`/`length` can be `null`.  `offset=null` means `offset=0` and `length=null` means `length=size-offset`. This validation is done in the constructor.

3. There is no way to get object size (`size`) in the constructor to calculate `length=size-offset` when `length=null` because the size is unknown.

4. At the time of `buildHeaders()`, `size` and `etag` are passed from `composeObject()` in `MinioClient` because `MinioClient` class is S3 client fetches `ObjectStat` for given `ComposeSource`.

5. `buildHeaders()` should calculate `length=size-offset` when `length=null`.  @sinhaashish This should be fixed. Note that its not required to set it to `this.length`.

6. S3 accepts byte range not offset/length and byte range are calculated like below
  * When `offset=null` and `length=null`, no byte range header is set.
  * When `offset=null` and `length=non-null`, byte range is `0-length`.
  * When `offset=non-null` and `length=null`, byte range is `offset-`.
  * When `offset=non-null` and `length=non-null`, byte range is `offset-length`.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2383,6 +2386,221 @@ public void copyObject(String bucketName, String objectName, Map<String,String>
...
+   * @throws InternalException           upon internal library error
+   * @throws InvalidArgumentException    upon invalid value is passed to a method.
+   * @throws InvalidResponseException    upon a non-xml response from server
+   */
+
+  public void composeObject(String bucketName, String objectName, List<ComposeSource> sources,
+      Map<String,String> headerMap, ServerSideEncryption sse)
+    throws InvalidBucketNameException, NoSuchAlgorithmException, InsufficientDataException, IOException,
+    InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
+    InternalException, InvalidArgumentException, InvalidResponseException {
```
ebozduman: `InvalidResponseException` is declared here as one of the possible exceptions `composeObject` can throw, but the `API.md` file does not explain about it, i.e. it is missing in the list under `Listed Exceptions:`.

api/src/main/java/io/minio/ComposeSource.java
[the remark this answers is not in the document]
sinhaashish: Adding a test case for this  using the object at `play/compose-object/15Lakh.csv`.
Test

api/src/main/java/io/minio/MinioClient.java
sinhaashish: done.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2383,6 +2386,225 @@ public void copyObject(String bucketName, String objectName, Map<String,String>
...
+
+      if (partsCount > MAX_MULTIPART_COUNT) {
+        throw new InvalidArgumentException("Compose sources create more than allowed multipart count "
+          + MAX_MULTIPART_COUNT);
+      }
+    }
+
+    if (partsCount == 1) {
+      ComposeSource src = sources.get(0);
+      copyObject(bucketName, objectName, headerMap, sse, src.bucketName(), src.objectName(), src.sse(),
```
balamurugana: If `src.offset() != null || src.length() != null`, you would need to handle add `x-amz-copy-source-range` appropriately.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2383,6 +2386,225 @@ public void copyObject(String bucketName, String objectName, Map<String,String>
...
+          }
+        }
+        long offset = 0;
+        if (src.offset() != null) {
+          offset = src.offset();
+        }
+
+        if (size <= MAX_PART_SIZE) {
+          partNumber++;
+          Map<String, String> headers  = (src.headers() == null) ? new HashMap<>() : src.headers();
```
balamurugana: use `if` than ternary operator i.e. avoid using it.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2383,6 +2386,225 @@ public void copyObject(String bucketName, String objectName, Map<String,String>
...
+          if (src.length() != null) {
+            headers.put("x-amz-copy-source-range", "bytes=" + offset + "-" + (offset + src.length() - 1));
+          } else if (src.offset() != null) {
+            headers.put("x-amz-copy-source-range", "bytes=" + offset + "-" + (offset + size ));
```
balamurugana: Because byte range values are indices, `offset + size - 1` is correct else you will end up copying one byte more.

sinhaashish: done

api/src/main/java/io/minio/MinioClient.java
sinhaashish: done.

api/src/main/java/io/minio/MinioClient.java
sinhaashish: done.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2383,6 +2386,240 @@ public void copyObject(String bucketName, String objectName, Map<String,String>
...
+      long size = stat.length();
+      if (src.length() != null) {
+        size = src.length();
+      } else {
+        if (src.offset() != null) {
+          size -= src.offset();
+        }
+      }
+
+      if (size < MIN_MULTIPART_SIZE && sources.size() != 1 && i != (sources.size() - 1)) {
```
ebozduman: @sinhaashish 
Could you explain this line for me, specifically the 2nd and 3rd conditions?

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2383,6 +2386,240 @@ public void copyObject(String bucketName, String objectName, Map<String,String>
...
+    Part[] totalParts = new Part[partsCount];
+    try {
+      for (int i = 0; i < sources.size(); i++) {
+        ComposeSource src = sources.get(i);
+
+        long size = src.objectSize();
+        if (src.length() != null) {
+          size = src.length();
+        } else {
+          if (src.offset() != null) {
```
ebozduman: Just a minor comment:
You can combine these 2 lines:
```
} else {
   if (src.offset() != null) {
```
=> `} else if (src.offset() != null) {`

api/src/main/java/io/minio/MinioClient.java
sinhaashish: Yes @ebozduman  . Its similar to `minio-go`  implementation [here ](https://github.com/minio/minio-go/blob/e11840609c348b5d1e46185b7f6e98ce6e6bc313/api-compose-object.go#L395)

ebozduman:
I want to copy/paste my question to Ashish on Slack and his answer, just for tracking and historical purposes, just in case someone comes up with the same question.
```
ersan 10:30 AM
Hey @ashish
I just noticed something in the code you recently changed and wanted to ask you on Slack,
not to wait for another day. Here is the code:
      long size = stat.length();
      if (src.length() != null) {
        size = src.length();
      } else if (src.offset() != null) {
        size -= src.offset();
      }
Why do we need the `else` here? `length` and `offset` are independent. They can have
some value (not null) at the same time or they can be null at the same time. That is,
with the above code, you are not checking and initializing the offset if length is not
null. You do it only if length is null. Shouldn't they be independent of each other as
far as their initializations go?

ashish 10:59 AM
This is the intended behavior. In this code we are just finding the size which we have
to copy. the default is the complete object which we get from stat.length(). If the user
has specified the length then that becomes the size to copy. If the user has not
specified the length but has specified the offset then the code subtracts the offset
from stat.length() to find the actual size. Offset is the location from where to start
and length is the no of bytes to copy.

ersan 11:55 AM
I am repeating the same, but I just want to make sure the requirement is absolutely
clear :
You say our requirement is like "The end-user cannot specify a source offset and
some length at the same time to be able to copy only a subsection of a file when
creating an object by combining different source objects." Is that right?
I cannot think of any good reason why we want to restrict it this way. 
```

sinhaashish:
@ebozduman 

Explaining it with an example. Let say the object size is 100 bytes. `statObject` will give me the size as 100 bytes which is passed and set at as `objectSize` in the source object  using the `buildHeader` method.
```
  ObjectStat stat = statObject(src.bucketName(), src.objectName(), src.sse());
   src.buildHeaders(stat.length(), stat.etag());
```
Just to make things clear redefining the offset and length 
Offset : offset is an integer (including negative numbers) refers an index
Length : Length is a non-negative integer ( which denotes the no of bytes )

 If size is 10, offset = 4 and length = 2 refers two bytes from 4th index 
 offset = -4 and length = 2 refers two bytes from end of 4th index.

 Now the code under discussion  is :
 ```
 long size = stat.length();
      if (src.length() != null) {
        size = src.length();
      } else if (src.offset() != null) {
        size -= src.offset();
      }

```
 
 **Case 1 : _No Offset No length_**
 In this case the complete object will be copied. i.e. complete 100 bytes. We are finding the size from `statObject` method.
  code :
 ```
long size = src.objectSize()
```
 
 **Case 2 : _Length present but No offset_**
 Let say 	`length = 50 ` . In this case the size to copy will be 50 bytes only. 
 code which find the size 
 ```
 if (src.length() != null) {
   size = src.length();
}
``` 
 **Case 3:  _Offset present but no length_** 
Let say  `Offset= 20` . In this case the index from where the copy will start is 20 , but the size will be the total size of object - offset  i.e. (100-20) = 80 bytes.
 Code 
```
else if (src.offset() != null) {
    size -= src.offset();
}
```
 
**Case 4 : _Length and offset both present_**
 Let say offset = 20 and length = 50 . In this case still the size to copy will be 50 bytes only. The only difference is that the bytes to copy will start from 20 and go up til 69  . start index = 20 , end index = start index + length -1 = 20 +50 -1 = 69 . Character at index 20 is included
 x-amz-copy-source-range = "bytes=20-69" .  
 
 If you want to know how it will decide to copy the bytes from 20 to 69 then the below code explains it
 ```
if ((src.offset() != null) && ( src.length() == null)) {
    headerMap.put("x-amz-copy-source-range", "bytes=" + src.offset() + "-" );
}
if ((src.offset() != null) && ( src.length() != null)) {
   headerMap.put("x-amz-copy-source-range", "bytes=" + src.offset() + "-" + (src.offset() + src.length() - 1));
}
```

PR #800, opened 2019-09-28
Docs and examples added for Object Lock

sinhaashish:
Docs updated with below api:
public void makeBucket(String bucketName, String region, boolean objectLock)
public void enableVersioning(String bucketName)
public void disableVersioning(String bucketName)
public void setDefaultRetention(String bucketName, ObjectLockConfiguration config)
public ObjectLockConfiguration getDefaultRetention(String bucketName)

Examples added for make bucket with lock and set/get object lock configuration.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -26,6 +26,10 @@ MinioClient s3Client = new MinioClient("https://s3.amazonaws.com", "YOUR-ACCESSK
 | [`listenBucketNotification`](#listenBucketNotification) |  |   |   |
 | [`setBucketNotification`](#setBucketNotification) |  |   |   |
 | [`getBucketNotification`](#getBucketNotification) |  |   |   |
+| [`enableVersioning`](#enableVersioning) |  |   |   |
+| [`disableVersioning`](#disableVersioning) |  |   |   |
+| [`setDefaultRetention`](#setDefaultRetention) |  |   |   |
+| [`getDefaultRetention`](#getDefaultRetention) |  |   |   |
 
 ## 1. Constructors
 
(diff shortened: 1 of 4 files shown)


docs/API.md
```diff
@@ -951,6 +1008,217 @@ __Example__
...
+|:--- |:--- |:--- |
+| ``bucketName``  | _String_ | Name of the bucket.  |
+| ``config``  | _ObjectLockConfiguration_ | Object lock Configuration  |
+
+
+| Return Type	  | Exceptions	  |
+|:--- |:--- |
+| ``None``  | Listed Exceptions: |
+|        |  ``InvalidBucketNameException`` : upon invalid bucket name. |
+|        |  ``RegionConflictException`` : upon  passed region conflicts with the one previously specified. |
```
kannappanr: Instead of `RegionConflictException` you should talk about `InvalidArgumentException` as that is what is being thrown in the code.

docs/API.md
```diff
@@ -951,6 +1008,217 @@ __Example__
...
+|:--- |:--- |:--- |
+| ``bucketName``  | _String_ | Name of the bucket.  |
+
+
+
+| Return Type	  | Exceptions	  |
+|:--- |:--- |
+| ``None``  | Listed Exceptions: |
+|        |  ``InvalidBucketNameException`` : upon invalid bucket name. |
+|        |  ``RegionConflictException`` : upon  passed region conflicts with the one previously specified. |
```
kannappanr: Instead of RegionConflictException you should talk about InvalidArgumentException as that is what is being thrown in the code.

sinhaashish:
> Please incorporate the changes requested in the comments.

Thanks for pointing this out . Done .

PR #801, opened 2019-10-10
Code cleanup and docs update

sinhaashish:
1. Removed unused exception  `InvalidObjectPrefixException`.
2. API.md updated with `removeAllBucketNotification`


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -4691,7 +4690,6 @@ public void putObject(String bucketName, String objectName, InputStream stream,
    * @throws InternalException           upon internal library error
    * @throws InvalidArgumentException    upon invalid value is passed to a method.
    * @throws InvalidResponseException    upon a non-xml response from server
-   * @see #putObject(String bucketName, String objectName, String fileName)
```
Praveenrajmani: Why do we remove this?

docs/API.md
```diff
@@ -1008,6 +1008,74 @@ __Example__
     }
 ```
 
+  <a name="removeAllBucketNotification"></a>
```
Praveenrajmani: An extra space in the head

api/src/main/java/io/minio/MinioClient.java
sinhaashish: Here we were referring to the overloaded function which is already deprecated.

docs/API.md
sinhaashish: done.

PR #803, opened 2019-10-15
Add configuration "minio.max_part_size" and "minio.max_object_size"

woxinqidai:
Add configuration "minio.max_part_size" and "minio.max_object_size"
example:
System.setProperty("minio.max_part_size", 5L * 1024 * 1024);


ZawadaZSE:
Hey,

I'm also interested in having this kind of change. But I don't think that using system properties is the way to go.
How about providing this configuration via constructor as a global for client instance or method call for single upload?

woxinqidai:
> Hey,
> 
> I'm also interested in having this kind of change. But I don't think that using system properties is the way to go.
> How about providing this configuration via constructor as a global for client instance or method call for single upload?

Yes, I didn't want to make too many changes to the source code, so I used system properties,
When I have time, I change it to a constructor
Thank you for providing such good open source software !

A-Green:
Hello! Is there a workaround to reduce max part size until this change released?

vadmeste:
> Hello! Is there a workaround to reduce max part size until this change released?

In minio-java, the part size already holds the minimum possible value in S3 specificiation, which is 5 MB.

A-Green:
@vadmeste Thank you for quick reply. 
According to the documentation, there is a buffer up to 525MB allocated in case of uploading data from a stream with undefined size:
http://minio.github.io/minio-java/io/minio/MinioClient.html#putObject-java.lang.String-java.lang.String-java.io.InputStream-long-io.minio.ServerSideEncryption-
From the source code I see that part size of 525MB calculated based on MAX_OBJECT_SIZE, MAX_MULTIPART_COUNT and MIN_MULTIPART_SIZE. That is why I put my question here. 
I would like to know if there is a possibility to reduce part size? From the memory consumption perspective: uploading from 10 streams holding 1GB (the exact size undefined in general, but up to 1GB) data each simultaneously leads to 5+ GB RAM allocation

senthillogs:
When we can expect this feature?

harshavardhana:
@woxinqidai I would prefer if this is added as a part_size override, S3 spec disallows changing max_object_size.

It should be optional value for putObject function which allows overriding the automatic part_size combination

senthillogs:
PutObjectOptions as parameter to putObject() will give us more flexibility. It would great if this can be pushed at the earliest.

harshavardhana:
Closed as #843 is merged.

PR #807, opened 2019-11-03
Use Optionals for some frequently used get methods.

twhx99:
Frequently used methods like **statObject**, **getObjectUrl**, **getPresignedObjectUrl** and **getObject** logically throw exceptions if said objects cannot be accessed (for whatever reason). 

I implemented convenience methods for the above mentioned methods. Each convenience method **retrieves the result from the primary one**.

**Apart from the RuntimeException, all other exceptions are ignored and an Optional.empty() is returned.**

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -2067,6 +2067,179 @@ public class MinioClient {
     }
   }
 
+  /**
+   * Gets an Optional containing the entire object's data as {@link InputStream} in given bucket if it exists and is accessible. The
+   * InputStream must be closed after use else the connection will remain open.
+   *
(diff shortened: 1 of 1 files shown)


harshavardhana:
I think this defeats the primitive nature of the library, we should avoid writing application logic in SDKs.

I don't think these APIs are useful it simply bloats out SDK with incorrect requirements.  Can you explain what is the rationale behind this change ?

twhx99:
Basically, I just find it helpful if you don't have to deal with exceptions in all X lines. However, after some consideration I agree with you that such a thing only bloats out the SDK and does not belong in it.

I will probably put such changes in a small extension for the minio-java-sdk, which I will develop in my own repository.

PR #809, opened 2019-11-06
refactor removeObjects

KevinSmile:
### refactor removeObjects (batch remove)

**Issue to solve:**
MinioClient.removeObjects() is kind of strange and of low efficiency.
https://github.com/minio/minio-java/blob/cd1109b16023d43ad63fb7a10bef411094556662/api/src/main/java/io/minio/MinioClient.java#L3004

**Details:**

**1. strange usage of Iterable**
MinioClient.removeObjects() returns an **Iterable**.
This method is doing a DELETE job, but the actual delete job won't execute if you don't iterate over the returned Iterable-value (which will call https://github.com/minio/minio-java/blob/cd1109b16023d43ad63fb7a10bef411094556662/api/src/main/java/io/minio/MinioClient.java#L3014).
 (Call MinioClient.removeObjects() **alone** deletes nothing, which is strange).

 **2. improve generating batchDelete-objectList with java8-stream**
(default batchSize is 1000)

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -17,110 +17,41 @@
 
 package io.minio;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Multimap;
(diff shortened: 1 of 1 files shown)


KevinSmile:
close it as discussed in https://github.com/minio/minio-java/issues/810

Issue #810, opened 2019-11-06
the design of removeObjects


KevinSmile:
as described in PR https://github.com/minio/minio-java/pull/809

MinioClient.removeObjects() is kind of strange and of low efficiency.

MinioClient.removeObjects() returns an Iterable.
This method is doing a DELETE job, but the actual delete job won't execute if you don't iterate over the returned Iterable-value.
Call MinioClient.removeObjects() alone **deletes nothing**, which is strange.

vadmeste:
@KevinSmile thanks for your contribution. It is always better to discuss a new change before implementing it to save your time.

I don't see how this could be low efficiency, iterating should be much faster than calling a web service and delete 1000 elements.

Maybe it is your code which is doing something which takes time between iterations.

KevinSmile:
@vadmeste thanks for for your replay! 
I think you didn't get my point.

(There are 2 kinds of iterator in this issue.
 One is the return value of MinioClient.removeObjects();
the other is about how to generate a batch-toDelete-subList.
batch-toDelete-subList's size should be **less than 1000**, which is the limit set by s3.
I believe batch operation is faster than one-by-one delete since it saves http-calls.
In fact, the original code also used 1000 as batch-size(in a while loop https://github.com/minio/minio-java/blob/cd1109b16023d43ad63fb7a10bef411094556662/api/src/main/java/io/minio/MinioClient.java#L3019)
As mentioned in https://github.com/minio/minio-java/pull/809, I believe using Guava's Iterables.partition and java8's stream to generate a batch-toDelete-subList is better than the original while-loop.

see batch-size limit:
https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html
https://stackoverflow.com/questions/54255990/cheapest-way-to-delete-2-billion-objects-from-s3-ia)

However, I believe the **strange usage of Iterable** is the **key problem** here.

It's a common practice that using iterator to traverse elements of a collection without exposing its underlying representation. I believe iterator should be used as kind of GET method, you should not do something hack (e.g. DELETE) during the iterating, as the original did in the `hasNext` and `next` method:
https://github.com/minio/minio-java/blob/cd1109b16023d43ad63fb7a10bef411094556662/api/src/main/java/io/minio/MinioClient.java#L3041
https://github.com/minio/minio-java/blob/cd1109b16023d43ad63fb7a10bef411094556662/api/src/main/java/io/minio/MinioClient.java#L3067

[...]

PR #811, opened 2019-11-07
[doc] clarify lazy-eval of removeObjects

KevinSmile:
as discussed in https://github.com/minio/minio-java/issues/810

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -2985,7 +2985,8 @@ public class MinioClient {
 
 
   /**
-   * Removes multiple objects from a bucket.
+   * Removes multiple objects from a bucket. As objects removal are lazily executed, its required
+   * to iterate the returned Iterable.
    *
    * </p><b>Example:</b><br>
    * <pre>{@code // Create object list for removal.
@@ -3000,6 +3001,8 @@ public class MinioClient {
    *
    * @param bucketName Bucket name.
    * @param objectNames List of Object names in the bucket.
+   *
+   * @return (lazy) Iterable of the Result DeleteErrors.
    */
   public Iterable<Result<DeleteError>> removeObjects(final String bucketName, final Iterable<String> objectNames) {
     return new Iterable<Result<DeleteError>>() {
--- a/examples/RemoveObjects.java
+++ b/examples/RemoveObjects.java
@@ -48,6 +48,7 @@ public class RemoveObjects {
       objectNames.add("my-objectname3");
 
       // Remove object all objects 'objectNames' list from 'my-bucketname'.
+      // It is required to traverse over the returned Iterable for lazy evaluation.
       for (Result<DeleteError> errorResult: minioClient.removeObjects("my-bucketname", objectNames)) {
         DeleteError error = errorResult.get();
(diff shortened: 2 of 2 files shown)


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2987,19 +2987,27 @@ public void removeObject(String bucketName, String objectName)
   /**
    * Removes multiple objects from a bucket.
    *
+   * <p><b>Note:</b></p>
+   * This method does lazy evaluation. You need to traverse over the returned Iterable to do the
+   * real delete jobs.
```
balamurugana: You would need to modify description than adding new style here. Below is an idea.
```
Removes multiple objects from a bucket. As objects removal are lazily executed, its required to iterate returned Iterable.
```

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2987,19 +2987,27 @@ public void removeObject(String bucketName, String objectName)
...
+   *
    * </p><b>Example:</b><br>
    * <pre>{@code // Create object list for removal.
    * List<String> objectNames = new LinkedList<String>();
    * objectNames.add("my-objectname1");
    * objectNames.add("my-objectname2");
    * objectNames.add("my-objectname3");
+   * // need to traverse over the returned Iterable to do the real delete jobs.
```
balamurugana: I don't think this comment is needed as the javadoc is clear to say about lazy eval.

docs/API.md
```diff
@@ -3079,7 +3079,7 @@ __Parameters__
 
 |Return Type	  | Exceptions	  |
 |:--- |:--- |
-| ``Iterable<Result<DeleteError>>``:an iterator of Result DeleteError.  | _None_  |
+| ``Iterable<Result<DeleteError>>``:a lazy iterator of Result DeleteError.  | _None_  |
```
balamurugana: I don't think this change is needed as the description should carry about lay eval.

docs/API.md
```diff
@@ -3093,6 +3093,7 @@ objectNames.add("my-objectname2");
 objectNames.add("my-objectname3");
 try {
       // Remove object all objects in objectNames list from the bucket my-bucketname.
+      // Lazy evaluation. Need to traverse over the returned Iterable to do the real delete jobs.
```
balamurugana: I don't think this comment is needed as the description carries about lay eval.

examples/RemoveObjects.java
```diff
@@ -48,6 +48,7 @@ public static void main(String[] args)
       objectNames.add("my-objectname3");
 
       // Remove object all objects 'objectNames' list from 'my-bucketname'.
+      // Lazy evaluation. Need to traverse over the returned Iterable to do the real delete jobs.
```
balamurugana: You could add comment like `Its required to traverse over the returned Iterable for lazy evaluation`

api/src/main/java/io/minio/MinioClient.java
KevinSmile: ok

balamurugana:
@KevinSmile you would need to check travis-ci failures and fix them accordingly.

PR #812, opened 2019-11-07
better-batchList-generating in removeObjects

KevinSmile:
better-batchList-generating in removeObjects

as discussed in https://github.com/minio/minio-java/issues/810

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -18,6 +18,7 @@
 package io.minio;
 
 import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Multimaps;
 import com.google.common.io.ByteStreams;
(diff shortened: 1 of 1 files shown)


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3009,24 +3010,27 @@ public void removeObject(String bucketName, String objectName)
           private Result<DeleteError> error;
           private Iterator<DeleteError> errorIterator;
           private boolean completed = false;
-          private Iterator<String> objectNameIter = objectNames.iterator();
+          private final int batchSize = 1000;
```
balamurugana: you could hard-code `1000` i.e. no need of `batchSize` variable.

KevinSmile: ok. done.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3009,24 +3010,26 @@ public void removeObject(String bucketName, String objectName)
           private Result<DeleteError> error;
           private Iterator<DeleteError> errorIterator;
           private boolean completed = false;
-          private Iterator<String> objectNameIter = objectNames.iterator();
+          // divide the objectNames to sublists of the given batchSize
+          private Iterator<List<String>> subObjectListsIter =
+                  Iterables.partition(objectNames, 1000).iterator();
```
balamurugana: you could change the comment and variable name.
```java
// divide given objectNames into list of object names on which each list contains maximum of 1000 names.
private Iterator<List<String>>  batchedObjectNames = Iterables.partition(objectNames, 1000).iterator();
```

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3009,24 +3010,26 @@ public void removeObject(String bucketName, String objectName)
...
-              if (i > 0) {
-                errorList = removeObject(bucketName, objectList);
+              if (subObjectListsIter.hasNext()) {
+                List<DeleteObject> subDeleteObjectList = subObjectListsIter.next().stream().map(objectName -> {
+                  try {
+                    return new DeleteObject(objectName);
+                  } catch (XmlPullParserException e) {
+                    throw new RuntimeException(e);
+                  }
+                }).collect(Collectors.toList());
```
balamurugana: I don't know what advantages we get here. My choice is `for` loop which is quite simple (and to read) than java8 stream here.  Do we get any performance improvement here?

balamurugana:
@KevinSmile You would need to check travis-ci failures and fix them accordingly.

nitisht:
@KevinSmile closing this as stale, feel free to reopen once you have time to address the review comments.

PR #824, opened 2019-12-03
Validate bucket name, object name.

### LINKED ISSUE #819: statObject NPE when object name is empty
Calling statObject("whatever", "") provokes a NullPointerException within ResponseHeader.lastModified() because 'lastModified' field is null in this case.

using java minio-6.0.7

java.lang.NullPointerException
    at io.minio.ResponseHeader.lastModified(ResponseHeader.java:132)                                           
    at io.minio.ObjectStat.<init>(ObjectStat.java:51)
    at io.minio.MinioClient.statObject(MinioClient.java:1482)

It should at least fail the same way as putObject("whatever", "", (...)) does instead of throwing a runtime exception.

BigUstad:
Fixes #819
Also, checks for other possible errors by looking at header values.

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -1542,9 +1542,7 @@ public class MinioClient {
            InternalException, InvalidArgumentException, InvalidResponseException {
     checkReadRequestSse(sse);
     checkBucketName(bucketName);
-    if ( objectName == null ) {
-      throw new InvalidArgumentException("Invalid Stat Object Argument(s). Object name cannot be null.");
-    } else if ( objectName.equals("") ) {
+    if ( objectName == null || objectName.equals("") ) {
       throw new InvalidArgumentException("Invalid Stat Object Argument(s). Object name cannot be empty.");
     }
 
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -1541,6 +1541,12 @@ public class MinioClient {
            InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
            InternalException, InvalidArgumentException, InvalidResponseException {
     checkReadRequestSse(sse);
+    checkBucketName(bucketName);
+    if ( objectName == null ) {
+      throw new InvalidArgumentException("Invalid Stat Object Argument(s). Object name cannot be null.");
+    } else if ( objectName.equals("") ) {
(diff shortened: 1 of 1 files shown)


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1547,8 +1547,24 @@ public ObjectStat statObject(String bucketName, String objectName, ServerSideEnc
...
+      errorMessage += " Object name cannot be empty.";
+    }
+    if ( !errorMessage.equals("") ) {
+      throw new InvalidArgumentException("Invalid Stat Object Argument(s)." + errorMessage);
+    }
     HttpResponse response = executeHead(bucketName, objectName, headers);
     ResponseHeader header = response.header();
+    if ( header.contentLength() == 0
+        && header.contentType() == null
+        && header.statusCode() == null ) {
```
balamurugana: I don't think this check is useful.

According to S3 specification, `Content-Length` must be returned else its server bug. `statusCode` is always available (else you get an exception) and `contentType`can be treated as optional.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1547,8 +1547,24 @@ public ObjectStat statObject(String bucketName, String objectName, ServerSideEnc
       headers = sse.headers();
     }
 
+    checkBucketName(bucketName);
+    String errorMessage = "";
+    if ( objectName == null ) {
+      errorMessage += " Object name cannot be null.";
```
balamurugana: You throw `InvalidArgumentException` here than accumulating messages.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1541,6 +1541,12 @@ public ObjectStat statObject(String bucketName, String objectName, ServerSideEnc
            InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
            InternalException, InvalidArgumentException, InvalidResponseException {
     checkReadRequestSse(sse);
+    checkBucketName(bucketName);
+    if ( objectName == null ) {
+      throw new InvalidArgumentException("Invalid Stat Object Argument(s). Object name cannot be null.");
+    } else if ( objectName.equals("") ) {
```
balamurugana: 1. `else` is not needed here.
2. please fix ci failures.

BigUstad: 1. Else removed.
2. CI failures are not related to this change. We are debugging the root-cause of it.

PR #825, opened 2019-12-06
Set Accept-Encoding to identity by default

BigUstad:
Fixes #820

PR #828, opened 2019-12-16
gzip Accept-Encoding nginx fix.

### LINKED ISSUE #820: StatObject issue in Java Driver
This is a relatively low priority issue, I just stumbled on it by accident.

If MinIO is behind a HTTP proxy, like nginx, and you have gzip compression on ( I know you shouldn't ), stat always gives you a zero length/size for the object. ( Because "Content-Length" may be missing )

I wouldn't bother reporting it, bit oddly it works in the Go and Python driver, and mc command line,  just fine. *shrugs*

BigUstad:
Fixes #820
We need this change to get around the okhttp feature of using gzip by default. As this post by okhttp developers point it out, we need to set the Accept-encoding http header to identity to get around the problem.

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -957,6 +957,7 @@ public class MinioClient {
           len = 0;
         }
 
+        requestBuilder.header("Accept-Encoding", "identity");
         if (method == Method.POST && queryParamMap != null && queryParamMap.containsKey("delete")) {
           // Fix issue #579: Treat 'Delete Multiple Objects' specially which requires MD5 hash.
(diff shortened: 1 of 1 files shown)


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -957,6 +957,7 @@ private Request createRequest(Method method, String bucketName, String objectNam
           len = 0;
         }
 
+        requestBuilder.header("Accept-Encoding", "identity");
```
balamurugana: Please add a comment `disable default gzip compression`

kannappanr:
Closing this PR as we already have #825 open

PR #834, opened 2020-01-06
Add overloaded Listen Bucket Notification method

sinhaashish:
There was an infine while loop in `listenBucketNotification` which kept on reading for any event on the  server. 
With this PR, For the functional  test `listenBucketNotification_test1` once the event is received the infinite loop is stopped by calling the method `client.turnOffListening()` .

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -17,6 +17,8 @@
 
 package io.minio;
 
+import com.fasterxml.jackson.core.JsonParseException;
+import com.fasterxml.jackson.databind.JsonMappingException;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Multimaps;
@@ -5716,7 +5718,6 @@ public class MinioClient {
    * @throws InvalidResponseException    upon a non-xml response from server
    *
    */
-
   public void listenBucketNotification(String bucketName, String prefix, String suffix, String[] events,
       BucketEventListener eventCallback)
     throws InvalidBucketNameException, NoSuchAlgorithmException, InsufficientDataException, IOException,
@@ -5760,6 +5761,126 @@ public class MinioClient {
         scanner.close();
       }
     }
+  }
+
+  /**
+   * Listen to bucket notifications. As bucket notification are lazily executed,
+   * its required to iterate. The returned closeable iterator must be used with
+   * try with resource or else the stream will not be closed.
+   *
+   * </p><b>Example:</b><br>
+   * <pre>{@code
+   *  try (CloseableIterator<Result<NotificationInfo>> ci = client
+   * .listenBucketNotification("my-bucket", "", "", events)) {
+   *  while (ci.hasNext()) {
+   *    NotificationInfo info = ci.next().get();
+   *    System.out.println(info.toString());
+   *   }
+   *  } catch (IOException e) {
+   *    System.out.println("Error occurred: " + e);
+   *  }
+   * }
+   * </pre>
+   *
+   * @param bucketName Bucket name.
+   * @param prefix Prefix of concerned objects events.
+   * @param suffix Suffix of concerned objects events.
+   * @param events List of events to watch.
+   *
+   * @return (lazy) CloseableIterator of the Result NotificationInfo.
+   *
+   */
+  public CloseableIterator<Result<NotificationInfo>> listenBucketNotification(String bucketName, String prefix,
+      String suffix, String[] events)
+    throws IOException, InvalidKeyException, NoSuchAlgorithmException, InsufficientDataException,
+    InvalidResponseException, InternalException, NoResponseException, InvalidBucketNameException,
+    XmlPullParserException, ErrorResponseException {
+
+    Multimap<String, String> queryParamMap = HashMultimap.create();
+    queryParamMap.put("prefix", prefix);
+    queryParamMap.put("suffix", suffix);
+    for (String event : events) {
+      queryParamMap.put("events", event);
+    }
+
+    HttpResponse response = executeReq(Method.GET, getRegion(bucketName),
+        bucketName, "", null, queryParamMap, null, 0);
+
+    return new CloseableIterator<Result<NotificationInfo>>() {
+      Scanner scanner  = new Scanner(response.body().charStream()).useDelimiter("\n");
+
+      String notificationString = null;
+      ObjectMapper mapper = new ObjectMapper();
+      NotificationInfo notificationInfo = null;
+      boolean isClosed = false;
+
+      @Override
+      public void close() throws IOException {
+        if (!isClosed) {
+          try {
+            response.body().close();
+            scanner.close();
+          } finally {
+            isClosed = true;
+          }
+        }
+      }
+
+      public boolean populate()  {
+        if (isClosed) {
+          return false;
+        }
+
+        if (notificationString != null) {
+          return true;
+        }
+
+        while (scanner.hasNext()) {
+          notificationString = scanner.next().trim();
+          if ( !notificationString.equals("")) {
+            break;
+          }
+        }
+
+        if (notificationString  == null || notificationString.equals("")) {
+          try {
+            close();
+          } catch (IOException e) {
+            isClosed = true;
+          }
+          return false;
+        }
+        return true;
+      }
+
+      @Override
+      public boolean hasNext() {
+        return populate();
+      }
+
+      @Override
+      public Result<NotificationInfo> next() {
+        if (isClosed) {
+          throw new NoSuchElementException();
+        }
+        if ((notificationString  == null || notificationString.equals("")) &&  !populate() ) {
+              throw  new NoSuchElementException();
+        }
+
+        try {
+              notificationInfo = mapper.readValue(notificationString, NotificationInfo.class);
+              return new Result<>(notificationInfo, null);
+            } catch (JsonParseException e) {
+              return new Result<>(null, e);
+            } catch ( JsonMappingException e) {
+              return new Result<>(null, e);
+            } catch ( IOException e) {
+              return new Result<>(null, e);
+        } finally {
+          notificationInfo = null;
+        }
+      }
+    };
   }
 
 
--- a/api/src/main/java/io/minio/CloseableIterator.java
+++ b/api/src/main/java/io/minio/CloseableIterator.java
@@ -1,0 +1,25 @@
+/*
+ * MinIO Java SDK for Amazon S3 Compatible Cloud Storage,
+ * (C) 2020 MinIO, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.minio;
+
+import java.io.Closeable;
+import java.util.Iterator;
+
+public interface CloseableIterator<T> extends Iterator<T>, Closeable {
+}
+
--- a/api/src/main/java/io/minio/Result.java
+++ b/api/src/main/java/io/minio/Result.java
@@ -16,6 +16,8 @@
 
 package io.minio;
 
+import com.fasterxml.jackson.core.JsonParseException;
+import com.fasterxml.jackson.databind.JsonMappingException;
 import java.io.IOException;
 import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
@@ -47,7 +49,8 @@ public class Result<T> {
    * Returns given Type if exception is null, else respective exception is thrown.
    */
   public T get()
-    throws InvalidBucketNameException, NoSuchAlgorithmException, InsufficientDataException, IOException,
+    throws InvalidBucketNameException, NoSuchAlgorithmException, InsufficientDataException,
+           JsonParseException, JsonMappingException,IOException,
            InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
            InternalException {
     if (ex == null) {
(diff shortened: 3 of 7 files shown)


harshavardhana:
I think this is quite unusual style for Java and also racy, the proper way to do is to be able to write a canceller. Just like Go style doneCh we should communicate over safe values.

There are quite common ways to write cancellers in Java, I think you should perhaps try that approach @sinhaashish

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -5741,12 +5743,16 @@ public void listenBucketNotification(String bucketName, String prefix, String su
       scanner = new Scanner(response.body().charStream());
       scanner.useDelimiter("\n");
       while (scanner.hasNext()) {
```
balamurugana: My suggestion is to deprecate this style of API and introduce lazy eval'ed style like
```java
public Iterable<Result<NotificationInfo>> listenBucketNotification(String bucketName, String prefix, String suffix, String[] events)
```

If we stick with current style, the fix is just `while (!continueListening || scanner.hasNext())`

harshavardhana: A more idiomatic Java style is preferred, so I think lazy evaled style is better and bump the version for breakage

sinhaashish: @balamurugana  need your input on this . I have implemented the `listenBucketNotification` as below but it waits at `scanner.next()` .Please review the code.
```
public Iterable<Result<NotificationInfo>> listenBucketNotification1(String bucketName, String prefix,
      String suffix, String[] events) {
    return new Iterable<Result<NotificationInfo>>() {

      @Override
      public Iterator<Result<NotificationInfo>> iterator() {
        return new Iterator<Result<NotificationInfo>>() {

          private Result<NotificationInfo> error;
          NotificationInfo ni;
          private boolean completed = false;

          private synchronized void populate() {
            Multimap<String,String> queryParamMap = HashMultimap.create();
            queryParamMap.put("prefix", prefix);
            queryParamMap.put("suffix", suffix);
            for (String event: events) {
              queryParamMap.put("events", event);
            }

            String bodyContent = "";
            Scanner scanner = null;
            HttpResponse response = null;
            ObjectMapper mapper = new ObjectMapper();

            try {
              response = executeReq(Method.GET, getRegion(bucketName),
                bucketName, "", null, queryParamMap, null, 0);
              scanner = new Scanner(response.body().charStream());
              scanner.useDelimiter("\n");
              if (scanner.hasNext()) {
                bodyContent = scanner.next().trim();
                if (!bodyContent.equals("")) {
                  ni = mapper.readValue(bodyContent, NotificationInfo.class);
                }
              }
            } catch (InvalidBucketNameException | NoSuchAlgorithmException | InsufficientDataException | IOException
              | InvalidKeyException | NoResponseException | XmlPullParserException | ErrorResponseException
              | InternalException | InvalidResponseException | RuntimeException e) {
              this.error = new Result<>(null, e);
            }             finally {
              if (response != null) {
                response.body().close();
              }
              if (scanner != null) {
                scanner.close();
              }
            }

          }


          @Override
          public boolean hasNext() {
            if (this.completed) {
              return false;
            }

            if (this.error == null ) {
              populate();
            }

            if (this.ni == null ) {
              populate();
            }

            this.completed = true;
            return false;
          }

          @Override
          public Result<NotificationInfo> next() {
            if (this.completed) {
              throw new NoSuchElementException();
            }
            if (this.error == null ) {
              populate();
            }
            this.completed = true;
            throw new NoSuchElementException();
          }

          @Override
          public void remove() {
            throw new UnsupportedOperationException();
          }
        };
      }
    };
  }
```

harshavardhana: This looks neat :+1: @sinhaashish unless @balamurugana has any concerns

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -5692,77 +5693,110 @@ public void removeIncompleteUpload(String bucketName, String objectName)
...
   /**
-   * Listen to bucket notifications.
+   * Listen to bucket notifications. As bucket notification are lazily executed, its required
+   * to iterate the returned Iterable.
+   *
+   * </p><b>Example:</b><br>
+
+   * <pre>{@code
+   * Iterator<Result<NotificationInfo>> it =
+   *  client.listenBucketNotification("sinha1", "", "", events).iterator();
```
kannappanr:
```suggestion
   *  client.listenBucketNotification("bucketname", "", "", events).iterator();
```

build.gradle
```diff
@@ -35,7 +35,7 @@ apply plugin: 'io.codearte.nexus-staging'
 
 allprojects {
     group = 'io.minio'
-    version = '6.0.12'
+    version = '7.0.0'
```
balamurugana: major version change should remove all deprecated methods

sinhaashish: @balamurugana  ptal

build.gradle
```diff
@@ -35,7 +35,7 @@ apply plugin: 'io.codearte.nexus-staging'
 
 allprojects {
     group = 'io.minio'
-    version = '6.0.12'
+    version = '6.0.13'
```
kannappanr: This is not required if you are not having a breaking change

kannappanr:
```suggestion
    version = '6.0.12'
```

sinhaashish: As per discussion with @balamurugana  we will have two listenBucketNotification method. Old one will be depricated. After some time we can have a major release in which all the deprecated methods will be removed .

harshavardhana:
PTAL @balamurugana

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -5717,6 +5719,7 @@ public void removeIncompleteUpload(String bucketName, String objectName)
    *
    */
 
+  @Deprecated
```
balamurugana: do not make changes in deprecated methods instead you remove them in separate PR

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -5763,6 +5770,56 @@ public void listenBucketNotification(String bucketName, String prefix, String su
...
+   * <pre>{@code
+   * String[] events = {"s3:ObjectCreated:*", "s3:ObjectAccessed:*"};
+   *   try (CloseableIterator<Result<NotificationInfo>> ci = client
+   *     .listenBucketNotification("bucket-name", "", "", events).getMyCloseableIterator()) {
+   *      while (ci.hasNext()) {
+   *        NotificationInfo info = ci.next().get();
+   *        if ( info != null) {
+   *
+   *      System.out.println(info.records[0].s3.bucket.name + "/"
+   *          + info.records[0].s3.object.key + " has been created");
```
balamurugana: this is wrong message. you would need to print the event

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -5763,6 +5770,56 @@ public void listenBucketNotification(String bucketName, String prefix, String su
...
+    InvalidResponseException, InternalException, NoResponseException, InvalidBucketNameException,
+    XmlPullParserException, ErrorResponseException {
+
+    Multimap<String, String> queryParamMap = HashMultimap.create();
+    queryParamMap.put("prefix", prefix);
+    queryParamMap.put("suffix", suffix);
+    for (String event : events) {
+      queryParamMap.put("events", event);
+    }
+    HttpResponse response = executeReq(Method.GET, getRegion(bucketName),
```
balamurugana: you could use `executeGet()`

api/src/main/java/io/minio/NotificationListener.java
```diff
@@ -0,0 +1,100 @@
...
+
+  NotificationListener( Reader reader) {
+    this.inputReader = reader;
+
+    this.sc = new Scanner(inputReader);
+    this.sc.useDelimiter("\n");
+  }
+  /**
+   * Return a closeable iterator.
+   */
```
balamurugana: cleanup all spacing and stylng

api/src/main/java/io/minio/NotificationListener.java
```diff
@@ -0,0 +1,100 @@
...
+
+  public  CloseableIterator<Result<NotificationInfo>> getMyCloseableIterator() {
```
balamurugana: this is incorrect. this class supposes to have implemented `ClosableIterable`

api/src/main/java/io/minio/NotificationListener.java
```diff
@@ -0,0 +1,100 @@
...
+      @Override
+      public boolean hasNext() {
+        if (isClosed) {
+          return false;
+        }
+        if (!NotificationListener.this.sc.hasNext()) {
+          try {
+            close();
+          } catch (IOException e) {
+            e.printStackTrace();
```
balamurugana: its not good to print exceptions

api/src/main/java/io/minio/NotificationListener.java
```diff
@@ -0,0 +1,100 @@
...
+
+  public  CloseableIterator<Result<NotificationInfo>> getMyCloseableIterator() {
+    return new CloseableIterator<Result<NotificationInfo>>() {
+
+      @Override
+      public boolean hasNext() {
+        if (isClosed) {
+          return false;
+        }
+        if (!NotificationListener.this.sc.hasNext()) {
```
balamurugana: you would need to also check for empty events and ignore them

api/src/main/java/io/minio/NotificationListener.java
```diff
@@ -0,0 +1,100 @@
...
+          }
+          return false;
+        }
+        return true;
+      }
+
+      @Override
+      public Result<NotificationInfo> next() {
+        try {
+          while (sc.hasNext()) {
```
balamurugana: this `while` should be in `hasNext()`

api/src/main/java/io/minio/NotificationListener.java
```diff
@@ -0,0 +1,100 @@
...
+        if (!isClosed) {
+          if (NotificationListener.this.inputReader != null) {
+            inputReader.close();
+          }
+
+          if (NotificationListener.this.sc != null) {
+            sc.close();
+          }
+          isClosed = true;
+        }
```
balamurugana: below logic is crisp and clear
```java
if (!isClosed) {
    try {
        reader.close();
        scanner.close();
    } finally {
        reader = null;
        scanner = null;
        isClosed = true;
    }
}
```

api/src/main/java/io/minio/NotificationListener.java
```diff
@@ -0,0 +1,100 @@
...
+ * NotificationListener is used for listening notification.
+ */
+public class NotificationListener  {
+
+  Scanner sc = null;
+  String notificationString = "";
+  NotificationInfo ni = null;
+  Reader inputReader;
+  boolean isClosed = false;
+  ObjectMapper mapper = new ObjectMapper();
```
balamurugana: 1. use `reader`, `scanner`, `notificationInfo` instead of `inputReader`, `sc`, `ni`
2. group relevant fields together `reader`, `scanner` and `mapper`, `notificationInfo`, then `isClosed`

api/src/main/java/io/minio/MinioClient.java
sinhaashish: we cant use `executeGet` as it uses a HashMap for `queryParameter`. While using a hashmap the values are overridden in the below code

api/src/main/java/io/minio/NotificationListener.java
sinhaashish: When i tried to have the while in hasNext() as shown in the code below 
```
@Override
      public boolean hasNext()  {
        if (isClosed) {
          return false;
        }
        if (!scanner.hasNext()) {
          try {
            close();
          } catch (IOException e) {
            exception = e;
          }
          return false;
        }

        while (scanner.hasNext()) {
          notificationString = scanner.next().trim();
          if (notificationString.equals("")) {
            continue;
          } else {
            break;
          }

        }
        return true;
      }

      @Override
      public Result<NotificationInfo> next() {
        try {
          notificationInfo = mapper.readValue(notificationString, NotificationInfo.class);
        } catch (JsonParseException e) {
          exception = e;
        } catch (  JsonMappingException mappingEx) {
          exception = mappingEx;
        } catch ( IOException ioexception) {
          exception = ioexception;
        }
        if ( exception != null ) {
          return new Result<>(null, exception);
        }
        return new Result<>(notificationInfo, null);
      }
```
Got a build error as shown 
```
io.minio.MinioClient$8.next() can't throw NoSuchElementException
 ```

Thus while is not a part of `hasNext`

**Comment by sinhaashish** [comment]
@balamurugana  PTAL

File: api/src/main/java/io/minio/MinioClient.java [hunk 15]
```diff
@@ -5762,6 +5763,131 @@ public void listenBucketNotification(String bucketName, String prefix, String su
...
+   * </p><b>Example:</b><br>
+   * <pre>{@code
+   *  try (CloseableIterator<Result<NotificationInfo>> ci = client
+   * .listenBucketNotification("my-bucket", "", "", events)) {
+   *  while (ci.hasNext()) {
+   *    NotificationInfo info = ci.next().get();
+   *    System.out.println(info.toString());
+   *    if (info != null) {
+   *      break;
+   *      }
```
balamurugana: you could remove this `if` block i.e. let it run indefinitely in the example.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -5762,6 +5763,131 @@ public void listenBucketNotification(String bucketName, String prefix, String su
...
+        try {
+              notificationInfo = mapper.readValue(notificationString, NotificationInfo.class);
+              return new Result<>(notificationInfo, null);
+            } catch ( IOException ex ) {
+              if ( ex instanceof JsonParseException) {
+                return new Result<>(null, (JsonParseException)ex);
+              } else if ( ex instanceof  JsonMappingException) {
+                return new Result<>(null, (JsonMappingException)ex);
+              } else {
+                return new Result<>(null, ex);
```
balamurugana: below code is correct.

```java
} catch (JsonParseException | JsonMappingException | IOException e) {
    return new Result(null, e);
}
```
In `Result.java`, you need to have `get()` like below (Note: I modified `if` block better)
```java
  public T get()
    throws InvalidBucketNameException, NoSuchAlgorithmException, InsufficientDataException, IOException,
           JsonParseException, JsonMappingException, InternalException {
    if (ex == null) {
      return type;
    }

    if (ex instanceof InvalidBucketNameException) {
      throw (InvalidBucketNameException) ex;
    }

    if (ex instanceof NoSuchAlgorithmException) {
      throw (NoSuchAlgorithmException) ex;
    }

    if (ex instanceof InsufficientDataException) {
      throw (InsufficientDataException) ex;
    }

    if (ex instanceof IOException) {
      throw (IOException) ex;
    }

    if (ex instanceof InvalidKeyException) {
      throw (InvalidKeyException) ex;
    }

    if (ex instanceof NoResponseException) {
      throw (NoResponseException) ex;
    }

    if (ex instanceof XmlPullParserException) {
      throw (XmlPullParserException) ex;
    }

    if (ex instanceof ErrorResponseException) {
      throw (ErrorResponseException) ex;
    }

    if (ex instanceof JsonParseException) {
      throw (JsonParseException) ex;
    }

    if (ex instanceof JsonMappingException) {
      throw (JsonMappingException) ex;
    }

    throw (InternalException) ex;
  }
}
```

sinhaashish: Since `JsonParseException ` and `JsonMappingException` are sub classes of `IOException` , the pipe in catch would not work. So i have caught the child class exception first and then the parent class.
Refer https://stackoverflow.com/questions/11116853/declare-method-to-throw-an-exception-and-subclass-of-this-exception

```
            } catch (JsonParseException e) {
              return new Result<>(null, e);
            } catch ( JsonMappingException e) {
              return new Result<>(null, e);
            } catch ( IOException e) {
              return new Result<>(null, e);
```

balamurugana: :+1: You would need to do the same in `Result.get()`

kannappanr: @sinhaashish Is the change requested in `Result.get()` done?

sinhaashish: @balamurugana  this code is already added in Result.java.
```

    if (ex instanceof JsonParseException) {
      throw (JsonParseException) ex;
    }

    if (ex instanceof JsonMappingException) {
      throw (JsonMappingException) ex;
    }

    if (ex instanceof IOException) {
      throw (IOException) ex;
    }
```

PR #836, opened 2020-01-27
Add new method to set/get object retention & Legal Hold

sinhaashish:
Add new method to set/get object retention and Legal Hold

Function test not added as it will make the object WORM enabled and cant be deleted after successful run of test.

### DIFF
--- a/api/src/main/java/io/minio/DateFormat.java
+++ b/api/src/main/java/io/minio/DateFormat.java
@@ -40,5 +40,8 @@ public class DateFormat {
   public static final DateTimeFormatter HTTP_HEADER_DATE_FORMAT =
       DateTimeFormat.forPattern("EEE',' dd MMM yyyy HH':'mm':'ss 'GMT'").withZoneUTC().withLocale(Locale.US);
 
+  public static final DateTimeFormatter RETENTION_DATE_FORMAT = EXPIRATION_DATE_FORMAT;
+     // DateTimeFormat.forPattern("yyyy-MM-dd'T'HH':'mm':'ss'Z'").withZoneUTC().withLocale(Locale.US);    
+
   private DateFormat() {}
 }
--- a/api/src/main/java/io/minio/DateFormat.java
+++ b/api/src/main/java/io/minio/DateFormat.java
@@ -40,8 +40,5 @@ public class DateFormat {
   public static final DateTimeFormatter HTTP_HEADER_DATE_FORMAT =
       DateTimeFormat.forPattern("EEE',' dd MMM yyyy HH':'mm':'ss 'GMT'").withZoneUTC().withLocale(Locale.US);
 
-  public static final DateTimeFormatter RETENTION_DATE_FORMAT = EXPIRATION_DATE_FORMAT;
-     // DateTimeFormat.forPattern("yyyy-MM-dd'T'HH':'mm':'ss'Z'").withZoneUTC().withLocale(Locale.US);    
-
   private DateFormat() {}
 }
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -3906,14 +3906,14 @@ public class MinioClient {
    * Applies object retention lock onto an object.
    *
    * </p><b>Example:</b><br>
-   * <pre>{@code minioClient.setObjectRetention("my-bucketname", "my-object", config );
+   * <pre>{@code minioClient.setObjectRetention("my-bucketname", "my-object", config, true );
    * System.out.println("Set object retention on my-object successfully."); }</pre>
    *
    * @param bucketName Bucket name.
    * @param objectName Object name.
    * @param versionId  Object versio id.
-   * @param bypassGovernanceRetention  By pass governance retention.
    * @param config     Object lock configuration.
+   * @param bypassGovernanceRetention  By pass governance retention.
    *
    * @throws InvalidBucketNameException  upon invalid bucket name is given
    * @throws NoSuchAlgorithmException
@@ -3928,17 +3928,24 @@ public class MinioClient {
    * @throws ErrorResponseException      upon unsuccessful execution
    * @throws InternalException           upon internal library error
    * @throws InvalidResponseException    upon a non-xml response from server
+   * @throws InvalidArgumentException    upon invalid value is passed to a method.
    */
   public void setObjectRetention(String bucketName, String objectName, String versionId, 
-          boolean bypassGovernanceRetention, ObjectRetentionConfiguration config)
+          ObjectRetentionConfiguration config, boolean bypassGovernanceRetention)
     throws InvalidBucketNameException, NoSuchAlgorithmException, InsufficientDataException, IOException,
            InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
-           InternalException, InvalidResponseException {
+           InternalException, InvalidResponseException, InvalidArgumentException {
+
+    if (config == null ) {
+      throw new InvalidArgumentException("null value is not allowed in config.");
+    }
 
     Map<String, String> queryParamMap = new HashMap<>();
     queryParamMap.put("retention", "");
 
-    if (!versionId.isEmpty()) {
+    if (versionId == null) {
+      queryParamMap.put("versionId", "");
+    } else {
       queryParamMap.put("versionId", versionId);
     }
 
@@ -4000,21 +4007,17 @@ public class MinioClient {
     }
     return result;
   }
-
-
- 
   
-   /**
-   * Applies object legal hold on an object.
+  /**
+   * Enables object legal hold on an object.
    *
    * </p><b>Example:</b><br>
-   * <pre>{@code minioClient.setObjectLegalHold("my-bucketname", "my-object", "", true );
-   * System.out.println("Legal Hold set on my-object successfully."); }</pre>
+   * <pre>{@code minioClient.enableObjectLegalHold("my-bucketname", "my-object", "");
+   * System.out.println("Legal Hold enabled on my-object successfully."); }</pre>
    *
    * @param bucketName Bucket name.
    * @param objectName Object name.
-   * @param versionId  Object versio id.   * 
-   * @param legalHold  Legal Hold Status
+   * @param versionId  Object version id.
    *
    * @throws InvalidBucketNameException  upon invalid bucket name is given
    * @throws NoSuchAlgorithmException
@@ -4030,8 +4033,7 @@ public class MinioClient {
    * @throws InternalException           upon internal library error
    * @throws InvalidResponseException    upon a non-xml response from server
    */
-  public void setObjectLegalHold(String bucketName, String objectName, String versionId, 
-          boolean legalHold)
+  public void enableObjectLegalHold(String bucketName, String objectName, String versionId)
     throws InvalidBucketNameException, NoSuchAlgorithmException, InsufficientDataException, IOException,
            InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
            InternalException, InvalidResponseException {
@@ -4043,19 +4045,63 @@ public class MinioClient {
       queryParamMap.put("versionId", versionId);
     }
 
-    ObjectLockLegalHold objectLockLegalHold = new ObjectLockLegalHold(legalHold);
+    ObjectLockLegalHold objectLockLegalHold = new ObjectLockLegalHold(true);
 
     HttpResponse response = executePut(bucketName, objectName, null, queryParamMap, objectLockLegalHold, 0);
     response.body().close();
   }
 
   /**
-   * Fetches Legal hold status of an object.
+   * Disable object legal hold on an object.
+   *
+   * </p><b>Example:</b><br>
+   * <pre>{@code minioClient.disableObjectLegalHold("my-bucketname", "my-object", "");
+   * System.out.println("Legal Hold disabled on my-object successfully."); }</pre>
+   *
+   * @param bucketName Bucket name.
+   * @param objectName Object name.
+   * @param versionId  Object version id.
+   *
+   * @throws InvalidBucketNameException  upon invalid bucket name is given
+   * @throws NoSuchAlgorithmException
+   *           upon requested algorithm was not found during signature calculation
+   * @throws InsufficientDataException  upon getting EOFException while reading given
+   *           InputStream even before reading given length
+   * @throws IOException                 upon connection error
+   * @throws InvalidKeyException
+   *           upon an invalid access key or secret key
+   * @throws NoResponseException         upon no response from server
+   * @throws XmlPullParserException      upon parsing response xml
+   * @throws ErrorResponseException      upon unsuccessful execution
+   * @throws InternalException           upon internal library error
+   * @throws InvalidResponseException    upon a non-xml response from server
+   */
+  public void disableObjectLegalHold(String bucketName, String objectName, String versionId)
+    throws InvalidBucketNameException, NoSuchAlgorithmException, InsufficientDataException, IOException,
+           InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
+           InternalException, InvalidResponseException {
+
+    Map<String, String> queryParamMap = new HashMap<>();
+    queryParamMap.put("legal-hold", "");
+
+    if (!versionId.isEmpty()) {
+      queryParamMap.put("versionId", versionId);
+    }
+
+    ObjectLockLegalHold objectLockLegalHold = new ObjectLockLegalHold(false);
+
+    HttpResponse response = executePut(bucketName, objectName, null, queryParamMap, objectLockLegalHold, 0);
+    response.body().close();
+  }
+
+  /**
+   * Returns true is the object legal hold is enabled.
    *
    * </p><b>Example:</b><br>
    * <pre>{@code 
-   * ObjectLockLegalHold objectLockLegalHold = minioClient.getObjectLegalHold("my-bucketname", "my-object", "" );
-   * System.out.println("Legal Hold Status " + objectLockLegalHold.getStatus()); }</pre>
+   * Boolean isObjectLegalHoldEnabled = minioClient.isObjectLegalHoldEnabled("my-bucketname", 
+   *  "my-object", "" );
+   * System.out.println("Is Object Legal Hold enabled " + isObjectLegalHoldEnabled); }</pre>
    *
    * @param bucketName Bucket name.
    * @param objectName Object name.
@@ -4075,7 +4121,7 @@ public class MinioClient {
    * @throws InternalException           upon internal library error
    * @throws InvalidResponseException    upon a non-xml response from server
    */
-  public ObjectLockLegalHold getObjectLegalHold(String bucketName, String objectName, String versionId)
+  public Boolean isObjectLegalHoldEnabled(String bucketName, String objectName, String versionId)
     throws InvalidBucketNameException, NoSuchAlgorithmException, InsufficientDataException, IOException,
             InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
             InternalException, InvalidResponseException {
@@ -4094,7 +4140,7 @@ public class MinioClient {
     } finally {
       response.body().close();
     }
-    return result;
+    return result.getStatus();
   }
 
   /**
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -59,6 +59,7 @@ import io.minio.messages.ListBucketResultV1;
 import io.minio.messages.ListMultipartUploadsResult;
 import io.minio.messages.ListPartsResult;
 import io.minio.messages.ObjectLockConfiguration;
+import io.minio.messages.ObjectLockLegalHold;
 import io.minio.messages.OutputSerialization;
 import io.minio.messages.Part;
 import io.minio.messages.Prefix;
@@ -67,6 +68,7 @@ import io.minio.messages.NotificationConfiguration;
 import io.minio.messages.SelectObjectContentRequest;
 import io.minio.org.apache.commons.validator.routines.InetAddressValidator;
 
+import io.minio.messages.ObjectRetentionConfiguration;
 import io.minio.notification.NotificationInfo;
 
 import okhttp3.HttpUrl;
@@ -976,10 +978,15 @@ public class MinioClient {
           String[] hashes = Digest.sha256Md5Hashes(data, len);
           sha256Hash = hashes[0];
           md5Hash = hashes[1];
-        } else if (method == Method.PUT && queryParamMap != null && queryParamMap.containsKey("lifecycle")) {
+        } else if (method == Method.PUT && queryParamMap != null &&  queryParamMap.containsKey("lifecycle"))  {
           String[] hashes = Digest.sha256Md5Hashes(data, len);
           sha256Hash = hashes[0];
           md5Hash = hashes[1];
+        } else if (method == Method.PUT && headerMap.containsKey("retention")) {
+          String[] hashes = Digest.sha256Md5Hashes(data, len);
+          sha256Hash = hashes[0];
+          md5Hash = hashes[1];
+          headerMap.remove("retention","");
         } else {
           // Fix issue #567: Compute SHA256 hash only.
           sha256Hash = Digest.sha256Hash(data, len);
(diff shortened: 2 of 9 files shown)


api/src/main/java/io/minio/messages/ObjectLockRetentionConfiguration.java
```diff
@@ -0,0 +1,84 @@
...
+import io.minio.DateFormat;
+import org.joda.time.DateTime;
+import com.google.api.client.util.Key;
+
+
+/**
+ * Helper class to construct create bucket configuration request XML for Amazon AWS S3.
+ */
+@edu.umd.cs.findbugs.annotations.SuppressFBWarnings(value = "URF_UNREAD_FIELD")
+public class ObjectLockRetentionConfiguration extends XmlEntity {
```
balamurugana: You could name it similar to S3 API `ObjectRetentionConfiguration`

api/src/main/java/io/minio/messages/ObjectLockRetentionConfiguration.java
```diff
@@ -0,0 +1,84 @@
...
+  @Key("ObjectRetention")
+  private ObjectLockRetention objectLockRetention;
+  @Key("BypassGovernanceRetention")
+  private boolean bypassGovernanceRetention;
+  @Key("Version")
+  private String versionId;
```
balamurugana: `versionId` is not part of the configuration. It is a param in the API.

api/src/main/java/io/minio/messages/ObjectLockRetentionConfiguration.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Helper class to construct create bucket configuration request XML for Amazon AWS S3.
+ */
+@edu.umd.cs.findbugs.annotations.SuppressFBWarnings(value = "URF_UNREAD_FIELD")
+public class ObjectLockRetentionConfiguration extends XmlEntity {
+  @Key("ObjectRetention")
+  private ObjectLockRetention objectLockRetention;
+  @Key("BypassGovernanceRetention")
+  private boolean bypassGovernanceRetention;
```
balamurugana: This is not part of the config. Its a param to the S3 API.

api/src/main/java/io/minio/messages/ObjectLockRetentionConfiguration.java
```diff
@@ -0,0 +1,84 @@
...
+  public String version() { 
+    return this.versionId;
+  }
+
+  /**
+   * Returns boolean for governanceBypass.
+   */
+  public boolean governanceBypass() { 
+    return this.bypassGovernanceRetention;
+  }
```
balamurugana: Above two getters are not required

api/src/main/java/io/minio/messages/ObjectLockRetention.java
```diff
@@ -0,0 +1,79 @@
...
+   */
+  public ObjectLockRetention() throws XmlPullParserException {
+    super();
+    super.name = "Retention";
+  }
+
+  /**
+   * Constructs a new CustomRetention object with given retention.
+   */
+  public ObjectLockRetention(RetentionMode mode,  String retainUntilDate) throws XmlPullParserException {
```
balamurugana: `retainUntilDate` should be a date/time object not string.

api/src/main/java/io/minio/messages/ObjectLockRetention.java
```diff
@@ -0,0 +1,79 @@
...
+   */
+  public DateTime retainUntil() {
+    return DateFormat.RETENTION_DATE_FORMAT.parseDateTime(retainUntilDate);
+  }
+
+  /**
+   * Returns retain until date.
+   */
+  public Date getRetentionDate() {
+    return DateFormat.RETENTION_DATE_FORMAT.parseDateTime(retainUntilDate).toDate();
```
balamurugana: above two methods are same getter. You would need to remove `getRetentionDate()`

api/src/main/java/io/minio/messages/ObjectLockRetention.java
```diff
@@ -0,0 +1,79 @@
...
+   */
+  public RetentionMode mode() {
+    return RetentionMode.fromString(mode);
+  }
+
+  /**
+   * Returns retain until date.
+   */
+  public DateTime retainUntil() {
+    return DateFormat.RETENTION_DATE_FORMAT.parseDateTime(retainUntilDate);
```
balamurugana: I guess if `retainUntilDate` is `null`,  `parseDateTime()` may throw an exception which is not needed here. Please check and return `null` or `DateTime`

api/src/main/java/io/minio/DateFormat.java
```diff
@@ -40,5 +40,8 @@
   public static final DateTimeFormatter HTTP_HEADER_DATE_FORMAT =
       DateTimeFormat.forPattern("EEE',' dd MMM yyyy HH':'mm':'ss 'GMT'").withZoneUTC().withLocale(Locale.US);
 
+  public static final DateTimeFormatter RETENTION_DATE_FORMAT =
```
balamurugana: you already have `EXPIRATION_DATE_FORMAT`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3895,6 +3898,107 @@ public ObjectLockConfiguration getDefaultRetention(String bucketName)
...
+   * @throws IOException                 upon connection error
+   * @throws InvalidKeyException
+   *           upon an invalid access key or secret key
+   * @throws NoResponseException         upon no response from server
+   * @throws XmlPullParserException      upon parsing response xml
+   * @throws ErrorResponseException      upon unsuccessful execution
+   * @throws InternalException           upon internal library error
+   * @throws InvalidResponseException    upon a non-xml response from server
+   */
+  public void setObjectLockRetention(String bucketName, String objectName, ObjectLockRetentionConfiguration config)
```
balamurugana: We need to comply with S3 API names to avoid confusion i.e. `setObjectRetention()` and `getObjectRetention()`.  Below are right signature as per the spec.

```java
public void setObjectRetention(String bucketName, String objectName, String versionId, boolean bypassGovernanceRetention, ObjectRetentionConfiguration config)
public ObjectRetentionConfiguration getObjectRetention(String bucketName, String objectName, String versionId)
```

sinhaashish:
@balamurugana  Need one confirmation .
I am using `  public static final DateTimeFormatter RETENTION_DATE_FORMAT = EXPIRATION_DATE_FORMAT;`  for date format.
 this is leading to an exception against MinIo but works fine against S3.

The exception is :
```
Exception in thread "main" java.lang.IllegalArgumentException: Invalid format: "2020-02-03T22:30:00Z" is malformed at "Z"
	at org.joda.time.format.DateTimeFormatter.parseDateTime(DateTimeFormatter.java:899)
	at io.minio.messages.ObjectRetentionConfiguration.retainUntil(ObjectRetentionConfiguration.java:80)
	at SetGetObjectRetentionConfig.main(SetGetObjectRetentionConfig.java:66)
```
As per S3 the retainUntil Date is of this form `2020-01-05T00:00:00.000Z`  [Refer ](https://docs.aws.amazon.com/AmazonS3/latest/API/archive-RESTObjectPUTRetention.html)

But on Minio Server it is stored as ``2020-01-05T00:00:00.00Z`  . Notice one `0` is less in the date format when stored in Minio. This leads to exception while parsing against Minio.

Can you please suggest that is it a server side bug , or do i need to have some change in client side.

balamurugana:
> @balamurugana Need one confirmation .
> I am using ` public static final DateTimeFormatter RETENTION_DATE_FORMAT = EXPIRATION_DATE_FORMAT;` for date format.

You don't need to have `RETENTION_DATE_FORMAT`, instead use `EXPIRATION_DATE_FORMAT` directly.

> this is leading to an exception against MinIo but works fine against S3.
> 
> The exception is :
> 
> ```
> Exception in thread "main" java.lang.IllegalArgumentException: Invalid format: "2020-02-03T22:30:00Z" is malformed at "Z"
> 	at org.joda.time.format.DateTimeFormatter.parseDateTime(DateTimeFormatter.java:899)
> 	at io.minio.messages.ObjectRetentionConfiguration.retainUntil(ObjectRetentionConfiguration.java:80)
> 	at SetGetObjectRetentionConfig.main(SetGetObjectRetentionConfig.java:66)
> ```
> 
> As per S3 the retainUntil Date is of this form `2020-01-05T00:00:00.000Z` [Refer ](https://docs.aws.amazon.com/AmazonS3/latest/API/archive-RESTObjectPUTRetention.html)
> 
> But on Minio Server it is stored as `2020-01-05T00:00:00.00Z` . Notice one `0` is less in the date format when stored in Minio. This leads to exception while parsing against Minio.
> 
> Can you please suggest that is it a server side bug , or do i need to have some change in client side.

Server needs to comply with the spec, else its server bug.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -976,10 +978,15 @@ private Request createRequest(Method method, String bucketName, String objectNam
...
-        } else if (method == Method.PUT && queryParamMap != null && queryParamMap.containsKey("lifecycle")) {
+        } else if (method == Method.PUT && queryParamMap != null &&  queryParamMap.containsKey("lifecycle"))  {
           String[] hashes = Digest.sha256Md5Hashes(data, len);
           sha256Hash = hashes[0];
           md5Hash = hashes[1];
+        } else if (method == Method.PUT && headerMap.containsKey("retention")) {
+          String[] hashes = Digest.sha256Md5Hashes(data, len);
+          sha256Hash = hashes[0];
+          md5Hash = hashes[1];
+          headerMap.remove("retention","");
```
balamurugana: Not a good approach to pass and remove headers or params for sake of checksum calculation. You could pass a flag from higher level calls whether MD5 checksum calculation is required or not. You could add this improvement and remove these nested `if` guesses.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3895,6 +3902,201 @@ public ObjectLockConfiguration getDefaultRetention(String bucketName)
...
+    throws InvalidBucketNameException, NoSuchAlgorithmException, InsufficientDataException, IOException,
+           InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
+           InternalException, InvalidResponseException {
+
+    Map<String, String> queryParamMap = new HashMap<>();
+    queryParamMap.put("retention", "");
+
+    if (!versionId.isEmpty()) {
+      queryParamMap.put("versionId", versionId);
+    }
```
balamurugana: Treat `null` versionId as empty string to support minio server and add it into query param i.e.
```java
if (versionId == null) {
  queryParamMap.put("versionId", "");
} else {
  queryParamMap.put("versionId", versionId);
}
```

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3895,6 +3902,201 @@ public ObjectLockConfiguration getDefaultRetention(String bucketName)
...
+   * @throws InvalidKeyException
+   *           upon an invalid access key or secret key
+   * @throws NoResponseException         upon no response from server
+   * @throws XmlPullParserException      upon parsing response xml
+   * @throws ErrorResponseException      upon unsuccessful execution
+   * @throws InternalException           upon internal library error
+   * @throws InvalidResponseException    upon a non-xml response from server
+   */
+  public void setObjectRetention(String bucketName, String objectName, String versionId, 
+          boolean bypassGovernanceRetention, ObjectRetentionConfiguration config)
```
balamurugana: move this `boolean` as last param looks like its optional.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3895,6 +3902,201 @@ public ObjectLockConfiguration getDefaultRetention(String bucketName)
...
+    if (!versionId.isEmpty()) {
+      queryParamMap.put("versionId", versionId);
+    }
+
+    Map<String, String> headerMap = new HashMap<>();
+    if (bypassGovernanceRetention) {
+      headerMap.put("x-amz-bypass-governance-retention", "True");
+    }
+
+    HttpResponse response = executePut(bucketName, objectName, headerMap, queryParamMap, config, 0);
```
balamurugana: `config` must be non-null and `Mode` and `RetainUntilDate` are mandatory configuration. You would need to add this validation properly.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3895,6 +3902,201 @@ public ObjectLockConfiguration getDefaultRetention(String bucketName)
...
+   * @throws InvalidKeyException
+   *           upon an invalid access key or secret key
+   * @throws NoResponseException         upon no response from server
+   * @throws XmlPullParserException      upon parsing response xml
+   * @throws ErrorResponseException      upon unsuccessful execution
+   * @throws InternalException           upon internal library error
+   * @throws InvalidResponseException    upon a non-xml response from server
+   */
+  public void setObjectLegalHold(String bucketName, String objectName, String versionId, 
+          boolean legalHold)
```
balamurugana: You would need to have below more friendlier APIs than `boolean` flag
```java
public void enableObjectLegalHold(String bucketName, String objectName, String versionId)
public void disableObjectLegalHold(String bucketName, String objectName, String versionId)
public boolean isObjectLegalHoldEnabled(String bucketName, String objectName, String versionId)
```

api/src/main/java/io/minio/messages/ObjectLockLegalHold.java
```diff
@@ -0,0 +1,58 @@
...
+    } else {
+      this.status = "OFF";
+    }
+  }
+
+  /**
+   * Indicates whether the specified object has a Legal Hold in place.
+   */
+  public String getStatus() {
+    return status;
```
balamurugana: You need to retrurn `boolean` depending on `status`; not string

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -4711,6 +4961,92 @@ public void putObject(String bucketName, String objectName, InputStream stream,
...
+              Map<String, String> headerMap, ServerSideEncryption sse, String contentType, boolean objectLockRetention)
+    throws InvalidBucketNameException, NoSuchAlgorithmException, IOException,
+           InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
+           InternalException, InvalidArgumentException, InsufficientDataException, InvalidResponseException {
+
+    if (!(stream instanceof BufferedInputStream)) {
+      stream = new BufferedInputStream(stream);
+    }
+    if (objectLockRetention) {
+      headerMap.put("retention","");
```
balamurugana: Two things you can do at the time of put object.
1. retention (`Mode` and `RetainUntilDate` are required)
2. legal hold (`boolean` to say `ON` or `OFF`)

You could send these changes in put object as separate PR.

sinhaashish: Will send a separate PR for this

sinhaashish:
@balamurugana  PTAL

api/src/main/java/io/minio/DateFormat.java
```diff
@@ -40,5 +40,7 @@
   public static final DateTimeFormatter HTTP_HEADER_DATE_FORMAT =
       DateTimeFormat.forPattern("EEE',' dd MMM yyyy HH':'mm':'ss 'GMT'").withZoneUTC().withLocale(Locale.US);
 
+  public static final DateTimeFormatter RETENTION_DATE_FORMAT = EXPIRATION_DATE_FORMAT;
```
balamurugana: Remove `RETENTION_DATE_FORMAT` and use `EXPIRATION_DATE_FORMAT` directly

api/src/main/java/io/minio/messages/ObjectLockLegalHold.java
```diff
@@ -0,0 +1,58 @@
...
+    } else {
+      this.status = "OFF";
+    }
+  }
+
+  /**
+   * Indicates whether the specified object has a Legal Hold in place or not.
+   */
+  public Boolean getStatus() {
+    return Boolean.parseBoolean(status);
```
balamurugana: 1. I don't see javadoc of `Boolean` parses `"ON"`/`"OFF"` string. Are you sure it works?
2. Return `boolean` not `Boolean`

api/src/main/java/io/minio/messages/ObjectLockLegalHold.java
```diff
@@ -0,0 +1,58 @@
...
+   */
+  public ObjectLockLegalHold() throws XmlPullParserException {
+    super();
+    super.name = "LegalHold";
+  }
+
+  /**
+   * Constructs a new CustomRetention object with given retention.
+   */
+  public ObjectLockLegalHold(boolean legalHold) throws  XmlPullParserException {
```
balamurugana: argument is `status` not `legalHold`

api/src/main/java/io/minio/messages/ObjectRetentionConfiguration.java
```diff
@@ -0,0 +1,80 @@
...
+    super.name = "Retention";
+  }
+
+  /**
+   * Constructs a new CustomRetention object with given retention.
+   */
+  public ObjectRetentionConfiguration(RetentionMode mode,  DateTime retainUntilDate) throws XmlPullParserException,
+          InvalidArgumentException {
+    super();
+    super.name = "Retention";
```
balamurugana: call `this()` than repeating these two lines

api/src/main/java/io/minio/messages/ObjectLockLegalHold.java
```diff
@@ -0,0 +1,58 @@
...
+    super();
+    super.name = "LegalHold";
```
balamurugana: use `this()` and don't repeat logic

api/src/main/java/io/minio/messages/ObjectRetentionConfiguration.java
```diff
@@ -0,0 +1,80 @@
...
+    if (mode == null) {
+      throw new InvalidArgumentException("null is not allowed in mode. Valid values are 'COMPLIANCE' and 'GOVERNANCE'");
```
balamurugana: change error message as `null mode is not allowed`

api/src/main/java/io/minio/messages/ObjectRetentionConfiguration.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+    this.mode = mode.toString();
+    if (retainUntilDate == null) {
+      throw new InvalidArgumentException("retain until date cant be null, it must be a valid date.");
```
balamurugana: change error message as `null retainUntilDate is not allowed`

api/src/main/java/io/minio/messages/ObjectRetentionConfiguration.java
```diff
@@ -0,0 +1,80 @@
...
+   * Returns mode.
+   */
+  public RetentionMode mode() {
+    return RetentionMode.fromString(mode);
+  }
+
+  /**
+   * Returns retain until date.
+   */
+  public Date retainUntil() {
```
balamurugana: Method name should be same as member field `retainUntilDate()` and return `DateTime` not `Date` i.e. should be same as argument type in the constructor

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1402,11 +1400,12 @@ private HttpResponse executeDelete(String bucketName, String objectName, Map<Str
    * @param data           HTTP request body data.
    */
   private HttpResponse executePost(String bucketName, String objectName, Map<String,String> headerMap,
-                                   Map<String,String> queryParamMap, Object data)
+                                   Map<String,String> queryParamMap, Object data, boolean md5Required)
```
balamurugana: Have overloaded `executePost()` i.e. with/without `md5Required` argument.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1439,13 +1438,13 @@ private HttpResponse executePost(String bucketName, String objectName, Map<Strin
    * @param length         Length of HTTP request body data.
    */
   private HttpResponse executePut(String bucketName, String objectName, Map<String,String> headerMap,
-                                  Map<String,String> queryParamMap, String region, Object data, int length)
+                  Map<String,String> queryParamMap, String region, Object data, int length, boolean md5Required)
```
balamurugana: Have overloaded `executePut()` i.e. with/without `md5Required`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1461,11 +1460,12 @@ private HttpResponse executePut(String bucketName, String objectName, Map<String
    * @param length         Length of HTTP request body data.
    */
   private HttpResponse executePut(String bucketName, String objectName, Map<String,String> headerMap,
-                                  Map<String,String> queryParamMap, Object data, int length)
+                                  Map<String,String> queryParamMap, Object data, int length,  boolean md5Required)
```
balamurugana: Same as above

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2388,7 +2388,7 @@ public void copyObject(String bucketName, String objectName, Map<String,String>
       headerMap.putAll(copyConditions.getConditions());
     }
 
-    HttpResponse response = executePut(bucketName, objectName, headerMap, null, "", 0);
+    HttpResponse response = executePut(bucketName, objectName, headerMap, null, "", 0, false);
```
balamurugana: Use overloaded `executePut()`/`executePost()` accepting `md5Required` argument to avoid unnecessary diffs.

api/src/main/java/io/minio/MinioClient.java
balamurugana: You have to use this logic in all methods accepting `versionId`

api/src/main/java/io/minio/messages/ObjectRetentionConfiguration.java
```diff
@@ -0,0 +1,78 @@
...
+      throw new InvalidArgumentException("null retainUntilDate is not allowed");
+    }    
+    this.retainUntilDate = retainUntilDate.toString(DateFormat.EXPIRATION_DATE_FORMAT);  
+  }
+
+  /**
+   * Returns mode.
+   */
+  public RetentionMode mode() {
+    return RetentionMode.fromString(mode);
```
balamurugana: `mode` can be `null`. You need to return `null` accordingly

api/src/main/java/io/minio/messages/ObjectLockLegalHold.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  /**
+   * Indicates whether the specified object has a Legal Hold in place or not.
+   */
+  public boolean status() {
+    if (status.equals("ON")) {
+      return true;
+    }
+    return false;
```
balamurugana: `return status.equals("ON");` is sufficient

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3895,6 +3956,247 @@ public ObjectLockConfiguration getDefaultRetention(String bucketName)
...
+    throws InvalidBucketNameException, NoSuchAlgorithmException, InsufficientDataException, IOException,
+           InvalidKeyException, NoResponseException, XmlPullParserException, ErrorResponseException,
+           InternalException, InvalidResponseException {
+
+    Map<String, String> queryParamMap = new HashMap<>();
+    queryParamMap.put("retention", "");
+
+    if ( !(versionId == null || versionId.isEmpty())) {
+      queryParamMap.put("versionId",versionId);
+    }
```
balamurugana: I don't know why you are missing out of `versionId` handling. Are you deferring in other places?

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -4096,6 +4398,51 @@ public void putObject(String bucketName, String objectName, String fileName,  Lo
...
+   *           upon an invalid access key or secret key
+   * @throws NoResponseException         upon no response from server
+   * @throws XmlPullParserException      upon parsing response xml
+   * @throws ErrorResponseException      upon unsuccessful execution
+   * @throws InternalException           upon internal library error
+   * @throws InvalidArgumentException    upon invalid value is passed to a method.
+   * @throws InsufficientDataException   upon getting EOFException while reading given
+   */
+  public void putObject(String bucketName, String objectName, String fileName, Long size, Map<String, String> headerMap,
+                        ServerSideEncryption sse, String contentType, boolean objectLockRetention)
```
balamurugana: As we agreed previously, you are going to send separate PR to have `putObject()` with retention and legal hold. Are you still considering to add in this PR?

sinhaashish: Yes, that i will send as another PR.

sinhaashish:
@balamurugana  PTAL

api/src/main/java/io/minio/MinioClient.java
balamurugana: If so, why are doing this change here?

sinhaashish: >If so, why are doing this change here?

This change is needed .
 There are two cases to this flow:
 User creates a bucket with lock function enabled. 
**Case 1** : User can put the object on the bucket  and then at later point of time, set the retenetion. This case is handled in this PR.
**Case 2** :  User asks for retention/legalhold at the time of putting object. This will be handled in another PR.

If we dont have the below overloaded putObject then we wont be able to put the object on 
lock enabled buckets and hence not able to test the PR. 
```
  public void putObject(String bucketName, String objectName, String fileName, Long size, Map<String, String> headerMap,
                        ServerSideEncryption sse, String contentType, boolean objectLockRetention) 

``` 

If you think that case 2 should also be present in this PR , then i can come up with that .

balamurugana: You cannot add any changes to putObject() if you plan for separate PR.

PR #844, opened 2020-02-13
optimize putObject() to avoid reading data for known object size

balamurugana:


functional/FunctionalTest.java
```diff
@@ -1623,7 +1656,7 @@ public static void listIncompleteUploads_test1() throws Exception {
     try {
       String objectName = getRandomName();
       try (final InputStream is = new ContentInputStream(6 * MB)) {
-        client.putObject(bucketName, objectName, is, Long.valueOf( 9 * MB), null, null, nullContentType);
+        client.putObject(bucketName, objectName, is, new PutObjectOptions(9 * MB, -1));
```
nitisht: can we have few tests with `partSize` set to specific numbers  e.g. here with `new PutObjectOptions(9 * MB, 3)`

balamurugana: `listIncompleteUploads_test1()` requires to have an incomplete (partial uploads). What do you mean by `PutObjectOptions(9 * MB, 3)`? Have you checked how those parameters mean to `PutObjectOptions()`?

nitisht: > Have you checked how those parameters mean to PutObjectOptions()?

I have checked. What I mean to say is I don't see a test case with `partSize` set to anything other than `-1`. We should have few tests where we pass `partSize` value to confirm if uploads work fine

balamurugana: I am going to change few `PutObjectOptions()` with valid `partSize` on existing tests.

nitisht: thanks

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -4163,9 +4163,7 @@ private void putObject(
       headerMap.putAll(options.sse().headers());
     }
 
-    if (headerMap.get("Content-Type") == null) {
-      headerMap.put("Content-Type", options.contentType());
-    }
+    headerMap.put("Content-Type", options.contentType());
```
kannappanr: `Content-Type` is passed as part of `options' header` will be overwritten by this. Is this the expected behavior?

balamurugana: Please check `PutObjectOptions.contentType()` method for precedence.

PR #852, opened 2020-02-21
refactor http execution to compute md5sum when needed

balamurugana:


harshavardhana:
@balamurugana can you list the behavior pattern here under what conditions what is expected now?

balamurugana:
> @balamurugana can you list the behavior pattern here under what conditions what is expected now?

In an essence, PUT/POST requests contain MD5 sum all the time.  As `putObject()` is mostly interesting here, below are the detail.
1. With `PutObjectOptions()`, API users have control on memory usage on unknown sized stream e.g. `new PutObjectOptions(-1, 5 * MB)` limits 5MB memory usage of each part and 50GB of overall stream size. As we have to hold maximum of 5MB in memory, sha256 and md5 sums are computed.
2. API users have control on memory usage on known sized stream as well e.g. if `new PutObjectOptions(3 * GB, 5 * MB)` limits 5MB memory usage and sha256 and md5 are computed.
3. If `RandomAccessFile` is an input stream, no memory allocation involved, but sha256 and md5 are computed for its parts upload.

Generic behaviour for all PUT/POST requests including `putObject()`.
1. MD5 sum is calculated and sent to all anonymous requests.
2. For insecure HTTP requests, SHA256 and MD5 sums are calculated and sent to all user requests.
3. For secure HTTP requests, MD5 sums is calculated and sent to all user requests.

balamurugana:
@nitisht 
> @balamurugana I tried the below scenario, I expected to see `content-md5` header but don't see it - may be I am missing something, can you take a look

You are not using this PR. This PR removes `STREAMING-AWS4-HMAC-SHA256-PAYLOAD`, but your trace shows.

nitisht:
> @nitisht
> 
> > @balamurugana I tried the below scenario, I expected to see `content-md5` header but don't see it - may be I am missing something, can you take a look
> 
> You are not using this PR. This PR removes `STREAMING-AWS4-HMAC-SHA256-PAYLOAD`, but your trace shows.

I checked out the PR and built the jar, but let me recheck

nitisht:
> I checked out the PR and built the jar, but let me recheck

Looks like I was using the wrong jar. I see `content-md5` with correct jar.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -818,97 +818,55 @@ private void checkObjectName(String objectName) throws IllegalArgumentException
...
-        };
+  private void checkWriteRequestSse(ServerSideEncryption sse) throws IllegalArgumentException {
+    if (sse == null) {
+      return;
+    }
 
-    final SSLContext sslContext = SSLContext.getInstance("SSL");
-    sslContext.init(null, trustAllCerts, new java.security.SecureRandom());
-    final SSLSocketFactory sslSocketFactory = sslContext.getSocketFactory();
+    if (sse.type().requiresTls() && !this.baseUrl.isHttps()) {
```
kannappanr: why do we need `TLS` for `SSE-S3` and `SSE-KMS` ?

nitisht:
@balamurugana could you pls resolve the conflicts

kannappanr:
@balamurugana Can you please rebase this PR again?

PR #853, opened 2020-02-23
Put bucket encryption feature

sinhaashish:
Raising this PR for `PutBucketEncryption` by @manasiK891 to keep a track of changes.


sinhaashish:
Few Comments

I dont think that this code is working, Have you tested it against minio/s3 @manasiK891 ?

1. `putBucketEncryption` using Bucket object is wrong. You cant pass Bucket object as bucket object needs creation time  which the user doesn't know. You must be using string as bucket name.

2. No need to return the response  status in `  public int putBucketEncryption(Bucket bucket, ServerSideEncryption serv)`

3.   Where have you created the request as shown in https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketEncryption.html.

4. If this code of yours; works then you need to add test case in FunctionalTest.java , an example file i.e. how to set/get/delete bucket encryption and API.md file for documentation.

Please complete this PR  so that it can be reviewed again .

sinhaashish:
Closing this for now.

PR #862, opened 2020-03-06
remove all deprecated methods and usage

balamurugana:


balamurugana:
> Changes look fine @balamurugana are you planning to update API.md as well?

There are many PRs change APIs. Updating `API.md` wouldn't be advised for this PR.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -4254,925 +3951,39 @@ public void removeBucket(String bucketName)
...
+   *          Object name in the bucket.
+   * @param length
+   *          Length of object data.
+   * @param data
+   *          Object data.
+   * @param uploadId
+   *          Upload ID of multipart put object.
+   * @param partNumber
+   *          Part number of multipart put object.
    */
```
kannappanr: we should keep the comment about exceptions being thrown to be consistent

balamurugana: Its odd to have javadoc for private methods. I would suggest to remove such unnecessary docs.

kannappanr: ok. got it.

PR #863, opened 2020-03-08
upgrade to gradle v6.2.2 and use strict google java style

balamurugana:


harshavardhana:
PTAL at the conflicts @balamurugana

PR #867, opened 2020-03-18
enhance returned object of listenBucketNotification()

balamurugana:


harshavardhana:
@balamurugana we are planning for a release on Monday, is this required to be in?

balamurugana:
> @balamurugana we are planning for a release on Monday, is this required to be in?

It depends on how deprecating current `listenBucketNotification()`. @kannappanr is checking on it.

PR #900, opened 2020-04-13
handle accelerate and dual-stack aws endpoints

### LINKED ISSUE #893: Add support to accelerated Amazon AWS endpoint

### LINKED ISSUE #894: Add support to dualstack Amazon AWS endpoint

### LINKED ISSUE #895: Add support to probe region for expanded Amazon AWS endpoint

### LINKED ISSUE #897: Update region cache after successful makeBucket()

### LINKED ISSUE #899: Use `us-east-1` region for anonymous requests

### LINKED ISSUE #926: S3 Object Operation returns HTTP 307 Temporary redirect
Details: 
1. When new S3 buckets are created (especially eu regions), the DNS propagation time is around 24 hrs.
2. When a minio client is created with a region (ex: eu-west-3), the S3 Object operations are suppose to send to the specific S3 region end point. Since the region specific end point mapping is not defined in AwsS3Endpoints.java (for eu-west-3), the request goes to global S3 end point.
3. The result will be 307 Temporary Redirect HTTP response ("MIN076", "Temporary redirect due to DNS updates in progress" ) . (For a new bucket created)

Solution:
1. If the request can be send to the region specific S3 end point, then request can be processed successfully
2. The following S3 region specific endpoints are not mapped in AwsS3Endpoints.java - eu-west-3, eu-north-1, eu-south-1, ap-northeast-3, ap-east-1, af-south-1

balamurugana:
Fixes #893
Fixes #894
Fixes #895
Fixes #897
Fixes #899
Fixes #926


kannappanr:
ping @BigUstad @sinhaashish

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -947,55 +977,77 @@ private HttpUrl buildUrl(
       Multimap<String, String> queryParamMap)
       throws IllegalArgumentException, InvalidBucketNameException, NoSuchAlgorithmException {
     if (bucketName == null && objectName != null) {
-      throw new InvalidBucketNameException(
-          NULL_STRING, "null bucket name for object '" + objectName + "'");
+      throw new IllegalArgumentException("null bucket name for object '" + objectName + "'");
```
anjalshireesh, line 980: - This method (`buildUrl`) seems too big. It might be a good idea to refactor it and make it smaller.
- For the AWS specific logic in this method, maybe introduce a new class called `AwsUrlBuilder` and use it from here?

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1317,50 +1373,43 @@ private Response execute(
...
-      updateRegionCache(bucketName);
-      region = BucketRegionCache.INSTANCE.region(bucketName);
-    } else {
-      region = this.region;
+    String region = AwsRegionCache.INSTANCE.get(bucketName);
+    if (region != null) {
+      return region;
+    }
+
+    // Updates region cache for this bucket.
```
anjalshireesh, line 1394: This comment is probably more appropriate just before the line

`AwsRegionCache.INSTANCE.set(bucketName, region);`

functional/FunctionalTest.java
```diff
@@ -4345,7 +4345,7 @@ public static void runTests() throws Exception {
 
     // SSE_S3 and SSE_KMS only work with endpoint="s3.amazonaws.com"
     String requestUrl = endpoint;
-    if (requestUrl.equals("s3.amazonaws.com")) {
+    if (requestUrl.contains(".amazonaws.com")) {
```
anjalshireesh, line 4348: The above comment also needs to change accordingly?

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -734,53 +779,38 @@ public MinioClient(
...
-      this.accessKey = accessKey;
-      this.secretKey = secretKey;
-      this.region = region;
+  private static boolean isAwsDualStackEndpoint(String endpoint) {
+    return endpoint.contains(".dualstack.");
+  }
 
-      return;
-    }
+  private static String extractRegion(String endpoint) {
```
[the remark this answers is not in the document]
balamurugana: * Done
* No. A regular AWS endpoints may contain region.

api/src/main/java/io/minio/MinioClient.java
balamurugana: I think this can be done when we bring builder pattern to constructor

api/src/main/java/io/minio/MinioClient.java
balamurugana: I changed the comment to `// Execute GetBucketLocation REST API to get region of the bucket.` for specific meaning.

functional/FunctionalTest.java
balamurugana: Done

PR #904, opened 2020-04-15
Rectify the validation condition & add functional test

sinhaashish:
Added the functional test case for 
1. `setDefaultRetention`
2. `getDefaultRetention`
3. `enableObjectLegalHold`
4. `disableObjectLegalHold`

I have not  added the test case for `setObjectRetention` and `getObjectRetention` as we need to put the object and then since its WORM , we wont be able to delete the object.

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -3520,11 +3520,10 @@ public class MinioClient {
     }
 
     try (Response response = executeGet(bucketName, objectName, null, queryParamMap)) {
-      ResponseBody body = response.body();
-      Retention retention = Xml.unmarshal(Retention.class, body.charStream());
+      Retention retention = Xml.unmarshal(Retention.class, response.body().charStream());
       return retention;
     } catch (ErrorResponseException e) {
-      if (e.errorResponse().errorCode() == ErrorCode.NO_SUCH_OBJECT_LOCK_CONFIGURATION) {
+      if (e.errorResponse().errorCode() != ErrorCode.NO_SUCH_OBJECT_LOCK_CONFIGURATION) {
         throw e;
       }
     }
@@ -3565,7 +3564,6 @@ public class MinioClient {
     }
 
     LegalHold legalHold = new LegalHold(true);
-
     Response response = executePut(bucketName, objectName, null, queryParamMap, legalHold, 0);
     response.body().close();
   }
@@ -3650,8 +3648,7 @@ public class MinioClient {
     }
 
     try (Response response = executeGet(bucketName, objectName, null, queryParamMap)) {
-      ResponseBody body = response.body();
-      LegalHold result = Xml.unmarshal(LegalHold.class, body.charStream());
+      LegalHold result = Xml.unmarshal(LegalHold.class, response.body().charStream());
       return result.status();
     } catch (ErrorResponseException e) {
       if (e.errorResponse().errorCode() != ErrorCode.NO_SUCH_OBJECT_LOCK_CONFIGURATION) {
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -3469,9 +3469,7 @@ public class MinioClient {
     Map<String, String> queryParamMap = new HashMap<>();
     queryParamMap.put("retention", "");
 
-    if (versionId == null) {
-      queryParamMap.put("versionId", "");
-    } else {
+    if (versionId != null && !versionId.isEmpty()) {
       queryParamMap.put("versionId", versionId);
     }
 
@@ -3517,16 +3515,27 @@ public class MinioClient {
     Map<String, String> queryParamMap = new HashMap<>();
     queryParamMap.put("retention", "");
 
-    if (versionId == null) {
-      queryParamMap.put("versionId", "");
-    } else {
+    if (versionId != null && !versionId.isEmpty()) {
       queryParamMap.put("versionId", versionId);
     }
 
-    Response response = executeGet(bucketName, objectName, null, queryParamMap);
-    try (ResponseBody body = response.body()) {
-      return Xml.unmarshal(Retention.class, body.charStream());
+    Response response = null;
+    Retention retention = new Retention();
+    try {
+      response = executeGet(bucketName, objectName, null, queryParamMap);
+      ResponseBody body = response.body();
+      retention = Xml.unmarshal(Retention.class, body.charStream());
+
+    } catch (ErrorResponseException e) {
+      if (e.errorResponse().errorCode() == ErrorCode.NO_SUCH_OBJECT_LOCK_CONFIGURATION) {
+        throw e;
+      }
+    } finally {
+      if (response != null) {
+        response.body().close();
+      }
     }
+    return retention;
   }
 
   /**
@@ -3558,9 +3567,7 @@ public class MinioClient {
     Map<String, String> queryParamMap = new HashMap<>();
     queryParamMap.put("legal-hold", "");
 
-    if (versionId == null) {
-      queryParamMap.put("versionId", "");
-    } else {
+    if (versionId != null && !versionId.isEmpty()) {
       queryParamMap.put("versionId", versionId);
     }
 
@@ -3599,9 +3606,7 @@ public class MinioClient {
     Map<String, String> queryParamMap = new HashMap<>();
     queryParamMap.put("legal-hold", "");
 
-    if (versionId == null) {
-      queryParamMap.put("versionId", "");
-    } else {
+    if (versionId != null && !versionId.isEmpty()) {
       queryParamMap.put("versionId", versionId);
     }
 
@@ -3641,23 +3646,31 @@ public class MinioClient {
    * @throws XmlParserException thrown to indicate XML parsing error.
    */
   public boolean isObjectLegalHoldEnabled(String bucketName, String objectName, String versionId)
-      throws ErrorResponseException, IllegalArgumentException, InsufficientDataException,
-          InternalException, InvalidBucketNameException, InvalidKeyException,
-          InvalidResponseException, IOException, NoSuchAlgorithmException, XmlParserException {
+      throws IllegalArgumentException, InsufficientDataException, InternalException,
+          InvalidBucketNameException, InvalidKeyException, InvalidResponseException, IOException,
+          NoSuchAlgorithmException, XmlParserException {
     Map<String, String> queryParamMap = new HashMap<>();
     queryParamMap.put("legal-hold", "");
 
-    if (versionId == null) {
-      queryParamMap.put("versionId", "");
-    } else {
+    if (versionId != null && !versionId.isEmpty()) {
       queryParamMap.put("versionId", versionId);
     }
-    Response response = executeGet(bucketName, objectName, null, queryParamMap);
-
-    try (ResponseBody body = response.body()) {
+    Response response = null;
+    try {
+      response = executeGet(bucketName, objectName, null, queryParamMap);
+      ResponseBody body = response.body();
       LegalHold result = Xml.unmarshal(LegalHold.class, body.charStream());
       return result.status();
+    } catch (ErrorResponseException e) {
+      if (e.errorResponse().errorCode() == ErrorCode.NO_SUCH_OBJECT_LOCK_CONFIGURATION) {
+        return false;
+      }
+    } finally {
+      if (response != null) {
+        response.body().close();
+      }
     }
+    return false;
   }
 
   /**
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -3519,23 +3519,16 @@ public class MinioClient {
       queryParamMap.put("versionId", versionId);
     }
 
-    Response response = null;
-    Retention retention = new Retention();
-    try {
-      response = executeGet(bucketName, objectName, null, queryParamMap);
+    try (Response response = executeGet(bucketName, objectName, null, queryParamMap)) {
       ResponseBody body = response.body();
-      retention = Xml.unmarshal(Retention.class, body.charStream());
-
+      Retention retention = Xml.unmarshal(Retention.class, body.charStream());
+      return retention;
     } catch (ErrorResponseException e) {
       if (e.errorResponse().errorCode() == ErrorCode.NO_SUCH_OBJECT_LOCK_CONFIGURATION) {
         throw e;
       }
-    } finally {
-      if (response != null) {
-        response.body().close();
-      }
     }
-    return retention;
+    return null;
   }
 
   /**
@@ -3646,28 +3639,23 @@ public class MinioClient {
    * @throws XmlParserException thrown to indicate XML parsing error.
    */
   public boolean isObjectLegalHoldEnabled(String bucketName, String objectName, String versionId)
-      throws IllegalArgumentException, InsufficientDataException, InternalException,
-          InvalidBucketNameException, InvalidKeyException, InvalidResponseException, IOException,
-          NoSuchAlgorithmException, XmlParserException {
+      throws ErrorResponseException, IllegalArgumentException, InsufficientDataException,
+          InternalException, InvalidBucketNameException, InvalidKeyException,
+          InvalidResponseException, IOException, NoSuchAlgorithmException, XmlParserException {
     Map<String, String> queryParamMap = new HashMap<>();
     queryParamMap.put("legal-hold", "");
 
     if (versionId != null && !versionId.isEmpty()) {
       queryParamMap.put("versionId", versionId);
     }
-    Response response = null;
-    try {
-      response = executeGet(bucketName, objectName, null, queryParamMap);
+
+    try (Response response = executeGet(bucketName, objectName, null, queryParamMap)) {
       ResponseBody body = response.body();
       LegalHold result = Xml.unmarshal(LegalHold.class, body.charStream());
       return result.status();
     } catch (ErrorResponseException e) {
-      if (e.errorResponse().errorCode() == ErrorCode.NO_SUCH_OBJECT_LOCK_CONFIGURATION) {
-        return false;
-      }
-    } finally {
-      if (response != null) {
-        response.body().close();
+      if (e.errorResponse().errorCode() != ErrorCode.NO_SUCH_OBJECT_LOCK_CONFIGURATION) {
+        throw e;
       }
     }
     return false;
--- a/api/src/main/java/io/minio/messages/ObjectLockConfiguration.java
+++ b/api/src/main/java/io/minio/messages/ObjectLockConfiguration.java
@@ -43,9 +43,7 @@ public class ObjectLockConfiguration {
       throws IllegalArgumentException {
     if (mode != null && duration != null) {
       this.rule = new Rule(mode, duration);
-    }
-
-    if (mode == null || duration == null) {
+    } else if (mode != null || duration != null) {
       throw new IllegalArgumentException("mode or duration is null");
     }
   }
--- a/api/src/main/java/io/minio/messages/ObjectLockConfiguration.java
+++ b/api/src/main/java/io/minio/messages/ObjectLockConfiguration.java
@@ -45,7 +45,7 @@ public class ObjectLockConfiguration {
       this.rule = new Rule(mode, duration);
     }
 
-    if (mode != null || duration != null) {
+    if (mode == null || duration == null) {
       throw new IllegalArgumentException("mode or duration is null");
     }
   }
--- a/api/src/main/java/io/minio/messages/ObjectLockConfiguration.java
+++ b/api/src/main/java/io/minio/messages/ObjectLockConfiguration.java
@@ -44,7 +44,10 @@ public class ObjectLockConfiguration {
     if (mode != null && duration != null) {
       this.rule = new Rule(mode, duration);
     } else if (mode != null || duration != null) {
-      throw new IllegalArgumentException("mode or duration is null");
+      if (mode == null) {
+        throw new IllegalArgumentException("mode is null");
+      }
+      throw new IllegalArgumentException("duration is null");
     }
   }
 
--- a/api/src/main/java/io/minio/messages/RetentionDurationDays.java
+++ b/api/src/main/java/io/minio/messages/RetentionDurationDays.java
@@ -25,6 +25,8 @@ public class RetentionDurationDays implements RetentionDuration {
   @Text(required = false)
   private Integer days;
 
+  public RetentionDurationDays() {}
+
   public RetentionDurationDays(int days) {
     this.days = Integer.valueOf(days);
   }
--- a/api/src/main/java/io/minio/messages/RetentionDurationDays.java
+++ b/api/src/main/java/io/minio/messages/RetentionDurationDays.java
@@ -38,4 +38,10 @@ public class RetentionDurationDays implements RetentionDuration {
   public int duration() {
     return days;
   }
+
+  /** Returns RetentionDurationDays as string. */
+  @Override
+  public String toString() {
+    return (days == 1) ? days + " day" : days + " days";
+  }
 }
--- a/api/src/main/java/io/minio/messages/RetentionDurationDays.java
+++ b/api/src/main/java/io/minio/messages/RetentionDurationDays.java
@@ -42,6 +42,9 @@ public class RetentionDurationDays implements RetentionDuration {
   /** Returns RetentionDurationDays as string. */
   @Override
   public String toString() {
-    return (days == 1) ? days + " day" : days + " days";
+    if (days == null) {
+      return "";
+    }
+    return days.toString() + ((days == 1) ? " day" : " days");
   }
 }
--- a/api/src/main/java/io/minio/messages/RetentionDurationYears.java
+++ b/api/src/main/java/io/minio/messages/RetentionDurationYears.java
@@ -42,6 +42,9 @@ public class RetentionDurationYears implements RetentionDuration {
   /** Returns RetentionDurationYears as string. */
   @Override
   public String toString() {
-    return (years == 1) ? years + " year" : years + " years";
+    if (years == null) {
+      return "";
+    }
+    return years.toString() + ((years == 1) ? " year" : " years");
   }
 }
--- a/api/src/main/java/io/minio/messages/RetentionDurationYears.java
+++ b/api/src/main/java/io/minio/messages/RetentionDurationYears.java
@@ -25,6 +25,8 @@ public class RetentionDurationYears implements RetentionDuration {
   @Text(required = false)
   private Integer years;
 
+  public RetentionDurationYears() {}
+
   public RetentionDurationYears(int years) {
     this.years = Integer.valueOf(years);
   }
@@ -35,5 +37,11 @@ public class RetentionDurationYears implements RetentionDuration {
 
   public int duration() {
     return years;
+  }
+
+  /** Returns RetentionDurationYears as string. */
+  @Override
+  public String toString() {
+    return (years == 1) ? years + " year" : years + " years";
   }
 }
--- a/examples/SetGetBucketObjectLockConfig.java
+++ b/examples/SetGetBucketObjectLockConfig.java
@@ -54,10 +54,8 @@ public class SetGetBucketObjectLockConfig {
       // Get object lock configuration
       ObjectLockConfiguration bucketConfig = s3Client.getDefaultRetention("my-bucketname");
 
-      System.out.println("Default retention configuration of bucket");
-      System.out.println("Mode: " + bucketConfig.mode());
-      System.out.println(
-          "Duration: " + bucketConfig.duration().duration() + " " + bucketConfig.duration().unit());
+      System.out.print("Default retention configuration of bucket is ");
+      System.out.println(bucketConfig.mode() + " mode for " + bucketConfig.duration().toString());
     } catch (MinioException e) {
       System.out.println("Error occurred: " + e);
     }
--- a/examples/SetGetBucketObjectLockConfig.java
+++ b/examples/SetGetBucketObjectLockConfig.java
@@ -56,7 +56,8 @@ public class SetGetBucketObjectLockConfig {
 
       System.out.println("Default retention configuration of bucket");
       System.out.println("Mode: " + bucketConfig.mode());
-      System.out.println("Duration: " + bucketConfig.duration());
+      System.out.println(
+          "Duration: " + bucketConfig.duration().duration() + " " + bucketConfig.duration().unit());
     } catch (MinioException e) {
       System.out.println("Error occurred: " + e);
     }
--- a/examples/SetGetBucketObjectLockConfig.java
+++ b/examples/SetGetBucketObjectLockConfig.java
@@ -54,8 +54,9 @@ public class SetGetBucketObjectLockConfig {
       // Get object lock configuration
       ObjectLockConfiguration bucketConfig = s3Client.getDefaultRetention("my-bucketname");
 
-      System.out.print("Default retention configuration of bucket is ");
-      System.out.println(bucketConfig.mode() + " mode for " + bucketConfig.duration().toString());
+      System.out.println("Default retention configuration of bucket");
+      System.out.println("Mode: " + bucketConfig.mode());
+      System.out.println("Duration: " + bucketConfig.duration());
     } catch (MinioException e) {
       System.out.println("Error occurred: " + e);
     }
--- a/functional/FunctionalTest.java
+++ b/functional/FunctionalTest.java
@@ -3284,6 +3284,196 @@ public class FunctionalTest {
     }
   }
 
+  /**
+   * Test: enableObjectLegalHold(String bucketName, String objectName, String versionId)
+   * disableObjectLegalHold(String bucketName, String objectName, String versionId)
+   */
+  public static void enableDisableObjectLegalHold_test() throws Exception {
+    if (!mintEnv) {
+      System.out.println(
+          "Test: enableObjectLegalHold(String bucketName, String objectName, String versionId)");
+      System.out.println(
+          "Test: disableObjectLegalHold(String bucketName, String objectName, String versionId)");
+    }
+    long startTime = System.currentTimeMillis();
+    String bucketNameLegalHold = getRandomName();
+    String objectName = getRandomName();
+    try {
+      client.makeBucket(bucketNameLegalHold, null, true);
+      StringBuilder builder = new StringBuilder();
+      for (int i = 0; i < 10; i++) {
+        builder.append(
+            "Sphinx of black quartz, judge my vow: Used by Adobe InDesign to display font samples. ");
+        builder.append("(29 letters)\n");
+        builder.append(
+            "Jackdaws love my big sphinx of quartz: Similarly, used by Windows XP for some fonts. ");
+        builder.append("---\n");
+      }
+      // Create a InputStream for object upload.
+      ByteArrayInputStream bais = new ByteArrayInputStream(builder.toString().getBytes("UTF-8"));
+      client.putObject(
+          bucketNameLegalHold, objectName, bais, new PutObjectOptions(bais.available(), -1));
+      bais.close();
+
+      // Enable object legal hold.
+      client.enableObjectLegalHold(bucketNameLegalHold, objectName, "");
+
+      if (!client.isObjectLegalHoldEnabled(bucketNameLegalHold, objectName, "")) {
+        System.out.println(
+            "FAILED. expected Object Legal Hold Enabled to be : "
+                + "True"
+                + ", got: "
+                + client.isObjectLegalHoldEnabled(bucketNameLegalHold, objectName, ""));
+      }
+      // Enable object legal hold.
+      client.disableObjectLegalHold(bucketNameLegalHold, objectName, "");
+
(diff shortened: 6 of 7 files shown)


harshavardhana:
Let's add a unit test also please

examples/SetGetBucketObjectLockConfig.java
```diff
@@ -56,7 +56,8 @@ public static void main(String[] args)
 
       System.out.println("Default retention configuration of bucket");
       System.out.println("Mode: " + bucketConfig.mode());
-      System.out.println("Duration: " + bucketConfig.duration());
+      System.out.println(
+          "Duration: " + bucketConfig.duration().duration() + " " + bucketConfig.duration().unit());
```
nitisht, line 60: Shouldn't this be `bucketConfig.duration()` instead of `bucketConfig.duration().duration()`

sinhaashish: `bucketConfig.duration()` returns a `RetentionDuration` object which actually contains the `duration` and `unit` i.e. days or years. 
Returning `bucketConfig.duration()`  will  just print the object which is  composed of the name of the class, an @ symbol and the hashcode of the object in hexadecimal. 
We need to fetch the exact duration and unit and thus this change is needed.

api/src/main/java/io/minio/messages/RetentionDurationDays.java
```diff
@@ -25,6 +25,8 @@
   @Text(required = false)
   private Integer days;
 
+  public RetentionDurationDays() {}
```
balamurugana, line 28: Have no arg constructor to `RetentionDurationYears` too.

examples/SetGetBucketObjectLockConfig.java
balamurugana: Have `toString()` to `RetentionDurationDays` and `RetentionDurationYears` classes to have friendlier message.

api/src/main/java/io/minio/messages/RetentionDurationDays.java
```diff
@@ -36,4 +38,10 @@ public RetentionDurationUnit unit() {
   public int duration() {
     return days;
   }
+
+  /** Returns RetentionDurationDays as string. */
+  @Override
+  public String toString() {
+    return (days == 1) ? days + " day" : days + " days";
```
balamurugana, line 45: `days` may be `null`. Below is the right way to fix
```java
if (days == null) {
  return null;
}

return days.toString() + ((days == 1) ? " day" : " days");
```

api/src/main/java/io/minio/messages/RetentionDurationYears.java
```diff
@@ -36,4 +38,10 @@ public RetentionDurationUnit unit() {
   public int duration() {
     return years;
   }
+
+  /** Returns RetentionDurationYears as string. */
+  @Override
+  public String toString() {
+    return (years == 1) ? years + " year" : years + " years";
```
balamurugana, line 45: same as above

examples/SetGetBucketObjectLockConfig.java
```diff
@@ -54,9 +54,8 @@ public static void main(String[] args)
       // Get object lock configuration
       ObjectLockConfiguration bucketConfig = s3Client.getDefaultRetention("my-bucketname");
 
-      System.out.println("Default retention configuration of bucket");
-      System.out.println("Mode: " + bucketConfig.mode());
-      System.out.println("Duration: " + bucketConfig.duration());
+      System.out.print("Default retention configuration of bucket is ");
+      System.out.println(bucketConfig.mode() + " mode for " + bucketConfig.duration().toString());
```
balamurugana, line 58: no change required

examples/SetGetBucketObjectLockConfig.java
```diff
@@ -54,7 +54,7 @@ public static void main(String[] args)
       // Get object lock configuration
       ObjectLockConfiguration bucketConfig = s3Client.getDefaultRetention("my-bucketname");
 
-      System.out.println("Default retention configuration of bucket");
+      System.out.print("Default retention configuration of bucket  ");
```
balamurugana, line 57: Why not a new line?

api/src/main/java/io/minio/messages/ObjectLockConfiguration.java
```diff
@@ -45,7 +45,7 @@ public ObjectLockConfiguration(RetentionMode mode, RetentionDuration duration)
       this.rule = new Rule(mode, duration);
     }
 
-    if (mode != null || duration != null) {
+    if (mode == null || duration == null) {
```
balamurugana, line 48: This fix is incorrect. Either both values non-null or null. i.e. `mode != null && duration != null` and `mode == null && duration == null` are valid.

Basically you need to have current `if` as `else if`

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,196 @@ public static void composeObject_test6() throws Exception {
...
+      for (int i = 0; i < 10; i++) {
+        builder.append(
+            "Sphinx of black quartz, judge my vow: Used by Adobe InDesign to display font samples. ");
+        builder.append("(29 letters)\n");
+        builder.append(
+            "Jackdaws love my big sphinx of quartz: Similarly, used by Windows XP for some fonts. ");
+        builder.append("---\n");
+      }
+      // Create a InputStream for object upload.
+      ByteArrayInputStream bais = new ByteArrayInputStream(builder.toString().getBytes("UTF-8"));
```
balamurugana, line 3313: use `ContentInputStream` here

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,196 @@ public static void composeObject_test6() throws Exception {
...
+  /**
+   * Test: enableObjectLegalHold(String bucketName, String objectName, String versionId)
+   * disableObjectLegalHold(String bucketName, String objectName, String versionId)
+   */
+  public static void enableDisableObjectLegalHold_test() throws Exception {
+    if (!mintEnv) {
+      System.out.println(
+          "Test: enableObjectLegalHold(String bucketName, String objectName, String versionId)");
+      System.out.println(
+          "Test: disableObjectLegalHold(String bucketName, String objectName, String versionId)");
```
balamurugana, line 3296: we would need to have two separate tests here.
1. `enableObjectLegalHold()` - enable legal hold on an object; on success, overwrite/delete object should fail
2. `disableObjectLegalHold()` - disable legal hold on an object; on success, overwrite/delete object should succeed

api/src/main/java/io/minio/messages/ObjectLockConfiguration.java
```diff
@@ -43,9 +43,7 @@ public ObjectLockConfiguration(RetentionMode mode, RetentionDuration duration)
       throws IllegalArgumentException {
     if (mode != null && duration != null) {
       this.rule = new Rule(mode, duration);
-    }
-
-    if (mode == null || duration == null) {
+    } else if (mode != null || duration != null) {
       throw new IllegalArgumentException("mode or duration is null");
```
harshavardhana, line 47: Can we not elaborate a bit more on the error here? by doing 

```
if (mode != null && duration != null) {
} else if (mode == null) {
       IllegalArgumentException("mode is null");
} else if (duration == null) {
       IllegalArgumentException("duration is null");
}
```

api/src/main/java/io/minio/messages/ObjectLockConfiguration.java
```diff
@@ -43,10 +43,10 @@ public ObjectLockConfiguration(RetentionMode mode, RetentionDuration duration)
...
     if (mode != null && duration != null) {
       this.rule = new Rule(mode, duration);
-    }
-
-    if (mode != null || duration != null) {
-      throw new IllegalArgumentException("mode or duration is null");
+    } else if (mode == null) {
+      throw new IllegalArgumentException("mode is null");
+    } else if (duration == null) {
+      throw new IllegalArgumentException("duration is null");
```
balamurugana, line 49: This is wrong check. You would need to preserve current `if` with `else if` with specific error message is requested
```java
    else if (mode != null || duration != null) {
      if (mode == null) {
        throw new IllegalArgumentException("mode is null");
      }
      throw new IllegalArgumentException("duration is null");
    }
```

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
   }
 
+  /** Test: enableObjectLegalHold(String bucketName, String objectName, String versionId) */
+  public static void enableObjectLegalHold_test() throws Exception {
+    if (!mintEnv) {
+      System.out.println(
+          "Test: enableObjectLegalHold(String bucketName, String objectName, String versionId)");
+    }
+    long startTime = System.currentTimeMillis();
+    String bucketNameLegalHold = getRandomName();
```
balamurugana, line 3294: `bucketName` is right name here.

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+      }
+
+      try {
+        try (final InputStream is = new ContentInputStream(1 * KB)) {
+          client.putObject(bucketNameLegalHold, objectName, is, new PutObjectOptions(1 * KB, -1));
+        }
+      } catch (ErrorResponseException e) {
+        if (!e.errorResponse()
+            .message()
+            .equals("Object is WORM protected and cannot be overwritten")) {
```
balamurugana, line 3320: Is this error message same in AWS S3?

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+  }
+
+  /** Test: disableObjectLegalHold(String bucketName, String objectName, String versionId) */
+  public static void disableObjectLegalHold_test() throws Exception {
+    if (!mintEnv) {
+      System.out.println(
+          "Test: disableObjectLegalHold(String bucketName, String objectName, String versionId)");
+    }
+    long startTime = System.currentTimeMillis();
+    String bucketNameLegalHold = getRandomName();
```
balamurugana, line 3385: same as above

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+  public static void setDefaultObjectLock_test() throws Exception {
+    if (!mintEnv) {
+      System.out.println("Test: setDefaultRetention(String bucketName)");
+    }
+
+    long startTime = System.currentTimeMillis();
+
+    try {
+
+      String bucketNameWithLock = getRandomName();
```
balamurugana, line 3480: just `bucketName` is good enough

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+            startTime,
+            null,
+            e.toString() + " >>> " + Arrays.toString(e.getStackTrace()));
+        throw e;
+      }
+    }
+  }
+
+  /** Test: setDefaultRetention(String bucketName). */
+  public static void setDefaultObjectLock_test() throws Exception {
```
balamurugana, line 3471: align with method name i.e. `setDefaultRetention`

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+            startTime,
+            null,
+            e.toString() + " >>> " + Arrays.toString(e.getStackTrace()));
+        throw e;
+      }
+    }
+  }
+
+  /** Test: getDefaultRetention(String bucketName). */
+  public static void getDefaultObjectLock_test() throws Exception {
```
balamurugana, line 3514: same as above

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+    if (!mintEnv) {
+      System.out.println("Test: getDefaultRetention(String bucketName)");
+    }
+
+    long startTime = System.currentTimeMillis();
+    String bucketNameWithLock = getRandomName();
```
balamurugana, line 3520: same as above

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+      client.makeBucket(bucketNameWithLock, null, true);
+      // Declaring config with Retention mode as Compliance and duration as 10 day
+      ObjectLockConfiguration config =
+          new ObjectLockConfiguration(RetentionMode.COMPLIANCE, new RetentionDurationDays(10));
+      // Set object lock configuration
+      client.setDefaultRetention(bucketNameWithLock, config);
+      // Get object lock configuration
+      ObjectLockConfiguration bucketConfig = client.getDefaultRetention(bucketNameWithLock);
+
+      if (!bucketConfig.mode().equals(RetentionMode.COMPLIANCE)) {
```
balamurugana, line 3532: use `==` operator to avoid `NullPointerException`

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+    String bucketNameLegalHold = getRandomName();
+    String objectName = getRandomName();
+    try {
+      client.makeBucket(bucketNameLegalHold, null, true);
+      try (final InputStream is = new ContentInputStream(1 * KB)) {
+        client.putObject(bucketNameLegalHold, objectName, is, new PutObjectOptions(1 * KB, -1));
+      }
+
+      // Enable object legal hold.
+      client.enableObjectLegalHold(bucketNameLegalHold, objectName, "");
```
balamurugana, line 3303: For AWS S3, `versionId` should be passed i.e. this call will fail. Can you confirm?

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+
+      if (!client.isObjectLegalHoldEnabled(bucketNameLegalHold, objectName, "")) {
+        System.out.println(
```
balamurugana, line 3306: you are not capturing this error in case of mint mode.

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+            "FAILED. expected Object Legal Hold Enabled to be : "
+                + "True"
+                + ", got: "
+                + client.isObjectLegalHoldEnabled(bucketNameLegalHold, objectName, ""));
+      }
+
+      try {
```
balamurugana, line 3313: this is meaningless if above condition is `true`

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+              e.toString() + " >>> " + Arrays.toString(e.getStackTrace()));
+          throw e;
+        }
+      }
+      try {
+        client.removeObject(bucketNameLegalHold, objectName);
+      } catch (ErrorResponseException e) {
+        if (!e.errorResponse()
+            .message()
+            .equals("Object is WORM protected and cannot be overwritten")) {
```
balamurugana, line 3335: same as above

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+              startTime,
+              null,
+              e.toString() + " >>> " + Arrays.toString(e.getStackTrace()));
+          throw e;
+        }
+      }
+
+      client.disableObjectLegalHold(bucketNameLegalHold, objectName, "");
+      client.removeObject(bucketNameLegalHold, objectName);
+      client.removeBucket(bucketNameLegalHold);
```
balamurugana, line 3348: this cleanup won't be executed if one of above conditions throw exceptions. you would need to fix it accordingly.

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+      client.makeBucket(bucketNameLegalHold, null, true);
+      try (final InputStream is = new ContentInputStream(1 * KB)) {
+        client.putObject(bucketNameLegalHold, objectName, is, new PutObjectOptions(1 * KB, -1));
+      }
+
+      // Enable object legal hold.
+      client.enableObjectLegalHold(bucketNameLegalHold, objectName, "");
+
+      if (!client.isObjectLegalHoldEnabled(bucketNameLegalHold, objectName, "")) {
+        System.out.println(
```
balamurugana, line 3397: same as above

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+                + "True"
+                + ", got: "
+                + client.isObjectLegalHoldEnabled(bucketNameLegalHold, objectName, ""));
+      }
+
+      // Enable object legal hold.
+      client.disableObjectLegalHold(bucketNameLegalHold, objectName, "");
+
+      if (client.isObjectLegalHoldEnabled(bucketNameLegalHold, objectName, "")) {
+        System.out.println(
```
balamurugana, line 3408: same as above

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+            "FAILED. expected Object Legal Hold Enabled to be : "
+                + "False"
+                + ", got: "
+                + client.isObjectLegalHoldEnabled(bucketNameLegalHold, objectName, ""));
+      }
+
+      try {
+        try (final InputStream is = new ContentInputStream(1 * KB)) {
```
balamurugana, line 3416: no need to have inner `try` block.

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+          client.putObject(bucketNameLegalHold, objectName, is, new PutObjectOptions(1 * KB, -1));
+        }
+      } catch (Exception e) {
```
balamurugana, line 3419: repetitive `catch` block here.

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+            "disableObjectLegalHold",
+            "(String bucketName, String objectName, String versionId)",
+            startTime,
+            null,
+            e.toString() + " >>> " + Arrays.toString(e.getStackTrace()));
+        throw e;
+      }
+      try {
+        client.removeObject(bucketNameLegalHold, objectName);
+      } catch (Exception e) {
```
balamurugana, line 3430: repetitive `catch` block here.

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+        mintFailedLog(
+            "disableObjectLegalHold",
+            "(String bucketName, String objectName, String versionId)",
+            startTime,
+            null,
+            e.toString() + " >>> " + Arrays.toString(e.getStackTrace()));
+        throw e;
+      }
+      try {
```
balamurugana, line 3428: this `try` has no meaning

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+        mintFailedLog(
+            "disableObjectLegalHold",
+            "(String bucketName, String objectName, String versionId)",
+            startTime,
+            null,
+            e.toString() + " >>> " + Arrays.toString(e.getStackTrace()));
+        throw e;
+      }
+
+      client.removeBucket(bucketNameLegalHold);
```
balamurugana, line 3440: this clean up won't be executed if above throws an exception.

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+      System.out.println("Test: getDefaultRetention(String bucketName)");
+    }
+
+    long startTime = System.currentTimeMillis();
+    String bucketNameWithLock = getRandomName();
+    try {
+      // Create bucket with object lock functionality enabled
+      client.makeBucket(bucketNameWithLock, null, true);
+      // Declaring config with Retention mode as Compliance and duration as 10 day
+      ObjectLockConfiguration config =
```
balamurugana, line 3525: rename has `expectedConfig`

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+          new ObjectLockConfiguration(RetentionMode.COMPLIANCE, new RetentionDurationDays(10));
+      // Set object lock configuration
+      client.setDefaultRetention(bucketNameWithLock, config);
+      // Get object lock configuration
+      ObjectLockConfiguration bucketConfig = client.getDefaultRetention(bucketNameWithLock);
```
balamurugana, line 3530: have it as `config`

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+
+      if (!bucketConfig.mode().equals(RetentionMode.COMPLIANCE)) {
+        System.out.println(
```
balamurugana, line 3533: this should be captured in mint mode as well

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+            "FAILED. expected mode : "
+                + RetentionMode.COMPLIANCE
+                + ", got: "
+                + bucketConfig.mode());
+      }
+
+      if (!(bucketConfig.duration().duration() + " " + bucketConfig.duration().unit())
+          .equals("10 DAYS")) {
```
balamurugana, line 3541: This check should be optimal
```java
config.duration().unit() == expectedConfig.duration().unit() && config.duration().duration() == expectedConfig.duration().duration()
```

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+        System.out.println(
```
balamurugana, line 3542: capture in mint mode too

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+            "FAILED. expected duration : "
+                + "10 DAYS"
+                + ", got: "
+                + (bucketConfig.duration().duration() + " " + bucketConfig.duration().unit()));
```
balamurugana, line 3546: there is already stringifier.

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+                + RetentionMode.COMPLIANCE
+                + ", got: "
+                + bucketConfig.mode());
+      }
+
+      if (!(bucketConfig.duration().duration() + " " + bucketConfig.duration().unit())
+          .equals("10 DAYS")) {
+        System.out.println(
+            "FAILED. expected duration : "
+                + "10 DAYS"
```
balamurugana, line 3544: you should use `expectedConfig.duration()`

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,293 @@ public static void composeObject_test6() throws Exception {
...
+                + ", got: "
+                + (bucketConfig.duration().duration() + " " + bucketConfig.duration().unit()));
+      }
+
+      client.removeBucket(bucketNameWithLock);
```
balamurugana, line 3549: cleanup issue in case of exception raised.

functional/FunctionalTest.java
sinhaashish: Modified the code in MinioClient.java to 
```
    if (versionId != null && !versionId.isEmpty()) {
      queryParamMap.put("versionId", versionId);
    }
``` 
So here in test case if we pass `""` , the versionId is not set and this test passes against S3.

balamurugana: We may need to use this logic in all methods accepting `versionId` argument.

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,254 @@ public static void composeObject_test6() throws Exception {
...
+            "enableObjectLegalHold(String bucketName, String objectName, String versionId)",
+            null,
+            startTime,
+            null,
+            e.toString() + " >>> " + Arrays.toString(e.getStackTrace()));
+        throw e;
+      }
+    } finally {
+      client.removeObject(bucketName, objectName);
+      client.removeBucket(bucketName);
```
balamurugana, line 3346: This is problematic to have an failure here. Below logic is better to handle corners.
```java
    try {
      client.makeBucket(...);
      try {
        client.putObject(...);
        client.enableObjectLegalHold(...);
        if (!client.isObjectLegalHoldEnabled(...)) {
          throw Exception(...);
        }
        client.disableObjectLegalHold(...);
        mintSucessLog(...);
      } finally {
        client.removeObject(...);
        client.removeBucket(...);
      }
    } catch (Exception e) {
      ErrorResponse errorResponse = null;
      if (e instanceof ErrorResponseException) {
        ErrorResponseException exp = (ErrorResponseException) e;
        errorResponse = exp.errorResponse();
      }

      // Ignore NotImplemented error
      if (errorResponse != null && errorResponse.errorCode() == ErrorCode.NOT_IMPLEMENTED) {
        mintIgnoredLog(
            "enableObjectLegalHold(String bucketName, String objectName, String versionId)",
            null,
            startTime);
      } else {
        mintFailedLog(
            "enableObjectLegalHold(String bucketName, String objectName, String versionId)",
            null,
            startTime,
            null,
            e.toString() + " >>> " + Arrays.toString(e.getStackTrace()));
        throw e;
      }
    }
```

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,254 @@ public static void composeObject_test6() throws Exception {
...
+    String bucketName = getRandomName();
+    String objectName = getRandomName();
+    try {
+      client.makeBucket(bucketName, null, true);
+      try (final InputStream is = new ContentInputStream(1 * KB)) {
+        client.putObject(bucketName, objectName, is, new PutObjectOptions(1 * KB, -1));
+      }
+
+      // Enable object legal hold.
+      client.enableObjectLegalHold(bucketName, objectName, "");
```
balamurugana, line 3303: return `null` to `versionId` than empty string. you can fix this everywhere in `FunctionalTest`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3641,23 +3646,31 @@ public void disableObjectLegalHold(String bucketName, String objectName, String
    * @throws XmlParserException thrown to indicate XML parsing error.
    */
   public boolean isObjectLegalHoldEnabled(String bucketName, String objectName, String versionId)
-      throws ErrorResponseException, IllegalArgumentException, InsufficientDataException,
-          InternalException, InvalidBucketNameException, InvalidKeyException,
-          InvalidResponseException, IOException, NoSuchAlgorithmException, XmlParserException {
+      throws IllegalArgumentException, InsufficientDataException, InternalException,
+          InvalidBucketNameException, InvalidKeyException, InvalidResponseException, IOException,
+          NoSuchAlgorithmException, XmlParserException {
```
balamurugana, line 3651: We still throw `ErrorResponseException`, hence there is no change here.

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,254 @@ public static void composeObject_test6() throws Exception {
...
+        throw e;
+      }
+    } finally {
+      client.removeObject(bucketName, objectName);
+      client.removeBucket(bucketName);
+    }
+  }
+
+  /** Test: disableObjectLegalHold(String bucketName, String objectName, String versionId) */
+  public static void disableObjectLegalHold_test() throws Exception {
```
balamurugana, line 3351: above logic applies here.

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,254 @@ public static void composeObject_test6() throws Exception {
...
+        throw e;
+      }
+    } finally {
+      client.removeObject(bucketName, objectName);
+      client.removeBucket(bucketName);
+    }
+  }
+
+  /** Test: setDefaultRetention(String bucketName). */
+  public static void setDefaultRetention_test() throws Exception {
```
balamurugana, line 3421: above logic applies here.

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,254 @@ public static void composeObject_test6() throws Exception {
...
+            e.toString() + " >>> " + Arrays.toString(e.getStackTrace()));
+        throw e;
+      }
+    } finally {
+      client.removeBucket(bucketName);
+    }
+  }
+
+  /** Test: getDefaultRetention(String bucketName). */
+  public static void getDefaultRetention_test() throws Exception {
```
balamurugana, line 3464: above logic applies here.

functional/FunctionalTest.java
sinhaashish: AWS doesn't error out if  version is  not passed. So removed this check.

functional/FunctionalTest.java
sinhaashish: added finally block for cleanup

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3517,16 +3515,20 @@ public Retention getObjectRetention(String bucketName, String objectName, String
...
+    if (versionId != null && !versionId.isEmpty()) {
       queryParamMap.put("versionId", versionId);
     }
 
-    Response response = executeGet(bucketName, objectName, null, queryParamMap);
-    try (ResponseBody body = response.body()) {
-      return Xml.unmarshal(Retention.class, body.charStream());
+    try (Response response = executeGet(bucketName, objectName, null, queryParamMap)) {
+      ResponseBody body = response.body();
+      Retention retention = Xml.unmarshal(Retention.class, body.charStream());
```
balamurugana, line 3524: you could directly use `response.body().charStream()`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3517,16 +3515,20 @@ public Retention getObjectRetention(String bucketName, String objectName, String
...
+      return retention;
+    } catch (ErrorResponseException e) {
+      if (e.errorResponse().errorCode() == ErrorCode.NO_SUCH_OBJECT_LOCK_CONFIGURATION) {
```
balamurugana, line 3527: wrong check. `e.errorResponse().errorCode() != ErrorCode.NO_SUCH_OBJECT_LOCK_CONFIGURATION`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3647,17 +3645,20 @@ public boolean isObjectLegalHoldEnabled(String bucketName, String objectName, St
...
-    } else {
+    if (versionId != null && !versionId.isEmpty()) {
       queryParamMap.put("versionId", versionId);
     }
-    Response response = executeGet(bucketName, objectName, null, queryParamMap);
 
-    try (ResponseBody body = response.body()) {
+    try (Response response = executeGet(bucketName, objectName, null, queryParamMap)) {
+      ResponseBody body = response.body();
       LegalHold result = Xml.unmarshal(LegalHold.class, body.charStream());
```
balamurugana, line 3654: same as above

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,241 @@ public static void composeObject_test6() throws Exception {
...
+        boolean expectedStatus = true;
+        client.enableObjectLegalHold(bucketName, objectName, null);
+
+        boolean isObjectLockEnabled = client.isObjectLegalHoldEnabled(bucketName, objectName, null);
+        if (!isObjectLockEnabled) {
+          throw new Exception(
+              "[FAILED] Expected: Is Object Legal Hold Enabled to be "
+                  + expectedStatus
+                  + ", Got: "
+                  + isObjectLockEnabled);
```
balamurugana, line 3312: you could directly use and remove unused variables.
```java
if (!client.isObjectLegalHoldEnabled(bucketName, objectName, null)) {
  throw new Exception("[FAILED] isObjectLegalHoldEnabled(): expected: true, got: false");
}
```

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,241 @@ public static void composeObject_test6() throws Exception {
...
+        // Enable object legal hold.
+        client.enableObjectLegalHold(bucketName, objectName, null);
+
+        // Enable object legal hold.
+        boolean expectedStatus = false;
+        client.disableObjectLegalHold(bucketName, objectName, null);
+
+        boolean isObjectLockEnabled = client.isObjectLegalHoldEnabled(bucketName, objectName, null);
+
+        if (isObjectLockEnabled) {
```
balamurugana, line 3373: you could directly use and remove unused variables.
```java
if (client.isObjectLegalHoldEnabled(bucketName, objectName, null)) {
  throw new Exception("[FAILED] isObjectLegalHoldEnabled(): expected: false, got: true");
}
```

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,241 @@ public static void composeObject_test6() throws Exception {
...
+    }
+
+    long startTime = System.currentTimeMillis();
+    String bucketName = getRandomName();
+
+    try {
+      // Create bucket with object lock functionality enabled
+      client.makeBucket(bucketName, null, true);
+      try {
+        // Declaring config with Retention mode as Compliance and duration as 10 day
```
balamurugana, line 3426: you can remove these obvious comments.

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,241 @@ public static void composeObject_test6() throws Exception {
...
+        ObjectLockConfiguration config =
+            new ObjectLockConfiguration(RetentionMode.COMPLIANCE, new RetentionDurationDays(10));
+
+        // Set object lock configuration
```
balamurugana, line 3430: you can remove these obvious comments everywhere.

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,241 @@ public static void composeObject_test6() throws Exception {
...
+
+    long startTime = System.currentTimeMillis();
+    String bucketName = getRandomName();
+    try {
+      // Create bucket with object lock functionality enabled
+      client.makeBucket(bucketName, null, true);
+      try {
+        // Declaring config with Retention mode as Compliance and duration as 10 day
+        ObjectLockConfiguration expectedConfig =
+            new ObjectLockConfiguration(RetentionMode.COMPLIANCE, new RetentionDurationDays(10));
```
balamurugana, line 3472: you can use `new RetentionDurationYears(N)` for completeness.

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,230 @@ public static void composeObject_test6() throws Exception {
...
+      try {
+        try (final InputStream is = new ContentInputStream(1 * KB)) {
+          client.putObject(bucketName, objectName, is, new PutObjectOptions(1 * KB, -1));
+        }
+
+        client.enableObjectLegalHold(bucketName, objectName, null);
+        if (!client.isObjectLegalHoldEnabled(bucketName, objectName, null)) {
+          throw new Exception("[FAILED] isObjectLegalHoldEnabled(): expected: true, got: false");
+        }
+        client.disableObjectLegalHold(bucketName, objectName, null);
```
nitisht, line 3307: After setting `enableObjectLegalHold` we should attempt deleting the object and confirm that fails.

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,230 @@ public static void composeObject_test6() throws Exception {
...
+    String bucketName = getRandomName();
+    String objectName = getRandomName();
+    try {
+      client.makeBucket(bucketName, null, true);
+      try {
+        try (final InputStream is = new ContentInputStream(1 * KB)) {
+          client.putObject(bucketName, objectName, is, new PutObjectOptions(1 * KB, -1));
+        }
+        client.enableObjectLegalHold(bucketName, objectName, null);
+        client.disableObjectLegalHold(bucketName, objectName, null);
```
nitisht, line 3357: Here after `disableObjectLegalHold` we can attempt a delete and confirm that works

functional/FunctionalTest.java
```diff
@@ -3284,6 +3284,230 @@ public static void composeObject_test6() throws Exception {
...
+
+    long startTime = System.currentTimeMillis();
+    String bucketName = getRandomName();
+
+    try {
+      client.makeBucket(bucketName, null, true);
+      try {
+        ObjectLockConfiguration config =
+            new ObjectLockConfiguration(RetentionMode.COMPLIANCE, new RetentionDurationDays(10));
+        client.setDefaultRetention(bucketName, config);
```
nitisht, line 3408: Here also we can attempt an upload and confirm the uploaded object can't be deleted.

balamurugana: If we do so, we can't do clean up of the test.

functional/FunctionalTest.java
sinhaashish: This was earlier implemented, but removed as of now as it works with MinIO, but when ran against S3 with versionId as null it succeeds. (deleted the object).
You need specific object version to fail thus it was removed.

functional/FunctionalTest.java
nitisht: IMO Testing the MinIO Server's behaviour is more important, since this test will become part of Mint. It will be used to validate MinIO Server behaviour. We generally test against Play or Local MinIO instances - in both cases cleanup is not that important

functional/FunctionalTest.java
nitisht: IMO we can target to validate MinIO implementation

PR #908, opened 2020-04-15
move to simple-xml-safe

### LINKED ISSUE #905: CVE-2017-1000190
This minio client lib depends directly on org.simpleframework:simple-xml:2.7.1 which has an active CVE: https://nvd.nist.gov/vuln/detail/CVE-2017-1000190 which sounds pretty bad.

[It doesn't look like there will be a fix in simple-xml because the maintainers state that it's more of a configuration issue](https://github.com/ngallagher/simplexml/issues/18). But forks show that it could at least have a secure default configuration.

As the CVE is of 2017, it doesn't look like it may ever be resolved. I don't know if this minio client uses a secure configuration for simple-xml that mitigates this vulnerability but IMO with this CVE and no action at all of the maintainers, org.simpleframework:simple-xml is dead.

Do we have any chance that this minio client moves away from org.simpleframework:simple-xml:2.7.1?

balamurugana:
Fixes #905

### DIFF
--- a/api/src/main/java/io/minio/SelectResponseStream.java
+++ b/api/src/main/java/io/minio/SelectResponseStream.java
@@ -31,7 +31,6 @@ import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.zip.CRC32;
-import org.xmlpull.v1.XmlPullParserException;
 
 /** Select object content response stream. */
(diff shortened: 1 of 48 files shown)


balamurugana:
> LGTM, does this break any existing releases? @balamurugana

No. `simple-xml-safe` matches with same signature of `simple-xml`

kannappanr:
ping @harshavardhana

PR #909, opened 2020-04-18
fix: do not set empty content-encoding

kannappanr:
Recent versions of the SDK sets Content-Encoding to empty string when a PutObject call is made. This PR sets Content-Encoding header only if it is not null or empty.


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -970,7 +971,7 @@ private Request createRequest(
       }
     }
 
-    if (contentEncoding != null) {
+    if (!Strings.isNullOrEmpty(contentEncoding)) {
```
balamurugana, line 974: you can also filter out standard headers with empty value in `normalizeHeaders()`

PR #911, opened 2020-04-22
[WIP]Delete object/objects even if it has a Governance-type Object Lock in…

sinhaashish:
… place

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -23,6 +23,7 @@ import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Strings;
+import com.google.common.collect.FluentIterable;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Multimaps;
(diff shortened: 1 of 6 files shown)


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -23,6 +23,7 @@
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Strings;
+import com.google.common.collect.FluentIterable;
```
balamurugana, line 26: Do not use discouraged class instead use java stream.
```java
 ImmutableList<String> results =
     FluentIterable.from(database.getClientList())
         .filter(Client::isActiveInLastMonth)
         .transform(Object::toString)
         .limit(10)
         .toList();
 ```

The approximate stream equivalent is:

```java
 List<String> results =
     database.getClientList()
         .stream()
         .filter(Client::isActiveInLastMonth)
         .map(Object::toString)
         .limit(10)
         .collect(Collectors.toList());
 ```

File: api/src/main/java/io/minio/MinioClient.java [hunk 2]
```diff
@@ -1448,6 +1449,28 @@ private Response executeDelete(
     return response;
```
balamurugana, line 1449: use below method instead of directly calling `execute()`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2620,6 +2643,52 @@ public void removeObject(String bucketName, String objectName)
...
+   * @throws InvalidBucketNameException thrown to indicate invalid bucket name passed.
+   * @throws InvalidKeyException thrown to indicate missing of HMAC SHA-256 library.
+   * @throws InvalidResponseException thrown to indicate S3 service returned invalid or no error
+   *     response.
+   * @throws IOException thrown to indicate I/O error on S3 operation.
+   * @throws NoSuchAlgorithmException thrown to indicate missing of MD5 or SHA-256 digest library.
+   * @throws XmlParserException thrown to indicate XML parsing error.
+   */
+  public void removeObject(
+      String bucketName, DeleteObject deleteObject, boolean bypassGovernanceRetention)
```
balamurugana, line 2670: Many S3 APIs supported version ID. An object is a referred as `name` and `version ID` optionally,  My recommendation is to introduce `Object` class containing `name` and `versionId` as separate PR.

Its bad idea to have `DeleteObject` class. Here you just introduce `bypassGovernanceRetention` in single object delete and multi-object delete and limit this PR scope.

PR #921, opened 2020-05-04
fix: make owner field optional in Item class.

### LINKED ISSUE #920: ListObjects fails if Owner element is absent in response
The listObjects command fails if the ListBucketResult/Contents/Owner element is missing in the response. The element is according to [the amazon s3 docs](https://docs.aws.amazon.com/AmazonS3/latest/API/API_Object.html#AmazonS3-Type-Object-Owner) optional.
Furthermore, there seems to be no public accessible method to call the internal listObjectsV2 method with fetchOwner=true. Or am I mistaken?
This behaviour has been introduced with version 7.0.0.

```
org.simpleframework.xml.core.ValueRequiredException: Unable to satisfy @org.simpleframework.xml.Element(name="Owner", type=void.class, data=false, required=true) on field 'owner' private io.minio.messages.Owner io.minio.messages.Item.owner for class io.minio.messages.Item at line 1
```

balamurugana:
Fixes #920

### DIFF
--- a/api/src/main/java/io/minio/messages/Item.java
+++ b/api/src/main/java/io/minio/messages/Item.java
@@ -42,7 +42,7 @@ public class Item {
   @Element(name = "StorageClass")
   private String storageClass;
 
-  @Element(name = "Owner")
+  @Element(name = "Owner", required = false)
   private Owner owner;
 
   @Element(name = "UserMetadata", required = false)


selfenergy:
Tested this fix for #920 twith our local setup where this issue has been found. This fix resvoled the issue.

PR #932, opened 2020-05-14
add arg builder to {set,get,delete}BucketNotification APIs

balamurugana:


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -4840,14 +4840,47 @@ public String getBucketLifeCycle(GetBucketLifeCycleArgs args)
    * @throws NoSuchAlgorithmException thrown to indicate missing of MD5 or SHA-256 digest library.
    * @throws XmlParserException thrown to indicate XML parsing error.
    */
+  @Deprecated
```
anjalshireesh, line 4843: Good to add deprecation comment also with link to new method

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -4922,12 +5005,50 @@ public void setBucketNotification(
...
       throws ErrorResponseException, IllegalArgumentException, InsufficientDataException,
           InternalException, InvalidBucketNameException, InvalidKeyException,
           InvalidResponseException, IOException, NoSuchAlgorithmException, XmlParserException {
-    NotificationConfiguration notificationConfiguration = new NotificationConfiguration();
-    setBucketNotification(bucketName, notificationConfiguration);
+    setBucketNotification(
+        SetBucketNotificationArgs.builder()
+            .bucket(bucketName)
+            .config(new NotificationConfiguration())
+            .build());
```
anjalshireesh, line 5017: Might be better to call `deleteBucketNotification` here

api/src/main/java/io/minio/MinioClient.java
balamurugana: Done

api/src/main/java/io/minio/MinioClient.java
balamurugana: Done

PR #933, opened 2020-05-14
add arg builder to {set,get,delete}BucketPolicy APIs

balamurugana:


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -4295,20 +4295,53 @@ public void putObject(
...
+   * @throws XmlParserException thrown to indicate XML parsing error.
+   */
+  public String getBucketPolicy(GetBucketPolicyArgs args)
+      throws BucketPolicyTooLargeException, ErrorResponseException, IllegalArgumentException,
+          InsufficientDataException, InternalException, InvalidBucketNameException,
+          InvalidKeyException, InvalidResponseException, IOException, NoSuchAlgorithmException,
+          XmlParserException {
+    if (args == null) {
+      throw new IllegalArgumentException("null arguments");
+    }
```
anjalshireesh, line 4337: This can be replaced with `checkArgs(args);` after rebasing with latest master branch

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -4295,20 +4295,53 @@ public void putObject(
    * @throws NoSuchAlgorithmException thrown to indicate missing of MD5 or SHA-256 digest library.
    * @throws XmlParserException thrown to indicate XML parsing error.
    */
+  @Deprecated
```
anjalshireesh, line 4298: Good to add a javadoc comment with `@deprecated` also providing link to the alternate method to be used

functional/FunctionalTest.java
```diff
@@ -3798,35 +3795,20 @@ public static void getBucketPolicy_test1() throws Exception {
               + "\"Principal\":{\"AWS\":[\"*\"]},\"Resource\":[\"arn:aws:s3:::"
               + bucketName
               + "/myobject*\"],\"Sid\":\"\"}]}";
-      client.setBucketPolicy(bucketName, policy);
-      client.getBucketPolicy(bucketName);
-      mintSuccessLog("getBucketPolicy(String bucketName)", null, startTime);
+      client.setBucketPolicy(
+          SetBucketPolicyArgs.builder().bucket(bucketName).config(policy).build());
+      client.getBucketPolicy(GetBucketPolicyArgs.builder().bucket(bucketName).build());
```
anjalshireesh, line 3800: Maybe we should verify that the value returned by `getBucketPolicy` matches what was set in the previous step?

api/src/main/java/io/minio/SetBucketPolicyArgs.java
```diff
@@ -0,0 +1,50 @@
...
+
+  /** Argument builder of {@link SetBucketPolicyArgs}. */
+  public static final class Builder extends BucketArgs.Builder<Builder, SetBucketPolicyArgs> {
+    private void validateConfig(String config) {
+      if (config == null) {
+        throw new IllegalArgumentException("null policy configuration");
+      }
+    }
+
+    protected void validate(SetBucketPolicyArgs args) {
```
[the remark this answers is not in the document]
balamurugana: Done

api/src/main/java/io/minio/MinioClient.java
balamurugana: Done

api/src/main/java/io/minio/MinioClient.java
balamurugana: Done

functional/FunctionalTest.java
balamurugana: I think it is hard to check. For example below two JSON string are same.
```json
{"Version":"2012-10-17","Statement":[{"Action":["s3:GetObject"],"Effect":"Allow","Principal":{"AWS":["*"]},"Resource":["arn:aws:s3:::MYBUCKET/myobject*"],"Sid":""}]}
```

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Action": [
                "s3:GetObject"
            ],
            "Effect": "Allow",
            "Principal": {
                "AWS": [
                    "*"
                ]
            },
            "Resource": [
                "arn:aws:s3:::MYBUCKET/myobject*"
            ],
            "Sid": ""
        }
    ]
}
```
Server may return a JSON which can be formatted differently but equally same.

anjalshireesh: Should be possible by marshalling/unmarshalling from/into a java object, but probably not worth the effort here.

PR #935, opened 2020-05-14
add arg builder to {set,get,delete}DefaultRetention APIs

balamurugana:


examples/GetDefaultRetention.java
```diff
@@ -0,0 +1,45 @@
...
+public class GetDefaultRetention {
+  /** MinioClient.getDefaultRetention() example. */
+  public static void main(String[] args)
+      throws IOException, NoSuchAlgorithmException, InvalidKeyException {
+    try {
+      /* Amazon S3: */
+      MinioClient s3Client =
+          new MinioClient("https://s3.amazonaws.com", "YOUR-ACCESSKEYID", "YOUR-SECRETACCESSKEY");
+
+      ObjectLockConfiguration config =
```
sinhaashish, line 34: IMO we should have a complete example, like creating the bucket, setting the default retention and then fetching it. So that  user just have to edit the credentials and run the file.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3502,23 +3502,60 @@ public void disableVersioning(String bucketName)
...
 
   /**
    * Gets default object retention in a bucket.
    *
    * <pre>Example:{@code
    * // bucket must be created with object lock enabled.
-   * minioClient.makeBucket("my-bucketname", null, true);
+   * minioClient.makeBucket(
+   *     MakeBucketArgs.builder().bucket("my-bucketname").objectLock(true).build());
    * ObjectLockConfiguration config = minioClient.getDefaultRetention("my-bucketname");
```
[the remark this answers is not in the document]
balamurugana: Its wrong to have `makeBucket()` here. Removing now.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3539,17 +3576,56 @@ public void setDefaultRetention(String bucketName, ObjectLockConfiguration confi
...
+  }
+
+  /**
+   * Gets default object retention in a bucket.
+   *
+   * <pre>Example:{@code
+   * // bucket must be created with object lock enabled.
+   * minioClient.makeBucket(
+   *     MakeBucketArgs.builder().bucket("my-bucketname").objectLock(true).build());
+   * ObjectLockConfiguration config =
```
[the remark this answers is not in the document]
balamurugana: Its wrong to have makeBucket() here. Removing now.

docs/API.md
```diff
@@ -834,22 +836,22 @@ minioClient.setBucketPolicy("my-bucketname", policyJson);
 ```
 
 <a name="setDefaultRetention"></a>
-### setDefaultRetention(String bucketName, ObjectLockConfiguration config)
-`public void setDefaultRetention(String bucketName, ObjectLockConfiguration config)` _[[Javadoc]](http://minio.github.io/minio-java/io/minio/MinioClient.html#setDefaultRetention-java.lang.String-io.minio.messages.ObjectLockConfiguration-)_
+### setDefaultRetention(SetDefaultRetentionArgs args)
+`public void setDefaultRetention(SetDefaultRetentionArgs args)` _[[Javadoc]](http://minio.github.io/minio-java/io/minio/MinioClient.html#setDefaultRetention-java.lang.String-io.minio.messages.ObjectLockConfiguration-)_
```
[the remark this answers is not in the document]
balamurugana: Fixed

examples/GetDefaultRetention.java
balamurugana: I think we need to be specific to examples for an API. Its not possible to run any example without modifying essentials like endpoint, credentials, bucket/object names, respective variables etc and some operations are chargeable in some service providers.

We wouldn't consider examples as tests.

functional/FunctionalTest.java
```diff
@@ -3593,37 +3596,22 @@ public static void setDefaultRetention_test() throws Exception {
...
         ObjectLockConfiguration config =
             new ObjectLockConfiguration(RetentionMode.COMPLIANCE, new RetentionDurationDays(10));
-        client.setDefaultRetention(bucketName, config);
+        client.setDefaultRetention(
+            SetDefaultRetentionArgs.builder().bucket(bucketName).config(config).build());
       } finally {
         client.removeBucket(RemoveBucketArgs.builder().bucket(bucketName).build());
       }
-      mintSuccessLog("setDefaultRetention (String bucketName)", null, startTime);
+      mintSuccessLog(methodName, null, startTime);
```
anjalshireesh, line 3604: Retention mode and duration can be logged as args

functional/FunctionalTest.java
```diff
@@ -3633,66 +3621,49 @@ public static void getDefaultRetention_test() throws Exception {
...
-                  + ", got: "
-                  + config.mode());
+              "[FAILED] mode: expected: " + expectedConfig.mode() + ", got: " + config.mode());
         }
+
+        if (config.duration().unit() != expectedConfig.duration().unit()
+            || config.duration().duration() != expectedConfig.duration().duration()) {
+          throw new Exception(
+              "[FAILED] duration: " + expectedConfig.duration() + ", got: " + config.duration());
+        }
```
anjalshireesh, line 3658: This whole code seems to be same as the previous lines except for the retention mode and duration. It can be made into a method and called twice with different arguments. Something like:

```
  testGetDefaultRetentionFor(RetentionMode.COMPLIANCE, new RetentionDurationDays(10));
  testGetDefaultRetentionFor(RetentionMode.GOVERNANCE, new RetentionDurationYears(1));
```

functional/FunctionalTest.java
```diff
@@ -3633,66 +3621,49 @@ public static void getDefaultRetention_test() throws Exception {
...
+
       } finally {
         client.removeBucket(RemoveBucketArgs.builder().bucket(bucketName).build());
       }
 
-      mintSuccessLog("getDefaultRetention (String bucketName)", null, startTime);
-
+      mintSuccessLog(methodName, null, startTime);
```
anjalshireesh, line 3664: The retention mode and duration can be logged as arguments

api/src/main/java/io/minio/SetDefaultRetentionArgs.java
```diff
@@ -0,0 +1,52 @@
...
+
+  /** Argument builder of {@link SetDefaultRetentionArgs}. */
+  public static final class Builder extends BucketArgs.Builder<Builder, SetDefaultRetentionArgs> {
+    private void validateConfig(ObjectLockConfiguration config) {
+      if (config == null) {
+        throw new IllegalArgumentException("null object-lock configuration");
+      }
+    }
+
+    protected void validate(SetDefaultRetentionArgs args) {
```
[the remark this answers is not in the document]
balamurugana: done

functional/FunctionalTest.java
balamurugana: done

functional/FunctionalTest.java
balamurugana: done

functional/FunctionalTest.java
balamurugana: It wouldn't be useful because the failure may happen any one of two checks. I am not favour of mint args because stack trace is the best to find the issue.

anjalshireesh: Actually if the `try-catch` and exception handling is also moved to the new method (`testGetDefaultRetention`) then the args logging can be done.

PR #937, opened 2020-05-15
Add arg builder to listIncompleteUpload API

sinhaashish:
Add arg builder to listIncompleteUpload

### DIFF
--- a/api/src/main/java/io/minio/ListIncompleteUploadsArgs.java
+++ b/api/src/main/java/io/minio/ListIncompleteUploadsArgs.java
@@ -1,7 +1,7 @@
 package io.minio;
 
-/** Argument class of @see #listIncompleteUploads(ListIncompleteUploadsBucketArgs args). */
-public class ListIncompleteUploadsBucketArgs extends BucketArgs {
+/** Argument class of @see #listIncompleteUploads(ListIncompleteUploadsArgs args). */
+public class ListIncompleteUploadsArgs extends BucketArgs {
   private String prefix;
   private boolean recursive;
 
@@ -19,9 +19,8 @@ public class ListIncompleteUploadsBucketArgs extends BucketArgs {
     return new Builder();
   }
 
-  /** Argument builder of @see #listIncompleteUploads(ListIncompleteUploadsBucketArgs args). */
-  public static final class Builder
-      extends BucketArgs.Builder<Builder, ListIncompleteUploadsBucketArgs> {
+  /** Argument builder of @see #listIncompleteUploads(ListIncompleteUploadsArgs args). */
+  public static final class Builder extends BucketArgs.Builder<Builder, ListIncompleteUploadsArgs> {
     public Builder prefix(String prefix) {
       operations.add(args -> args.prefix = prefix);
       return this;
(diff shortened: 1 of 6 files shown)


api/src/main/java/io/minio/ListIncompleteUploadsBucketArgs.java
```diff
@@ -0,0 +1,39 @@
+package io.minio;
+
+/** Argument class of @see #listIncompleteUploads(ListIncompleteUploadsBucketArgs args). */
+public class ListIncompleteUploadsBucketArgs extends BucketArgs {
```
balamurugana, line 4: We would need to sync name with API i.e. `ListIncompleteUploadsArgs` is more appropriate.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -4450,12 +4458,36 @@ public void removeAllBucketNotification(String bucketName)
...
+   * Iterable<Result<Upload>> results =
+   *     minioClient.listIncompleteUploads(ListIncompleteUploadsBucketArgs args);
+   * for (Result<Upload> result : results) {
+   *   Upload upload = result.get();
+   *   System.out.println(upload.uploadId() + ", " + upload.objectName());
+   * }
+   * }</pre>
+   *
+   * @param args Arguments for listing incomplete uploads.
+   * @return Iterable&ltResult&ltUpload&gt&gt - Lazy iterator contains object upload information.
```
anjalshireesh, line 4514:
```suggestion
   * @return Iterable&lt;Result&lt;Upload&gt;&gt; - Lazy iterator contains object upload information.
```

api/src/main/java/io/minio/ListIncompleteUploadsArgs.java
```diff
@@ -0,0 +1,50 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.minio;
+
+/** Argument class of @see #listIncompleteUploads(ListIncompleteUploadsArgs args). */
+public class ListIncompleteUploadsArgs extends BucketArgs {
```
balamurugana, line 20: Could you add all parameters like `listObjects()`? Similarly you could check reusing/modifying existing `objectIterator` class to accommodate this listing as well.

sinhaashish: IMO this not feasible as there are two different API's expecting different  query parameters. 
List Incomplete upload : https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListMultipartUploads.html
```
GET /?uploads&Delimiter=Delimiter&EncodingType=EncodingType&KeyMarker=KeyMarker&MaxUploads=MaxUploads&Prefix=Prefix&UploadIdMarker=UploadIdMarker HTTP/1.1
Host: Bucket.s3.amazonaws.com
```

List Objects : https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html
```
GET /?list-type=2&ContinuationToken=ContinuationToken&Delimiter=Delimiter&EncodingType=EncodingType&FetchOwner=FetchOwner&MaxKeys=MaxKeys&Prefix=Prefix&StartAfter=StartAfter HTTP/1.1
Host: Bucket.s3.amazonaws.com
x-amz-request-payer: RequestPayer
```
As far as using object Iterator is concerned it implements Iterator of Item type, `(ObjectIterator implements Iterator<Result<Item>> )` while list incompleteUpload needs Iterator of type Upload `Iterator<Result<Upload>>` . Item and Upload the the different response classes for these api. 
IMO  `listObjects` and `listIncompleteObjects` being separate is correct.

balamurugana: We would need to do little restructure to have this and it can be done later sometime.

PR #944, opened 2020-05-18
 Add validation support to the builder framework 

balamurugana:


api/src/main/java/io/minio/BucketArgs.java
```diff
@@ -34,7 +34,7 @@ public String region() {
   /** Base argument builder class. */
   public abstract static class Builder<B extends Builder<B, A>, A extends BucketArgs>
       extends BaseArgs.Builder<B, A> {
-    protected void validateName(String name) {
+    private static void validateName(String name) {
```
anjalshireesh, line 37: This method need not be static

api/src/main/java/io/minio/BucketArgs.java
```diff
@@ -60,15 +60,28 @@ protected void validateName(String name) {
       }
     }
 
-    @SuppressWarnings("unchecked") // Its safe to type cast to T as T is inherited by this class
+    private static void validateRegion(String region) {
```
anjalshireesh, line 63: This method need not be static

api/src/main/java/io/minio/BucketArgs.java
```diff
@@ -60,15 +60,28 @@ protected void validateName(String name) {
...
+      if (region != null && region.isEmpty()) {
```
anjalshireesh, line 64: This condition should be:

`if (region == null || region.isEmpty())`

balamurugana: Actually `null` region is allowed. When region is `null` we need to probe or use default region. However if region is provided, it should not be empty.

PR #946, opened 2020-05-18
Implement Args builder for listObjects API

anjalshireesh:
Added the new args class `ListObjectsArgs` to be used with the
listObjects API. Example usage:

```
Iterable<Result<Item>> results = minioClient.listObjects(
  ListObjectsArgs.builder().
    bucket("my-bucketname").
    includeUserMetadata(true).
    startAfter("start-after-entry").
    prefix("my-obj").
    maxKeys(100).
    fetchOwner(true)
  );
```

Also refactored some of the existing code to reduce duplication of code:

- Extract common part of the result of both V1 and V2 API into
`ListBucketResult`, and made the two specific classes
(`ListBucketResultV1` and `ListBucketResultV2`) extend from it. The main
difference between the two is the naming of the continuation token
(`continuationToken` in V2 vs `marker` in V1). Since this is only a
naming difference, both results can be treated the same way by
abstracting out the methods related to these result fields.

- Similarly, most of the logic of the iterators for the response of V1
and V2 was also common. Extracted it out into a new abstract class
called `ObjectIterator`.

### DIFF
--- a/api/src/main/java/io/minio/ListObjectsArgs.java
+++ b/api/src/main/java/io/minio/ListObjectsArgs.java
@@ -1,0 +1,128 @@
+/*
+ * MinIO Java SDK for Amazon S3 Compatible Cloud Storage, (C) 2020 MinIO, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.minio;
+
+/** Argument class of @see #listObjects(ListObjectsArgs args). */
+public class ListObjectsArgs extends BucketArgs {
+  private String continuationToken;
+  private String delimiter;
+  private boolean fetchOwner;
+  private Integer maxKeys;
+  private String prefix;
+  private String startAfter;
+  private boolean includeUserMetadata;
+  private String marker;
+  private boolean recursive;
+  private boolean useVersion1;
+
+  public String continuationToken() {
+    return continuationToken;
+  }
+
+  public boolean includeUserMetadata() {
+    return includeUserMetadata;
+  }
+
+  public String startAfter() {
+    return startAfter;
+  }
+
+  public String prefix() {
+    return prefix;
+  }
+
+  public Integer maxKeys() {
+    return maxKeys;
+  }
+
+  public boolean fetchOwner() {
+    return fetchOwner;
+  }
+
+  public String delimiter() {
+    return delimiter;
+  }
+
+  public String marker() {
+    return marker;
+  }
+
+  public boolean recursive() {
+    return recursive;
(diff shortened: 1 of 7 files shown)


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2872,293 +2877,250 @@ public void remove() {
...
+  private Iterable<Result<Item>> listObjectsV2(ListObjectsArgs args) {
+    return new Iterable<Result<Item>>() {
+      @Override
+      public Iterator<Result<Item>> iterator() {
+        return new ObjectIterator() {
+          protected void populateResult()
+              throws InvalidKeyException, InvalidBucketNameException, IllegalArgumentException,
+                  NoSuchAlgorithmException, InsufficientDataException, XmlParserException,
+                  ErrorResponseException, InternalException, InvalidResponseException, IOException {
+            String delimiter = "/";
```
balamurugana, line 3058: We need use `delimiter` from `args.delimiter()` if it is `null` and we can default to `/`

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2872,293 +2877,250 @@ public void remove() {
...
-          private boolean completed = false;
-
-          private synchronized void populate() {
+        return new ObjectIterator() {
+          @Override
+          protected void populateResult()
+              throws InvalidKeyException, InvalidBucketNameException, IllegalArgumentException,
+                  NoSuchAlgorithmException, InsufficientDataException, XmlParserException,
+                  ErrorResponseException, InternalException, InvalidResponseException, IOException {
             String delimiter = "/";
```
balamurugana, line 3098: same as above

balamurugana:
Please update `docs/API.md` too.

api/src/main/java/io/minio/ListObjectsArgs.java
```diff
@@ -0,0 +1,128 @@
...
+/** Argument class of @see #listObjects(ListObjectsArgs args). */
+public class ListObjectsArgs extends BucketArgs {
+  private String continuationToken;
+  private String delimiter;
+  private boolean fetchOwner;
+  private Integer maxKeys;
+  private String prefix;
+  private String startAfter;
+  private boolean includeUserMetadata;
+  private String marker;
```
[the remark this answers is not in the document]
anjalshireesh: Thanks, will change to start using `startAfter` in both places.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2872,293 +2877,250 @@ public void remove() {
...
+   *
+   * <pre>Example:{@code
+   * Iterable<Result<Item>> results = minioClient.listObjects(
+   *   ListObjectsArgs.builder().
+   *   bucket("my-bucketname").
+   *   includeUserMetadata(true).
+   *   startAfter("start-after-entry").
+   *   prefix("my-obj").
+   *   maxKeys(100).
+   *   fetchOwner(true)
```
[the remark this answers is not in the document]
anjalshireesh: Will change as suggested

api/src/main/java/io/minio/MinioClient.java
anjalshireesh: I had kept the code as is and just refactored to reduce duplication. However, your suggestion seems like an improvement over the existing code - so I will change the code accordingly.

api/src/main/java/io/minio/MinioClient.java
anjalshireesh: Will change as suggested

PR #954, opened 2020-05-26
Implement Args builder for getObject API

anjalshireesh:
New Argument class: `GetObjectArgs`

Arguments supported:
 - offset (Long)
 - length (Long)
 - ssec (ServerSideEncryptionWithCustomerKey)

Since there are multiple APIs / args classes that need `ssec` as an argument, introduced an intermediate abstract class `SsecObjectArgs` that supports the `ssec` argument, and provides common logic related to it e.g. validations.

Since the return type of the methods that support downloading the object as a file is `void`, we cannot name it the same as `getObject`. So introduced a new method called `downloadObject` for downloading the object, which internally uses `getObject`. And introduced a new args class `DownloadObjectArgs` for it.

Also,
  - Enhanced the method `FunctionalTest#handleException` to support  logging of the arguments as well.
  - Moved the class `ServerSideEncryptionWithCustomerKey` outside of `ServerSideEncryption` and made it into a standalone class with the name `ServerSideEncryptionCustomerKey`

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -1626,7 +1626,7 @@ public class MinioClient {
    *
    * <pre>Example:{@code
    * ObjectStat objectStat =
-   *     minioClient.statObject("my-bucketname", "my-objectname", ssec);
+   *     minioClient.statObject("my-bucketname", "my-objectname", sse);
    * }</pre>
    *
    * @param bucketName Name of the bucket.
@@ -1647,12 +1647,13 @@ public class MinioClient {
    * @see ObjectStat
    */
   @Deprecated
-  public ObjectStat statObject(String bucketName, String objectName, ServerSideEncryption sse)
+  public ObjectStat statObject(
+      String bucketName, String objectName, ServerSideEncryptionCustomerKey ssec)
       throws ErrorResponseException, IllegalArgumentException, InsufficientDataException,
           InternalException, InvalidBucketNameException, InvalidKeyException,
           InvalidResponseException, IOException, NoSuchAlgorithmException, XmlParserException {
     return statObject(
-        StatObjectArgs.builder().bucket(bucketName).object(objectName).ssec(sse).build());
+        StatObjectArgs.builder().bucket(bucketName).object(objectName).ssec(ssec).build());
   }
 
   /**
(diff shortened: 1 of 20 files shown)


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1626,7 +1626,7 @@ public ObjectStat statObject(String bucketName, String objectName)
    *
    * <pre>Example:{@code
    * ObjectStat objectStat =
-   *     minioClient.statObject("my-bucketname", "my-objectname", ssec);
+   *     minioClient.statObject("my-bucketname", "my-objectname", sse);
```
balamurugana, line 1629: This is still `ssec`. You could revert this change.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -1713,7 +1714,7 @@ public ObjectStat statObject(StatObjectArgs args)
           InternalException, InvalidBucketNameException, InvalidKeyException,
           InvalidResponseException, IOException, NoSuchAlgorithmException, XmlParserException {
     checkArgs(args);
-    checkReadRequestSse(args.ssec());
+    args.validateSsec(baseUrl);
```
balamurugana, line 1717: Is it a good idea to have this check inside `checkArgs()`?

api/src/main/java/io/minio/SsecObjectArgs.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+package io.minio;
+
+import okhttp3.HttpUrl;
+
+public abstract class SsecObjectArgs extends ObjectArgs {
+  protected ServerSideEncryptionCustomerKey ssec;
+  public static final int TYPE_READ = 0;
+  public static final int TYPE_WRITE = 1;
```
balamurugana, line 24: Are these constants still needed?

api/src/main/java/io/minio/MinioClient.java
anjalshireesh: Will revert

api/src/main/java/io/minio/MinioClient.java
anjalshireesh: `checkArgs` accepts the base class `BaseArgs`, so the `validateSsec()` method which is defined in `SsecObjectArgs` won't be available there.

api/src/main/java/io/minio/SsecObjectArgs.java
anjalshireesh: Not required! Thanks, will remove.

PR #966, opened 2020-06-03
Rectify compose object functional test

sinhaashish:
Delete the object in case of exception. Since Compose Object doesn't work against Azure, the object remained and was not deleted. With this PR, deletion of object is done in finally and thus will always execute.

### DIFF
--- a/functional/FunctionalTest.java
+++ b/functional/FunctionalTest.java
@@ -3130,17 +3130,20 @@ public class FunctionalTest {
    * composeSources,Map &lt;String, String&gt; headerMap, ServerSideEncryption sseTarget).
    */
   public static void composeObject_test1() throws Exception {
+
+    String methodName =
+        "composeObject(String bucketName, String objectName,List<ComposeSource> composeSources, "
+            + "Map <String,String > headerMap, ServerSideEncryption sseTarget)";
     if (!mintEnv) {
-      System.out.println(
-          "Test: composeObject(String bucketName, String objectName,List<ComposeSource> composeSources, "
-              + "Map <String,String > headerMap, ServerSideEncryption sseTarget).");
+      System.out.println("Test :" + methodName);
     }
     long startTime = System.currentTimeMillis();
-
+    String mintArgs = "size: 6 MB & 6 MB ";
     try {
       String destinationObjectName = getRandomName();
       String filename1 = createFile6Mb();
       String filename2 = createFile6Mb();
+
       PutObjectOptions options = new PutObjectOptions(6 * MB, -1);
       client.putObject(bucketName, filename1, filename1, options);
       client.putObject(bucketName, filename2, filename2, options);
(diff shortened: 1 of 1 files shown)


functional/FunctionalTest.java
```diff
@@ -3115,15 +3115,18 @@ public static void composeObject_test1() throws Exception {
...
-      client.composeObject(bucketName, destinationObjectName, listSourceObjects, null, null);
-      Files.delete(Paths.get(filename1));
-      Files.delete(Paths.get(filename2));
-
-      client.removeObject(RemoveObjectArgs.builder().bucket(bucketName).object(filename1).build());
-      client.removeObject(RemoveObjectArgs.builder().bucket(bucketName).object(filename2).build());
-      client.removeObject(
-          RemoveObjectArgs.builder().bucket(bucketName).object(destinationObjectName).build());
+      try {
+        client.composeObject(bucketName, destinationObjectName, listSourceObjects, null, null);
```
kannappanr, line 3119: what will happen to the mint log here? We should not say success in this case, if there is a `not implemented` error returned we need to log it as N/A

sinhaashish: It wont say success. 
Flow is : 
**case A : If the composeObject is implemented**
1. It will create the object with name `destinationObjectName` 
2. Then it will delete the object `destinationObjectName`
3. The code in finally gets executed which deletes the source objects

**case B: If the composeObject is not  implemented**
1. The object wont be created 
2. Apart from `finally` no code below try gets called . So mint success will not be executed.
3. Finally block get executed always so the source object will deleted.

functional/FunctionalTest.java
```diff
@@ -3119,45 +3122,22 @@ public static void composeObject_test1() throws Exception {
...
+        client.removeObject(
+            RemoveObjectArgs.builder().bucket(bucketName).object(destinationObjectName).build());
+      } finally {
+        Files.delete(Paths.get(filename1));
+        Files.delete(Paths.get(filename2));
+        client.removeObject(
+            RemoveObjectArgs.builder().bucket(bucketName).object(filename1).build());
+        client.removeObject(
+            RemoveObjectArgs.builder().bucket(bucketName).object(filename2).build());
```
anjalshireesh, line 3135: Since this whole finally block is being repeated multiple times (sometimes with 2 files, sometimes with 1), it can be extracted into a method that takes an array/list of filenames as argument and deletes them from both the local filesystem and the object store. Something like:

`private void deleteFilesAndObjects(String filenames[]);`

functional/FunctionalTest.java
```diff
@@ -340,6 +340,16 @@ private static void handleException(String methodName, String args, long startTi
     throw e;
   }
 
+  private static void deleteFilesAndObjects(Map<String, List<String>> deleteList) throws Exception {
```
anjalshireesh, line 343: The argument doesn't have to be a map as you're always passing a single bucket. I suggest the following signature for this method:

`private static void deleteFilesAndObjects(String bucket, String files[])`

functional/FunctionalTest.java
```diff
@@ -3158,14 +3168,11 @@ public static void composeObject_test1() throws Exception {
...
-        Files.delete(Paths.get(filename1));
-        Files.delete(Paths.get(filename2));
-        client.removeObject(
-            RemoveObjectArgs.builder().bucket(bucketName).object(filename1).build());
-        client.removeObject(
-            RemoveObjectArgs.builder().bucket(bucketName).object(filename2).build());
+        Map<String, List<String>> deleteObjectList = new HashMap<String, List<String>>();
+        deleteObjectList.put(
+            bucketName, new ArrayList<String>(Arrays.asList(filename1, filename2)));
+        deleteFilesAndObjects(deleteObjectList);
```
anjalshireesh, line 3174: If you go with my previous suggestion, this block will become far simpler:

`deleteFilesAndObjects(bucketName, new String[]{filename1, filename2});`

functional/FunctionalTest.java
sinhaashish: Objects are related to bucket. In case we have more one bucket then `private static void deleteFilesAndObjects(String bucket, String files[])` will not be of use. So i have created a map with bucketName as key and object as list of String.

anjalshireesh: This is just a method to reduce verbosity and duplication of code. If this itself makes the code more verbose, then the purpose is defeated.

Since all places where this is called currently work with a single bucket, I don't see a reason to add the complexity of supporting multiple buckets at this point.

PR #972, opened 2020-06-07
Add arg builder support to putObject api

### LINKED ISSUE #969: Create empty folder example
Resolved issues https://github.com/minio/minio/issues/2423 and https://github.com/minio/minio-java/issues/653 suggest that it's possible to create an empty folder via the Java client, by specifying an object name with a trailing slash, and ensuring the size is 0 bytes.   

Please could we have an example in the examples folder that demonstrates this?

balamurugana:
Fixes #969


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2587,7 +2588,7 @@ public void composeObject(
         size -= src.offset();
       }
 
-      if (size < PutObjectOptions.MIN_MULTIPART_SIZE
+      if (size < ObjectWriteArgs.MIN_MULTIPART_SIZE
```
anjalshireesh, line 2591: Some of the validations here are the same as those in `PutObjectArgs`. Might be a good idea to put them in a common method and call from both places.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -7202,31 +7309,91 @@ protected void abortMultipartUpload(String bucketName, String objectName, String
...
+      String bodyContent = new String(response.body().bytes(), StandardCharsets.UTF_8);
+      bodyContent = bodyContent.trim();
+      if (!bodyContent.isEmpty()) {
+        try {
+          if (Xml.validate(ErrorResponse.class, bodyContent)) {
+            ErrorResponse errorResponse = Xml.unmarshal(ErrorResponse.class, bodyContent);
+            throw new ErrorResponseException(errorResponse, response);
+          }
+        } catch (XmlParserException e) {
+          // As it is not <Error> message, ignore this exception
```
anjalshireesh, line 7384: Maybe just log/print the exception so that it can be debugged..

api/src/main/java/io/minio/MinioClient.java
balamurugana: This will be taken care when we add arg buildet to `composeObject()`

api/src/main/java/io/minio/PutObjectArgs.java
```diff
@@ -0,0 +1,162 @@
...
+          double pSize = Math.ceil((double) objectSize / MAX_MULTIPART_COUNT);
+          pSize = Math.ceil(pSize / MIN_MULTIPART_SIZE) * MIN_MULTIPART_SIZE;
+
+          partSize = (long) pSize;
+
+          if (pSize > 0) {
+            partCount = (int) Math.ceil(objectSize / pSize);
+          } else {
+            partCount = 1;
+          }
```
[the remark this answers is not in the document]
balamurugana: In case of input stream size is unknown, there is possibility that user could pass wrong  `partSize <= 0`

api/src/main/java/io/minio/UploadObjectArgs.java
```diff
@@ -0,0 +1,112 @@
...
+ */
+
+package io.minio;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+/** Argument class of MinioClient.uploadObject(). */
+public class UploadObjectArgs extends ObjectWriteArgs {
```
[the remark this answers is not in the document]
balamurugana: done

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -5053,47 +5074,21 @@ private void putObject(
    * @throws IOException thrown to indicate I/O error on S3 operation.
    * @throws NoSuchAlgorithmException thrown to indicate missing of MD5 or SHA-256 digest library.
    * @throws XmlParserException thrown to indicate XML parsing error.
+   * @deprecated use {@link #uploadObject(UploadObjectArgs)}
    */
+  @Deprecated
   public void putObject(
```
sinhaashish, line 5080: Why is `ssec` not used ?

balamurugana:
> I dont find the `docs/API.md` file updated.

Fixed

PR #984, opened 2020-06-16
Add versioning support in listObjects API

### LINKED ISSUE #949: Send delimiter/max-keys/prefix query parameters empty for ListObjects API
Bucket policy at server-side may prevent accessing ListObjects REST API if delimiter/max-keys/prefix query parameters are sent with empty values.

Below is an example valid bucket policy
```json
{
    "Statement": [
        {
            "Action": [
                "s3:ListBucket"
            ],
            "Effect": "Allow",
            "Principal": "*",
            "Resource": [
                "arn:aws:s3:::mybucket"
            ],
            "Condition": {
                "StringEquals": {
                    "s3:prefix": [
                        ""
                    ],
                    "s3:delimiter": [
                        ""
                    ],
                    "s3:max-keys": [
                        1000
                    ]
                }
            }
        }
    ],
    "Version": "2012-10-17"
}
```

balamurugana:
Fixes #949

### DIFF
--- a/api/src/main/java/io/minio/messages/ListBucketResultV1.java
+++ b/api/src/main/java/io/minio/messages/ListBucketResultV1.java
@@ -16,8 +16,6 @@
 
 package io.minio.messages;
 
-import java.util.Collections;
-import java.util.LinkedList;
 import java.util.List;
 import org.simpleframework.xml.Element;
 import org.simpleframework.xml.ElementList;
@@ -48,7 +46,8 @@ public class ListBucketResultV1 extends ListObjectsResult {
     return nextMarker;
   }
 
-  public List<Item> contents() {
-    return Collections.unmodifiableList((contents == null) ? new LinkedList<>() : contents);
+  @Override
+  public List<Contents> contents() {
+    return emptyIfNull(contents);
   }
 }
--- a/api/src/main/java/io/minio/messages/ListBucketResultV1.java
+++ b/api/src/main/java/io/minio/messages/ListBucketResultV1.java
@@ -16,7 +16,11 @@
 
 package io.minio.messages;
 
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
 import org.simpleframework.xml.Element;
+import org.simpleframework.xml.ElementList;
 import org.simpleframework.xml.Namespace;
 import org.simpleframework.xml.Root;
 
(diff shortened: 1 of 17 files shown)


api/src/main/java/io/minio/messages/Version.java
```diff
@@ -0,0 +1,31 @@
+/*
+ * MinIO Java SDK for Amazon S3 Compatible Cloud Storage, (C) 2015 MinIO, Inc.
```
kannappanr, line 2:
```suggestion
 * MinIO Java SDK for Amazon S3 Compatible Cloud Storage, (C) 2020 MinIO, Inc.
```

api/src/main/java/io/minio/messages/ListObjectsResult.java
```diff
@@ -87,26 +72,20 @@ public boolean isTruncated() {
...
   }
 
-  /** Returns continuation token. */
-  public abstract String continuationToken();
+  public List<Item> deleteMarkers() {
+    return deleteMarkers;
+  }
 
-  public abstract String nextContinuationToken();
+  public abstract List<Item> contents();
```
anjalshireesh, line 90:
```suggestion
  protected <T extends Item> List<T> emptyIfNull(List<T> lst) {
    return Collections.unmodifiableList(MoreObjects.firstNonNull(lst, new LinkedList<T>()));
  }
  public abstract List<Item> contents();
```

I suggest adding a generic method `emptyIfNull()`, which can then be called from the subclass implementations of `contents()`. Note that the return type of the implementations will have to change to the concrete types with this. e.g. `List<Contents>` and `List<Version>`

api/src/main/java/io/minio/messages/ListVersionsResult.java
```diff
@@ -0,0 +1,78 @@
...
+
+  public String versionIdMarker() {
+    return versionIdMarker;
+  }
+
+  public String nextVersionIdMarker() {
+    return nextVersionIdMarker;
+  }
+
+  public List<Item> contents() {
```
anjalshireesh, line 69: Add the `@Override` annotation, and change the implementation to call `emptyIfNull()` if you accept my suggestion in `ListObjectsResult`

api/src/main/java/io/minio/messages/ListBucketResultV1.java
```diff
@@ -26,22 +30,25 @@
...
   }
 
-  /** Returns next continuation token. */
-  @Override
-  public String nextContinuationToken() {
+  public String nextMarker() {
     return nextMarker;
   }
+
+  public List<Item> contents() {
```
anjalshireesh, line 51: Add the `@Override` annotation, and change the implementation to call `emptyIfNull()` if you accept my suggestion in `ListObjectsResult`

api/src/main/java/io/minio/messages/ListBucketResultV2.java
```diff
@@ -27,22 +31,44 @@
...
   }
 
   /** Returns next continuation token. */
-  @Override
   public String nextContinuationToken() {
     return nextContinuationToken;
   }
+
+  /** Returns List of Items. */
+  public List<Item> contents() {
```
anjalshireesh, line 71: Add the `@Override` annotation, and change the implementation to call `emptyIfNull()` if you accept my suggestion in `ListObjectsResult`

PR #989, opened 2020-06-24
Add base argument class for conditionally reading objects.

balamurugana:
`GetObject`, `HeadObject` and `CopyObject` S3 API support conditions
on reading objects. This patch introduces common base class
`ObjectConditionalReadArgs` for `GetObjectArgs`, `StatObjectArgs`,
`CopySource` and `ComposeSource` to avoid code duplication.

### DIFF
--- a/api/src/main/java/io/minio/ComposeSource.java
+++ b/api/src/main/java/io/minio/ComposeSource.java
@@ -20,84 +20,78 @@ package io.minio;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Multimaps;
-import java.time.ZonedDateTime;
-import okhttp3.HttpUrl;
+import io.minio.errors.InternalException;
 
-/** Source information to compose object. */
-public class ComposeSource extends ObjectVersionArgs {
-  private Long offset;
-  private Long length;
-  private long objectSize;
-  private String matchETag;
-  private String notMatchETag;
-  private ZonedDateTime modifiedSince;
-  private ZonedDateTime unmodifiedSince;
-  private ServerSideEncryptionCustomerKey ssec;
+/** A source object defintion for {@link ComposeObjectArgs}. */
+public class ComposeSource extends ObjectConditionalReadArgs {
+  private Long objectSize = null;
+  private Multimap<String, String> headers = null;
 
-  private Multimap<String, String> headers;
+  protected ComposeSource() {}
 
-  public Long offset() {
-    return offset;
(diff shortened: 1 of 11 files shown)


api/src/main/java/io/minio/StatObjectArgs.java
```diff
@@ -17,11 +17,28 @@
 package io.minio;
 
 /** Argument class of MinioClient.statObject(). */
-public class StatObjectArgs extends ObjectReadArgs {
+public class StatObjectArgs extends ObjectConditionalReadArgs {
+  protected StatObjectArgs() {}
+
+  public StatObjectArgs(ObjectConditionalReadArgs args) {
```
sinhaashish, line 23: As the name suggest  `StatObjectArgs` is an builder type then why is the constructor used. IMO it should be renamed to `StatObject` similar to `ComposeObject`.

api/src/main/java/io/minio/ComposeSource.java
```diff
@@ -115,90 +109,36 @@ public void buildHeaders(long objectSize, String etag) throws IllegalArgumentExc
...
-    }
-
-    public Builder length(long length) {
-      validateNullOrPositive(length, "length");
-      operations.add(args -> args.length = length);
-      return this;
+  public long objectSize() throws InternalException {
+    if (this.objectSize == null) {
+      throw new InternalException(
+          "buildHeaders(long objectSize, String etag) must be called prior this method invocation");
```
kannappanr, line 122:
```suggestion
          "buildHeaders(long objectSize, String etag) must be called prior to the invocation of this method");
```

api/src/main/java/io/minio/ObjectConditionalReadArgs.java
```diff
@@ -0,0 +1,143 @@
+/*
+ * MinIO Java SDK for Amazon S3 Compatible Cloud Storage,
+ * (C) 2019 MinIO, Inc.
```
kannappanr, line 3:
```suggestion
 * (C) 2020 MinIO, Inc.
```

api/src/main/java/io/minio/ComposeSource.java
balamurugana: added `to`

api/src/main/java/io/minio/ObjectConditionalReadArgs.java
balamurugana: This file is there since 2019, now added 2020

balamurugana: done

PR #998, opened 2020-06-30
Return higher level object GetObjectResponse for getObject API

balamurugana:


api/src/main/java/io/minio/GetObjectResponse.java
```diff
@@ -0,0 +1,103 @@
...
+ * limitations under the License.
+ */
+
+package io.minio;
+
+import java.io.IOException;
+import java.io.InputStream;
+import okhttp3.Headers;
+
+public class GetObjectResponse extends InputStream {
```
anjalshireesh, line 23: Why not extend from `GenericResponse` and have just one instance variable for `InputStream body` ?

balamurugana: `InputStream body` by `public InputStream body()` is already there.

We want `GetObjectResponse` needs to be an `InputStream` and have functionality of `GenericResponse`. As `GenericResponse` and `InputStream` are abstract classes, I cannot inherit both the classes, but inherit one class (`InputStream`) and have instance of another class (`GenericResponse`); if we do other way around; `InputStream` capability is lost.

anjalshireesh: @balamurugana the `InputStream` capability won't be lost. We can get the `InputStream` by calling `body()` from the instance of `GetObjectResponse` to get the input stream.

balamurugana: The idea here is to `GetObjectResponse` be an `InputStream` which is most common usage and it doesn't break backward compatibility.

anjalshireesh: Backward compatibility with the previous return type of the method - understood :+1: 
In that case, may I suggest to remove all the wrapper methods that call the same method on the `response` object, and expect the client to call `response()` method to get the object of `BaseResponse` and then use it to get further details? We can also consider making the `GenericResponse` non-abstract and using it directly.

api/src/main/java/io/minio/GetObjectResponse.java
```diff
@@ -0,0 +1,103 @@
...
+  private BaseResponse response;
+  private InputStream body;
+
+  public GetObjectResponse(
+      Headers headers, String bucket, String region, String object, InputStream body) {
+    this.response = new BaseResponse(headers, bucket, region, object);
+    this.body = body;
```
anjalshireesh, line 30: `GetObjectResponse` can extend from `BufferedInputStream` and then call `super(body)` here inside the constructor.
With this, The class would authentically become an `InputStream` and would not need to store the `body` variable and write wrapper methods for it.

balamurugana: 1. We would need to keep proxying methods in `GenericResponse` here to comply with other Response classes; else `GetObjectResponse` looks different. The idea is `GetObjectResponse` is same as `GenericResponse` but also an `InputStream`.

2. Yes. Making `GenericResponse` as non-abstract class.

3. I am going to use `FilterInputStream` which is more appropriate than `BufferedInputStream`

PR #1003, opened 2020-07-02
Add arg builder to listBuckets API

balamurugana:


sinhaashish:
Can we add an example file ?

balamurugana:
> Can we add an example file ?

Not required. Both `listBuckets()` and `listBuckets(ListBucketsArgs)` are available but `listBuckets(ListBucketsArgs)` is available only for advanced users.

PR #694, opened 2018-07-25
Server side encryption

aarushiarya:
Fixes #694. Added server side encryptions support for SSE_S3 , SSE_C and SSE_KMS with examples and functional tests.

PR #1021, opened 2020-07-16
fix: use separate config dir for running local minio server in functional tests

harshavardhana:


functional/FunctionalTest.java
```diff
@@ -3558,9 +3559,8 @@ public static void main(String[] args) throws Exception {
     } else {
       kmsKeyName = System.getenv("MINIO_JAVA_TEST_KMS_KEY_NAME");
```
kannappanr, line 3560: Can we remove this environment variable?

harshavardhana: I don't why it is useful though @kannappanr

harshavardhana:
> After discussion at local side, I see the PR is trying to fix an environment issue at user side when running `./gradlew runFunctionalTest`. It is user responsibility to run it in a clean environment with downloaded minio server.
> 
> However `./gradlew runFunctionalTest -Pendpoint=ENDPOINT -PaccessKey=ACCESSKEY -PsecretKey=SECRETKEY -Pregion=REGION` is the recommended way to run by user.

That is not correct @balamurugana we are running MinIO server internally here in the functional test there are many things that can come in the way. I don't like the way it is done right now it creates too much trouble for anyone to use it's hard to even figure out that is the correct command line. 

I made one more change now to use --config-dir instead of REGION_NAME - such that functional test MinIO server doesn't use anyone's home directory.

functional/FunctionalTest.java
```diff
@@ -3558,9 +3559,8 @@ public static void main(String[] args) throws Exception {
...
       if (kmsKeyName == null) {
-        kmsKeyName = System.getenv("MINT_KEY_ID");
+        kmsKeyName = System.getenv("MINT_KMS_KEY_NAME");
       }
-      sqsArn = System.getenv("MINIO_JAVA_TEST_SQS_ARN");
```
balamurugana, line 3563: Any reason for removing this?

functional/FunctionalTest.java
```diff
@@ -3558,9 +3559,8 @@ public static void main(String[] args) throws Exception {
     } else {
       kmsKeyName = System.getenv("MINIO_JAVA_TEST_KMS_KEY_NAME");
       if (kmsKeyName == null) {
-        kmsKeyName = System.getenv("MINT_KEY_ID");
```
balamurugana, line 3561: Are you sure the env is changed from `"MINT_KEY_ID"` to `"MINT_KMS_KEY_NAME"`?

harshavardhana: there is no MINT_KEY_ID env in mint @balamurugana - I just added this to make sure we keep it the same as MINIO_JAVA_TEST_KMS_KEY_NAME

functional/FunctionalTest.java
harshavardhana: There is no purpose for this @balamurugana - can you explain why this exists?

functional/FunctionalTest.java
balamurugana: `MINT_KEY_ID` was introduced in https://github.com/minio/minio-java/pull/694 and available since last two years. Do you say it was a mistake in the PR?

functional/FunctionalTest.java
balamurugana: It is needed when running functional tests manually `./gradlew runFunctionalTest -Pendpoint=ENDPOINT -PaccessKey=ACCESSKEY -PsecretKey=SECRETKEY -Pregion=REGION`; otherwise tests depend on `sqsArn` wouldn't run. There is no code without its purpose.

harshavardhana: Are you expecting users to know all of this run functional tests?

functional/FunctionalTest.java
harshavardhana: Yes it is not at all present in mint, it has been long removed.

functional/FunctionalTest.java
balamurugana: Yes. In fact users never run functional tests, but developers do. If a developer is indent to testing `minio-java` himself apart from CI, he has an option to do so. Also note that, `Mint` runs functional tests to test minio server by this way.

harshavardhana: Yes but this is completely wrong,  one way is env, and another way is command line. It looks clunky way to test things. Why have this option in the first place?

PR #1025, opened 2020-07-22
fix canonical header computation for multi-spaced header values

balamurugana:


kannappanr:
@balamurugana Please take a look at the CI failure

api/src/main/java/io/minio/Signer.java
```diff
@@ -142,7 +143,14 @@ private void setCanonicalHeaders() {
     for (String name : headers.names()) {
       String signedHeader = name.toLowerCase(Locale.US);
       if (!IGNORED_HEADERS.contains(signedHeader)) {
-        this.canonicalHeaders.put(signedHeader, headers.get(name));
+        this.canonicalHeaders.put(
```
harshavardhana, line 146: I am still unable to understand why this type of code is needed for canonicalHeaders, is the okhttp manipulating headers after they are set? 

No other language SDK needs this type of handling - HTTP header values are opaque and should be kept opaque without any escaping etc unless specified with the protocol/server. 

I wonder if this is the correct fix here.

PR #1028, opened 2020-07-24
Handle error code generically for any S3 service.

balamurugana:


api/src/main/java/io/minio/ErrorCode.java
```diff
@@ -32,6 +32,98 @@
...
+      "XMinioAdminBucketReplicationDisabled",
+      "Replication specified but disk usage crawl is disabled on MinIO server"),
+  X_MINIO_ADMIN_CONFIG_BAD_JSON(
+      "XMinioAdminConfigBadJSON", "JSON configuration provided is of incorrect format"),
+  X_MINIO_ADMIN_CONFIG_DUPLICATE_KEYS(
+      "XMinioAdminConfigDuplicateKeys",
+      "JSON configuration provided has objects with duplicate keys"),
+  X_MINIO_ADMIN_CONFIG_NO_QUORUM(
+      "XMinioAdminConfigNoQuorum", "Configuration update failed because server quorum was not met"),
+  X_MINIO_ADMIN_CONFIG_TOO_LARGE(
```
harshavardhana, line 56: @balamurugana, since this can change especially XMinioAdmin, is not part of S3 API why do we need to add this? 

Why not make non-translatable errors become ErrorResponse instead? instead of this type of hardcoded value - especially the error description for our Admin API errors can change anytime.

balamurugana: Better idea is to remove `ErrorCode.java` itself. It would ease out for any S3 services.

harshavardhana: Yeah agreed @balamurugana

harshavardhana:
PTAL @sinhaashish

PR #1033, opened 2020-08-09
add xml object representation of replication configuration

balamurugana:


functional/FunctionalTest.java
```diff
@@ -3750,6 +3705,7 @@ public static void main(String[] args) throws Exception {
         dataFile6Mb = Paths.get(dataDir, "datafile-6-MB");
       }
     }
+    replicationSrcBucket = System.getenv("MINIO_JAVA_TEST_REPLICATION_SRC_BUCKET");
```
poornas, line 3708: is this test really serving the purpose as it will always be skipped without remote target set up on MinIO server? Other than that, PR looks good to me.

balamurugana: Yes. It is needed. That's how I tested the implementation. Do you think any better way to do in automated tests?

poornas: the `MINIO_JAVA_TEST_REPLICATION_ROLE` is validated on MinIO server while setting the replication configuration - since java SDK does not implement the admin API to get the arn while setting up remote targets, I do not see how this test will get executed

balamurugana: You would need to run these functional tests by exporting 

```bash
$ export MINIO_JAVA_TEST_REPLICATION_SRC_BUCKET=<SRC-BUCKET>
$ export MINIO_JAVA_TEST_REPLICATION_ROLE=<REPLICATON-ROLE>
$ export MINIO_JAVA_TEST_REPLICATION_BUCKET_ARN=<REPLICATION-BUCKET-ARN>
$`./gradlew runFunctionalTest -Pendpoint=ENDPOINT -PaccessKey=ACCESSKEY -PsecretKey=SECRETKEY -Pregion=REGION`
```

Here you would need to have source/destination buckets and replication role created before running the test.

poornas: my comment was wr.t mint tests.

PR #1035, opened 2020-08-10
Added Javadoc bucket versioning setup restrictions

indalaterre:
Hello, I think that javadoc for the `enableVersioning` should contains information about the setup restrictions.
I copy paste the required setup from the [documentation](https://docs.minio.io/docs/minio-bucket-versioning-guide.html) and put the link to it.

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -3818,6 +3818,10 @@ public class MinioClient {
   /**
    * Enables object versioning feature in a bucket.
    *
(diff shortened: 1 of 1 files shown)


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3818,6 +3818,10 @@ public void enableVersioning(String bucketName)
   /**
    * Enables object versioning feature in a bucket.
    *
+   * Versioning feature is available only for erasure coded and distributed erasure coded setups
```
balamurugana, line 3821: `minio-java` is meant to any S3 service not limiting to MinIO server. This doc is not relevant here.

PR #1036, opened 2020-08-10
add assume-role credential provider

### LINKED ISSUE #817: Minio AssumeRole using SDK
Is there support for the AssumeRole funtion from either the Minio Java SDK or the AWS Java SDK?

I found no documentation for this in the Minio SDK

I have attempted to use the AWS Java SDK to make an assumeRole call to a Minio sever.
It consistently fails with:
com.amazonaws.services.securitytoken.model.AWSSecurityTokenServiceException: Generating temporary credentials not allowed for this request. (Service: AWSSecurityTokenService; Status Code: 403; Error Code: AccessDenied; Request ID: ########

I am hoping that there SDK support and perhaps an example is available. Thanks.

## Context
We are attempting to get credentials in a java app that a python client can subsequently use to put and get large files.

balamurugana:
Fixes #817

### DIFF
--- a/api/src/main/java/io/minio/credentials/AssumeRoleProvider.java
+++ b/api/src/main/java/io/minio/credentials/AssumeRoleProvider.java
@@ -1,0 +1,149 @@
+/*
+ * MinIO Java SDK for Amazon S3 Compatible Cloud Storage, (C) 2020 MinIO, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.minio.credentials;
+
+import io.minio.Digest;
+import io.minio.Signer;
+import io.minio.Time;
+import io.minio.Xml;
+import io.minio.errors.XmlParserException;
+import java.io.IOException;
+import java.security.InvalidKeyException;
+import java.security.NoSuchAlgorithmException;
+import java.time.ZonedDateTime;
+import java.util.Objects;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import okhttp3.HttpUrl;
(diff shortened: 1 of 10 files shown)


aalex21:
Can we support the "ExternalId" parameter as part of AssumeRole Query Params ?
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html

This will limit one minio client to be associated with one externalId !!

balamurugana:
> Can we support the "ExternalId" parameter as part of AssumeRole Query Params ?
> https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html
> 
> This will limit one minio client to be associated with one externalId !!

Sure. Added now.

api/src/main/java/io/minio/credentials/AssumeRoleProvider.java
```diff
@@ -0,0 +1,172 @@
...
+public class AssumeRoleProvider implements Provider {
+  public static final int DEFAULT_DURATION_SECONDS = (int) TimeUnit.HOURS.toSeconds(1);
+  private final HttpUrl stsEndpoint;
+  private final String accessKey;
+  private final String secretKey;
+  private final String region;
+  private final OkHttpClient httpClient;
+  private final String contentSha256;
+  private final Request request;
+  private Credentials credentials;
```
anjalshireesh, line 57: The fields `stsEndpoint`, `httpClient` and `credentials` seem to be common with `WebIdentityClientGrantsProvider`. We can consider adding an abstract base class that has a constructor that takes these arguments and call it from the constructors of both `WebIdentityClientGrantsProvider` and `AssumeRoleProvider`

api/src/main/java/io/minio/credentials/AssumeRoleProvider.java
```diff
@@ -0,0 +1,172 @@
...
+
+    String data = urlBuilder.build().encodedQuery();
+    this.contentSha256 = Digest.sha256Hash(data);
+    this.request =
+        new Request.Builder()
+            .url(this.stsEndpoint)
+            .method(
+                "POST",
+                RequestBody.create(MediaType.parse("application/x-www-form-urlencoded"), data))
+            .build();
```
anjalshireesh, line 122: The request building logic is inside the `fetch()` method in the `WebIdentityClientGrantsProvider` class. Can we try to have a consistent pattern in both? I see a small potential for an abstract base class that provides the `fetch()` logic as well, forcing the concrete sub-classes to build the request, which seems to be the main difference.

anjalshireesh (dismissed):
LGTM

sinhaashish (approved):
Tested & LGTM

anjalshireesh (approved):
LGTM

PR #1038, opened 2020-08-14
add IAM AWS credential provider

balamurugana:


api/src/main/java/io/minio/credentials/IamAwsProvider.java
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  @Override
+  public synchronized Credentials fetch() {
+    if (credentials != null && !credentials.isExpired()) {
+      return credentials;
+    }
+
+    HttpUrl url = this.customEndpoint;
+    if (getProperty("AWS_WEB_IDENTITY_TOKEN_FILE") != null) {
```
anjalshireesh, line 113: To making this call twice with same argument, we can save the output of this`getProperty` in a variable `tokenFile` and pass it as argument to `getToken`

api/src/main/java/io/minio/credentials/IamAwsProvider.java
```diff
@@ -0,0 +1,218 @@
...
+          new WebIdentityProvider(
+              () -> getToken(),
+              url.toString(),
+              null,
+              null,
+              getProperty("AWS_ROLE_ARN"),
+              getProperty("AWS_ROLE_SESSION_NAME"),
+              httpClient);
+      credentials = provider.fetch();
+      return credentials;
```
anjalshireesh, line 133: Can be extracted out into a method `fetchUsingAwsTokenFile`

api/src/main/java/io/minio/credentials/IamAwsProvider.java
```diff
@@ -0,0 +1,218 @@
...
+      if (url == null) {
+        url = HttpUrl.parse("http://169.254.169.254/latest/meta-data/iam/security-credentials/");
+      } else {
+        url =
+            new HttpUrl.Builder()
+                .scheme(url.scheme())
+                .host(url.host())
+                .addPathSegments("latest/meta-data/iam/security-credentials/")
+                .build();
+      }
```
anjalshireesh, line 160: Can be extracted into a method `getAwsCredentialsUrl`

api/src/main/java/io/minio/credentials/IamAwsProvider.java
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        roleNames = response.body().string().split("\\R");
+      } catch (IOException e) {
+        throw new IllegalStateException("Unable to parse response", e);
+      }
+
+      if (roleNames.length == 0) {
+        throw new IllegalStateException("No IAM roles attached to EC2 service " + url);
+      }
```
anjalshireesh, line 178: Can be extracted into a method `getAwsRoleName`

api/src/main/java/io/minio/credentials/IamAwsProvider.java
```diff
@@ -0,0 +1,218 @@
...
+    this.mapper = new ObjectMapper();
+    this.mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+    this.mapper.configure(MapperFeature.ACCEPT_CASE_INSENSITIVE_PROPERTIES, true);
+  }
+
+  private void checkLoopbackHost(HttpUrl url) {
+    try {
+      for (InetAddress addr : InetAddress.getAllByName(url.host())) {
+        if (!addr.isLoopbackAddress()) {
+          throw new IllegalArgumentException(url.host() + " is not loopback only host");
```
anjalshireesh, line 71: Not sure if `IllegalArgumentException` is the right choice for all the exceptions in this class.

anjalshireesh: Sorry I wrote this comment at the wrong place with the wrong exception name! I meant to say that I'm not sure if `IllegalStateException` makes sense for most of the exceptions in the class..

anjalshireesh (approved):
LGTM

sinhaashish (approved):
LGTM

PR #1039, opened 2020-08-15
add ldap identity credential provider

balamurugana:


api/src/main/java/io/minio/credentials/LdapIdentityProvider.java
```diff
@@ -0,0 +1,94 @@
...
+  protected Request getRequest() {
+    return this.request;
+  }
+
+  @Override
+  protected Credentials parseResponse(Response response) throws XmlParserException, IOException {
+    AssumeRoleWithLdapIdentityResponse result =
+        Xml.unmarshal(AssumeRoleWithLdapIdentityResponse.class, response.body().charStream());
+    return result.credentials();
+  }
```
anjalshireesh, line 80: Looking at how this code of parsing response and returning credentials is almost same in three other providers, here's an idea:

1) Introduce an interface called `HasCredentials` that has a method called `credentials()` and make all these response classes (`AssumeRoleResponse, ClientGrantsResponse, WebIdentityResponse and AssumeRoleWithLdapIdentityResponse`) implement it

2) Add an abstract method in `AssumeRoleBaseProvider` called `getResponseClass()` that just returns the response class and implement it in all its concrete subclasses. In case of this class (`LdapIdentityProvider`), it will return `AssumeRoleWithLdapIdentityResponse.class`

3) Make `AssumeRoleBaseProvider#parseResponse` concrete and implement it as follows:

```
protected Credentials parseResponse(Response response) throws XmlParserException, IOException {
    HasCredentials result =
        Xml.unmarshal(getResponseClass(), response.body().charStream());
    return result.credentials();
  }
```

4) Remove all the overridden versions of `parseResponse` from the subclasses.

anjalshireesh (approved):
LGTM

sinhaashish (approved):
Not able to test as no LDAP credentials. Code looks good to me.

PR #1042, opened 2020-08-17
add MinioClient credential provider from config file

balamurugana:


api/src/main/java/io/minio/credentials/MinioClientConfigProvider.java
```diff
@@ -0,0 +1,117 @@
...
+        throw new IllegalStateException(
+            "Access key does not exist in alias " + alias + " in MinioClient configuration file");
+      }
+
+      if (secretKey == null) {
+        throw new IllegalStateException(
+            "Secret key does not exist in alias " + alias + " in MinioClient configuration file");
+      }
+
+      return new Credentials(accessKey, secretKey, null, null);
```
kannappanr, line 104: you will have to handle `bucket` vs `Path` style access also. In that case, the format has changed recently, so both of them have to be supported

balamurugana: @kannappanr 
Can you explain what is  bucket vs Path style access? Also can you provide new format change?

kannappanr: with the new format
```
"default": {
                        "url": "http://localhost:9000",
                        "accessKey": "minioadmin",
                        "secretKey": "minioadmin",
                        "api": "s3v4",
                        "path": "auto"
                },
```
Here path has two possible values `auto` and `off`, `off` means it is bucket style access. If the endpoint is AWS, you still have to do bucket style access

kannappanr: with old format, 
```
 "s3": {
                        "url": "https://s3.amazonaws.com",
                        "accessKey": "YOUR-ACCESS-KEY-HERE",
                        "secretKey": "YOUR-SECRET-KEY-HERE",
                        "api": "S3v4",
                        "lookup": "dns"
                }

```
`lookup` will have one of three values `dns, path, auto` `dns` means it is bucket style access, in the case of auto, if the endpoint is AWS, we should do bucket style access

kannappanr (approved):
LGTM

nitisht (approved):
LGTM

PR #1043, opened 2020-08-17
add chained credential provider

balamurugana:


api/src/main/java/io/minio/credentials/ChainProvider.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+package io.minio.credentials;
+
+import java.util.Arrays;
+import java.util.List;
+import javax.annotation.Nonnull;
+
+/** Chained credential provider work with list of credential providers. */
+public class ChainProvider implements Provider {
```
anjalshireesh, line 24: `ChainedProvider` might sound better

api/src/main/java/io/minio/credentials/ChainProvider.java
```diff
@@ -0,0 +1,79 @@
...
+        credentials = currentProvider.fetch();
+      } catch (IllegalStateException e) {
+        // Ignore and fallback to iteration.
+      }
+    }
+
+    if (credentials != null) {
+      synchronized (this) {
+        this.credentials = credentials;
+      }
```
anjalshireesh, line 52: Not sure if synchronizing this block helps. Making the `fetch` method itself synchronized might be better.

api/src/main/java/io/minio/credentials/ChainProvider.java
```diff
@@ -0,0 +1,79 @@
...
+
+      return this.credentials;
+    }
+
+    for (Provider provider : providers) {
+      try {
+        credentials = provider.fetch();
```
anjalshireesh, line 59: Might want to skip doing this for `currentProvider` as that has already happened in previous lines.

api/src/main/java/io/minio/credentials/ChainProvider.java
balamurugana: done

api/src/main/java/io/minio/credentials/ChainProvider.java
balamurugana: If entire method is locked in execution, `fetch()` becomes sequential execution, that leads to API execution sequential.

api/src/main/java/io/minio/credentials/ChainProvider.java
balamurugana: Its OK to try one more time in a loop.

api/src/main/java/io/minio/credentials/ChainProvider.java
anjalshireesh: Can you please help me understand how the current synchronization works? Maybe with an example use case of concurrent requests.

sinhaashish (approved):
LGTM

PR #1065, opened 2020-10-01
Add object representation of LifecycleConfiguration with API rename.

balamurugana:


sinhaashish:
@balamurugana  why are we renaming  `getBucketLifeCycle` ->`getBucketLifecycle`. This is  just a case change , but the users will have to update the code. I think this can be a part of any major version change.

balamurugana:
> @balamurugana why are we renaming `getBucketLifeCycle` ->`getBucketLifecycle`. This is just a case change , but the users will have to update the code. I think this can be a part of any major version change.

@sinhaashish
1. `lifecycle` is one word as per S3 specification, but dictionary says `life-cycle` is correct. I am changing the name according to the spec.
2. We have already bumped up major version in the master

api/src/main/java/io/minio/messages/LifecycleRule.java
```diff
@@ -0,0 +1,112 @@
...
+  private NoncurrentVersionTransition noncurrentVersionTransition;
+
+  @Element(name = "Status")
+  private Status status;
+
+  @Element(name = "Transition", required = false)
+  private Transition transition;
+
+  /** Constructs new server-side encryption configuration rule. */
+  public LifecycleRule(
```
anjalshireesh, line 53: Instead of a constructor like this, it might be better to introduce a builder-like mechanism to build the rule. e.g. instead of 

```
new LifecycleRule(
  Status.ENABLED,
  null,
  new Expiration((ZonedDateTime) null, 365, null),
  new RuleFilter("logs/"),
  "rule2",
  null,
  null,
  null));
```

Something like this:

```
LifecycleRule.builder().
  id("rule2").
  status(Status.ENABLED).
  ruleFilter("logs/").
  expiration(new Expiration((ZonedDateTime) null, 365, null)).
  build()
```

functional/FunctionalTest.java
```diff
@@ -2742,17 +2745,25 @@ public static void deleteBucketPolicy() throws Exception {
...
+            null,
+            new Expiration((ZonedDateTime) null, 365, null),
+            new RuleFilter("logs/"),
+            "rule2",
+            null,
+            null,
+            null));
+    LifecycleConfiguration config = new LifecycleConfiguration(rules);
+    client.setBucketLifecycle(
+        SetBucketLifecycleArgs.builder().bucket(bucketName).config(config).build());
```
anjalshireesh, line 2762: Maybe we should call `getBucketLifecycle` also and verify that what was set is same as config passed in `setBucketLifecycle`

balamurugana: I added into `getBucketLifecycle()` test

api/src/main/java/io/minio/messages/LifecycleRule.java
balamurugana: Builder is surely helpful for users. However this constructor still serves to prevent invalid values in XML response from server (We see many weirdly implemented S3 servers these days).

The builder is an improvement for many such messages. I think it can be added on user requests.

sinhaashish (approved):

functional/FunctionalTest.java
```diff
@@ -2846,9 +2846,43 @@ public static void getBucketLifecycle() throws Exception {
...
+          throw new Exception(
+              "rule.expiration().days(): expected: 365, got: " + rule.expiration().days());
+        }
+        if (rule.filter() == null) {
+          throw new Exception("rule.filter(): expected: <non-null>, got: <null>");
+        }
+        if (!"logs/".equals(rule.filter().prefix())) {
+          throw new Exception(
+              "rule.filter().prefix(): expected: logs/, got: " + rule.filter().prefix());
+        }
```
anjalshireesh, line 2882: It might be better to introduce an `equals` method on the `LifecycleConfiguration` class and use that to confirm that the config that is received on getting the lifecycle is equal to the config that was passed when setting it.

balamurugana: As we would need to make a next major release, we can add this in a separate PR after the release.

anjalshireesh (approved):
LGTM

PR #1076, opened 2020-10-17
Refactor iterator in removeObjects() API.

### LINKED ISSUE #1075: Issue of removeObjects API only delete maximum of 2000 objects
When applying batch delete by calling the `removeObjects` API, it only deletes 2000 objects per API call. It shouldn't work that way by design, since there is nothing mentioned in the official document about the limit of `removeObjects` API. 

Several versions (6.0.7, 7.0.0, 7.1.3) have been tested, and all have the same issue. Suspect it's caused by a bug in the iterable implementation of the API.

balamurugana:
Fixes #1075


sinhaashish (approved):

kannappanr (approved):
LGTM

PR #1088, opened 2020-10-28
Fix version info and UserAgent string

vinu:
- Take the os name from system property
- Try to find the correct MANIFEST and load the version info


balamurugana (changes requested):
send the PR to `master` branch

vinu:
> send the PR to `master` branch
Sure - my bad

PR #1089, opened 2020-10-28
Refactor MinioProperties class and fix os name in user-agent

vinu:
- Take the os name from system properties
- Try to find the correct MANIFEST and load the version info (when the jar is used with multiple libraries)

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -194,7 +194,7 @@ public class MinioClient {
   private static final int DEFAULT_EXPIRY_TIME = 7 * 24 * 3600;
   private static final String DEFAULT_USER_AGENT =
       "MinIO ("
-          + System.getProperty("os.arch")
+          + System.getProperty("os.name")
           + "; "
           + System.getProperty("os.arch")
           + ") minio-java/"
--- a/api/src/main/java/io/minio/MinioProperties.java
+++ b/api/src/main/java/io/minio/MinioProperties.java
@@ -57,16 +57,14 @@ enum MinioProperties {
       while (resources.hasMoreElements()) {
         try (InputStream is = resources.nextElement().openStream()) {
           Manifest manifest = new Manifest(is);
-          String implementationTitleValue =
-              manifest.getMainAttributes().getValue("Implementation-Title");
-          if (implementationTitleValue != null && implementationTitleValue.equals("minio")) {
+          if ("minio".equals(manifest.getMainAttributes().getValue("Implementation-Title"))) {
             version.set(manifest.getMainAttributes().getValue("Implementation-Version"));
             return;
           }
         }
       }
     } catch (IOException e) {
-      LOGGER.log(Level.SEVERE, "IOException occured", e);
+      LOGGER.log(Level.SEVERE, "IOException occurred", e);
       version.set("unknown");
     }
   }
--- a/api/src/main/java/io/minio/MinioProperties.java
+++ b/api/src/main/java/io/minio/MinioProperties.java
@@ -43,55 +43,46 @@ enum MinioProperties {
 
   public String getVersion() {
     String result = version.get();
-    if (result == null) {
-      synchronized (INSTANCE) {
-        if (version.get() == null) {
-          try {
-            ClassLoader classLoader = getClass().getClassLoader();
-            setMinioClientJavaVersion(classLoader);
-            setDevelopmentVersion();
-          } catch (IOException e) {
-            LOGGER.log(Level.SEVERE, "IOException occured", e);
-            version.set("unknown");
-          }
-          result = version.get();
-        }
-      }
+    if (result != null) {
+      return result;
     }
-    return result;
+    setVersion();
+    return version.get();
   }
 
-  private void setDevelopmentVersion() {
-    if (version.get() == null) {
+  private synchronized void setVersion() {
+    ClassLoader classLoader = getClass().getClassLoader();
+    if (classLoader == null) {
       version.set("dev");
+      return;
     }
-  }
 
-  private void setMinioClientJavaVersion(ClassLoader classLoader) throws IOException {
-    if (classLoader != null) {
+    try {
       Enumeration<URL> resources = classLoader.getResources("META-INF/MANIFEST.MF");
-      boolean minioManifestFound = false;
-      boolean minioVersionFound = false;
       while (resources.hasMoreElements()) {
         try (InputStream is = resources.nextElement().openStream()) {
           Manifest manifest = new Manifest(is);
+          boolean minioManifestFound;
           for (Map.Entry<Object, Object> entry : manifest.getMainAttributes().entrySet()) {
-            if (entry.getKey().toString().equals(META_INF_ATTRIB_IMPLEMENTATION_TITLE)
-                && entry.getValue().toString().equals(META_INF_ATTRIB_IMPLEMENTATION_TITLE_VALUE)) {
-              minioManifestFound = true;
-            }
-            if (minioManifestFound
-                && entry.getKey().toString().equals(META_INF_ATTRIB_IMPLEMENTATION_VERSION)) {
-              version.set(entry.getValue().toString());
-              minioVersionFound = true;
-              break;
+            String key = entry.getKey().toString();
+            String value = entry.getValue().toString();
+            minioManifestFound =
+                key.equals(META_INF_ATTRIB_IMPLEMENTATION_TITLE)
+                    && value.equals(META_INF_ATTRIB_IMPLEMENTATION_TITLE_VALUE);
+
+            if (minioManifestFound) {
+              version.set(
+                  manifest.getMainAttributes().getValue(META_INF_ATTRIB_IMPLEMENTATION_VERSION));
+              return;
             }
           }
         }
-        if (minioVersionFound) {
-          break;
-        }
       }
+    } catch (IOException e) {
+      LOGGER.log(Level.SEVERE, "IOException occured", e);
+      version.set("unknown");
+      return;
     }
+    version.set("dev");
   }
 }
--- a/api/src/main/java/io/minio/MinioProperties.java
+++ b/api/src/main/java/io/minio/MinioProperties.java
@@ -20,7 +20,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
 import java.util.Enumeration;
-import java.util.Map;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.jar.Manifest;
 import java.util.logging.Level;
@@ -31,13 +30,6 @@ enum MinioProperties {
   INSTANCE;
 
   private static final Logger LOGGER = Logger.getLogger(MinioProperties.class.getName());
-
-  // These attributes are checked from the manifests in classpath.
-  private static final String META_INF_ATTRIB_IMPLEMENTATION_TITLE = "Implementation-Title";
-  private static final String META_INF_ATTRIB_IMPLEMENTATION_VERSION = "Implementation-Version";
-
-  // this is set from gradle
-  private static final String META_INF_ATTRIB_IMPLEMENTATION_TITLE_VALUE = "minio";
 
   private final AtomicReference<String> version = new AtomicReference<>(null);
 
@@ -51,9 +43,12 @@ enum MinioProperties {
   }
 
   private synchronized void setVersion() {
+    if (version.get() != null) {
+      return;
+    }
+    version.set("dev");
     ClassLoader classLoader = getClass().getClassLoader();
     if (classLoader == null) {
-      version.set("dev");
       return;
     }
 
@@ -62,27 +57,17 @@ enum MinioProperties {
       while (resources.hasMoreElements()) {
         try (InputStream is = resources.nextElement().openStream()) {
           Manifest manifest = new Manifest(is);
-          boolean minioManifestFound;
-          for (Map.Entry<Object, Object> entry : manifest.getMainAttributes().entrySet()) {
-            String key = entry.getKey().toString();
-            String value = entry.getValue().toString();
-            minioManifestFound =
-                key.equals(META_INF_ATTRIB_IMPLEMENTATION_TITLE)
-                    && value.equals(META_INF_ATTRIB_IMPLEMENTATION_TITLE_VALUE);
-
-            if (minioManifestFound) {
-              version.set(
-                  manifest.getMainAttributes().getValue(META_INF_ATTRIB_IMPLEMENTATION_VERSION));
-              return;
-            }
+          String implementationTitleValue =
+              manifest.getMainAttributes().getValue("Implementation-Title");
+          if (implementationTitleValue != null && implementationTitleValue.equals("minio")) {
+            version.set(manifest.getMainAttributes().getValue("Implementation-Version"));
+            return;
           }
         }
       }
     } catch (IOException e) {
       LOGGER.log(Level.SEVERE, "IOException occured", e);
       version.set("unknown");
-      return;
     }
-    version.set("dev");
   }
 }
--- a/api/src/main/java/io/minio/MinioProperties.java
+++ b/api/src/main/java/io/minio/MinioProperties.java
@@ -17,10 +17,11 @@
 package io.minio;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.URL;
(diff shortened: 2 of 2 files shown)


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -194,7 +194,7 @@
   private static final int DEFAULT_EXPIRY_TIME = 7 * 24 * 3600;
   private static final String DEFAULT_USER_AGENT =
       "MinIO ("
-          + System.getProperty("os.arch")
+          + System.getProperty("os.name")
```
balamurugana, line 197: :+1:

api/src/main/java/io/minio/MinioProperties.java
```diff
@@ -62,14 +70,27 @@ private void setDevelopmentVersion() {
...
-        for (Object k : manifest.getMainAttributes().keySet()) {
-          String versionString = "Implementation-Version";
-          if (k.toString().equals(versionString)) {
-            version.set(manifest.getMainAttributes().getValue((Attributes.Name) k));
+        try (InputStream is = resources.nextElement().openStream()) {
+          Manifest manifest = new Manifest(is);
+          for (Map.Entry<Object, Object> entry : manifest.getMainAttributes().entrySet()) {
+            if (entry.getKey().toString().equals(META_INF_ATTRIB_IMPLEMENTATION_TITLE)
+                && entry.getValue().toString().equals(META_INF_ATTRIB_IMPLEMENTATION_TITLE_VALUE)) {
+              minioManifestFound = true;
```
balamurugana, line 81: Why do we need this? or what exactly the problem in current approach?

vinu: When minio is included with other libraries - via maven/gradle - this manifest resources will list all the manifests from the classpath - We specifically need the manifest from the minio jar - so we can get the version number correctly.

An example is added here https://github.com/vinu/minio-check (using minio with spring boot)
I added `7.1.4` of minio library, so I was expecting the User-agent to have this version, but it was like this `User-Agent: MinIO (x86_64; x86_64) minio-java/1.7.30`

This is because the current implementation lists all manifests from the classpath, and just takes the `Implementation-Version` so the version info is overridden by some other manifest.

balamurugana: :+1: 

Just refactor the class to simplify everything like below
```diff
diff --git a/api/src/main/java/io/minio/MinioProperties.java b/api/src/main/java/io/minio/MinioProperties.java
index ad6f4c9b..a46430b4 100644
--- a/api/src/main/java/io/minio/MinioProperties.java
+++ b/api/src/main/java/io/minio/MinioProperties.java
@@ -20,7 +20,6 @@ import java.io.IOException;
 import java.net.URL;
 import java.util.Enumeration;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.jar.Attributes;
 import java.util.jar.Manifest;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -28,49 +27,49 @@ import java.util.logging.Logger;
 /** Identifies and stores version information of minio-java package at run time. */
 enum MinioProperties {
   INSTANCE;
-
   private static final Logger LOGGER = Logger.getLogger(MinioProperties.class.getName());
-
   private final AtomicReference<String> version = new AtomicReference<>(null);
 
-  public String getVersion() {
-    String result = version.get();
-    if (result == null) {
-      synchronized (INSTANCE) {
-        if (version.get() == null) {
-          try {
-            ClassLoader classLoader = getClass().getClassLoader();
-            setMinioClientJavaVersion(classLoader);
-            setDevelopmentVersion();
-          } catch (IOException e) {
-            LOGGER.log(Level.SEVERE, "IOException occured", e);
-            version.set("unknown");
-          }
-          result = version.get();
-        }
-      }
+  private synchronized void setVersion() {
+    if (version.get() != null) {
+      return;
     }
-    return result;
-  }
 
-  private void setDevelopmentVersion() {
-    if (version.get() == null) {
+    ClassLoader classLoader = getClass().getClassLoader();
+    if (classLoader == null) {
       version.set("dev");
+      return;
     }
-  }
 
-  private void setMinioClientJavaVersion(ClassLoader classLoader) throws IOException {
-    if (classLoader != null) {
-      Enumeration<URL> resources = classLoader.getResources("META-INF/MANIFEST.MF");
-      while (resources.hasMoreElements()) {
-        Manifest manifest = new Manifest(resources.nextElement().openStream());
-        for (Object k : manifest.getMainAttributes().keySet()) {
-          String versionString = "Implementation-Version";
-          if (k.toString().equals(versionString)) {
-            version.set(manifest.getMainAttributes().getValue((Attributes.Name) k));
+    Enumeration<URL> resources = classLoader.getResources("META-INF/MANIFEST.MF");
+    while (resources.hasMoreElements()) {
+      try (InputStream is = resources.nextElement().openStream()) {
+        Manifest manifest = new Manifest(is);
+        boolean minioManifestFound = false;
+        for (Map.Entry<Object, Object> entry : manifest.getMainAttributes().entrySet()) {
+          String key = entry.getKey().toString();
+          String value = entry.getValue().toString();
+          if (minioManifestFound && key.equals("Implementation-Version")) {
+            version.set(value);
+            break;
           }
+
+          minioManifestFound = value.equals("minio") && key.equals("Implementation-Title");
         }
+      } catch (IOException e) {
+        LOGGER.log(Level.SEVERE, "IOException occured", e);
+        version.set("unknown");
       }
     }
   }
+
+  public String getVersion() {
+    String result = version.get();
+    if (result != null) {
+      return result;
+    }
+
+    setVersion();
+    return version.get();
+  }
 }
```

balamurugana: :+1: 

Simplify everything like below
```diff
diff --git a/api/src/main/java/io/minio/MinioProperties.java b/api/src/main/java/io/minio/MinioProperties.java
index ad6f4c9b..72bc9e26 100644
--- a/api/src/main/java/io/minio/MinioProperties.java
+++ b/api/src/main/java/io/minio/MinioProperties.java
@@ -20,7 +20,6 @@ import java.io.IOException;
 import java.net.URL;
 import java.util.Enumeration;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.jar.Attributes;
 import java.util.jar.Manifest;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -28,49 +27,52 @@ import java.util.logging.Logger;
 /** Identifies and stores version information of minio-java package at run time. */
 enum MinioProperties {
   INSTANCE;
-
   private static final Logger LOGGER = Logger.getLogger(MinioProperties.class.getName());
-
   private final AtomicReference<String> version = new AtomicReference<>(null);
 
-  public String getVersion() {
-    String result = version.get();
-    if (result == null) {
-      synchronized (INSTANCE) {
-        if (version.get() == null) {
-          try {
-            ClassLoader classLoader = getClass().getClassLoader();
-            setMinioClientJavaVersion(classLoader);
-            setDevelopmentVersion();
-          } catch (IOException e) {
-            LOGGER.log(Level.SEVERE, "IOException occured", e);
-            version.set("unknown");
-          }
-          result = version.get();
-        }
-      }
+  private synchronized void setVersion() {
+    if (version.get() != null) {
+      return;
     }
-    return result;
-  }
 
-  private void setDevelopmentVersion() {
-    if (version.get() == null) {
+    ClassLoader classLoader = getClass().getClassLoader();
+    if (classLoader == null) {
       version.set("dev");
+      return;
     }
-  }
 
-  private void setMinioClientJavaVersion(ClassLoader classLoader) throws IOException {
-    if (classLoader != null) {
-      Enumeration<URL> resources = classLoader.getResources("META-INF/MANIFEST.MF");
-      while (resources.hasMoreElements()) {
-        Manifest manifest = new Manifest(resources.nextElement().openStream());
-        for (Object k : manifest.getMainAttributes().keySet()) {
-          String versionString = "Implementation-Version";
-          if (k.toString().equals(versionString)) {
-            version.set(manifest.getMainAttributes().getValue((Attributes.Name) k));
+    Enumeration<URL> resources = classLoader.getResources("META-INF/MANIFEST.MF");
+    while (resources.hasMoreElements()) {
+      try (InputStream is = resources.nextElement().openStream()) {
+        Manifest manifest = new Manifest(is);
+        boolean minioManifestFound = false;
+        for (Map.Entry<Object, Object> entry : manifest.getMainAttributes().entrySet()) {
+          String key = entry.getKey().toString();
+          String value = entry.getValue().toString();
+          if (minioManifestFound && key.equals("Implementation-Version")) {
+            version.set(value);
+            return;
           }
+
+          minioManifestFound = value.equals("minio") && key.equals("Implementation-Title");
         }
+      } catch (IOException e) {
+        LOGGER.log(Level.SEVERE, "IOException occured", e);
+        version.set("unknown");
+        return;
       }
     }
+
+    version.set("dev");
+  }
+
+  public String getVersion() {
+    String result = version.get();
+    if (result != null) {
+      return result;
+    }
+
+    setVersion();
+    return version.get();
   }
 }
```

balamurugana: More simpler and error free
```diff
diff --git a/api/src/main/java/io/minio/MinioProperties.java b/api/src/main/java/io/minio/MinioProperties.java
index ad6f4c9b..d10eedda 100644
--- a/api/src/main/java/io/minio/MinioProperties.java
+++ b/api/src/main/java/io/minio/MinioProperties.java
@@ -17,10 +17,11 @@
 package io.minio;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.URL;
 import java.util.Enumeration;
+import java.util.Map;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.jar.Attributes;
 import java.util.jar.Manifest;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -28,49 +29,53 @@ import java.util.logging.Logger;
 /** Identifies and stores version information of minio-java package at run time. */
 enum MinioProperties {
   INSTANCE;
-
   private static final Logger LOGGER = Logger.getLogger(MinioProperties.class.getName());
-
   private final AtomicReference<String> version = new AtomicReference<>(null);
 
-  public String getVersion() {
-    String result = version.get();
-    if (result == null) {
-      synchronized (INSTANCE) {
-        if (version.get() == null) {
-          try {
-            ClassLoader classLoader = getClass().getClassLoader();
-            setMinioClientJavaVersion(classLoader);
-            setDevelopmentVersion();
-          } catch (IOException e) {
-            LOGGER.log(Level.SEVERE, "IOException occured", e);
-            version.set("unknown");
-          }
-          result = version.get();
-        }
-      }
+  private synchronized void setVersion() {
+    if (version.get() != null) {
+      return;
     }
-    return result;
-  }
 
-  private void setDevelopmentVersion() {
-    if (version.get() == null) {
-      version.set("dev");
+    version.set("dev");
+
+    ClassLoader classLoader = getClass().getClassLoader();
+    if (classLoader == null) {
+      return;
     }
-  }
 
-  private void setMinioClientJavaVersion(ClassLoader classLoader) throws IOException {
-    if (classLoader != null) {
+    try {
       Enumeration<URL> resources = classLoader.getResources("META-INF/MANIFEST.MF");
       while (resources.hasMoreElements()) {
-        Manifest manifest = new Manifest(resources.nextElement().openStream());
-        for (Object k : manifest.getMainAttributes().keySet()) {
-          String versionString = "Implementation-Version";
-          if (k.toString().equals(versionString)) {
-            version.set(manifest.getMainAttributes().getValue((Attributes.Name) k));
+        try (InputStream is = resources.nextElement().openStream()) {
+          Manifest manifest = new Manifest(is);
+          boolean minioManifestFound = false;
+          for (Map.Entry<Object, Object> entry : manifest.getMainAttributes().entrySet()) {
+            String key = entry.getKey().toString();
+            String value = entry.getValue().toString();
+            if (minioManifestFound && key.equals("Implementation-Version")) {
+              version.set(value);
+              return;
+            }
+
+            minioManifestFound = value.equals("minio") && key.equals("Implementation-Title");
           }
         }
       }
+    } catch (IOException e) {
+      LOGGER.log(Level.SEVERE, "IOException occured", e);
+      version.set("unknown");
+      return;
+    }
+  }
+
+  public String getVersion() {
+    String result = version.get();
+    if (result != null) {
+      return result;
     }
+
+    setVersion();
+    return version.get();
   }
 }
```

vinu: Please check the latest push, hope that cleans things :)

api/src/main/java/io/minio/MinioProperties.java
```diff
@@ -31,46 +32,57 @@
 
   private static final Logger LOGGER = Logger.getLogger(MinioProperties.class.getName());
 
+  // These attributes are checked from the manifests in classpath.
+  private static final String META_INF_ATTRIB_IMPLEMENTATION_TITLE = "Implementation-Title";
+  private static final String META_INF_ATTRIB_IMPLEMENTATION_VERSION = "Implementation-Version";
+
+  // this is set from gradle
+  private static final String META_INF_ATTRIB_IMPLEMENTATION_TITLE_VALUE = "minio";
```
balamurugana, line 40: Use these values directly than having constants.

api/src/main/java/io/minio/MinioProperties.java
```diff
@@ -31,46 +32,57 @@
...
+      return result;
     }
-    return result;
+    setVersion();
+    return version.get();
   }
 
-  private void setDevelopmentVersion() {
-    if (version.get() == null) {
+  private synchronized void setVersion() {
```
balamurugana, line 53: Do no-op if version is already set in another thread.

```java
if (version.get() != null) {
   return;
}
```

api/src/main/java/io/minio/MinioProperties.java
```diff
@@ -31,46 +32,57 @@
...
+    ClassLoader classLoader = getClass().getClassLoader();
+    if (classLoader == null) {
       version.set("dev");
```
balamurugana, line 56: Set `dev` as default value i.e. move above this line and remove setting it later.

api/src/main/java/io/minio/MinioProperties.java
```diff
@@ -31,46 +32,57 @@
...
-          String versionString = "Implementation-Version";
-          if (k.toString().equals(versionString)) {
-            version.set(manifest.getMainAttributes().getValue((Attributes.Name) k));
+        try (InputStream is = resources.nextElement().openStream()) {
+          Manifest manifest = new Manifest(is);
+          boolean minioManifestFound;
+          for (Map.Entry<Object, Object> entry : manifest.getMainAttributes().entrySet()) {
+            String key = entry.getKey().toString();
+            String value = entry.getValue().toString();
+            minioManifestFound =
```
balamurugana, line 69: This is not going to work. You are resetting the value on every iteration.

balamurugana (changes requested):

api/src/main/java/io/minio/MinioProperties.java
balamurugana: Actually we need only below in the `for`
```java
          if (entry.getKey().toString().equals("Implementation-Title") && entry.getValue().toString().equals("minio")) {
            version.set(manifest.getMainAttributes().getValue("Implementation-Version"));
            return;
          }
```

vinu: Got rid of the for, we can directly pull in the values from the attributes and check

api/src/main/java/io/minio/MinioProperties.java
```diff
@@ -35,42 +35,39 @@
...
-        Manifest manifest = new Manifest(resources.nextElement().openStream());
-        for (Object k : manifest.getMainAttributes().keySet()) {
-          String versionString = "Implementation-Version";
-          if (k.toString().equals(versionString)) {
-            version.set(manifest.getMainAttributes().getValue((Attributes.Name) k));
+        try (InputStream is = resources.nextElement().openStream()) {
+          Manifest manifest = new Manifest(is);
+          String implementationTitleValue =
+              manifest.getMainAttributes().getValue("Implementation-Title");
+          if (implementationTitleValue != null && implementationTitleValue.equals("minio")) {
```
balamurugana, line 62: Just reduce to `if ("minio".equals(manifest.getMainAttributes().getValue("Implementation-Title")))`

vinu: 👍  fixed

balamurugana (approved):
LGTM

kannappanr (approved):
LGTM

PR #1100, opened 2020-11-06
fix client grants provider example

kanagarajkm:
- Updated grant_type to client_credentials
- Removed optional session policy

### DIFF
--- a/examples/MinioClientWithClientGrantsProvider.java
+++ b/examples/MinioClientWithClientGrantsProvider.java
@@ -38,17 +38,15 @@ public class MinioClientWithClientGrantsProvider {
   static Jwt getJwt(
       @Nonnull String clientId,
       @Nonnull String clientSecret,
-      @Nonnull String idpClientId,
       @Nonnull String idpEndpoint) {
     Objects.requireNonNull(clientId, "Client id must not be null");
     Objects.requireNonNull(clientSecret, "ClientSecret must not be null");
 
     RequestBody requestBody =
         new FormBody.Builder()
-            .add("username", clientId)
-            .add("password", clientSecret)
-            .add("grant_type", "password")
-            .add("client_id", idpClientId)
+            .add("client_id", clientId)
+            .add("client_secret", clientSecret)
+            .add("grant_type", "client_credentials")
             .build();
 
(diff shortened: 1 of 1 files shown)


nitisht (approved):

examples/MinioClientWithClientGrantsProvider.java
```diff
@@ -72,42 +70,20 @@ public static void main(String[] args) throws Exception {
...
     // Client-ID to fetch JWT.
-    String clientId = "USER-ID";
+    String clientId = "CLIENT-ID";
 
     // Client secret to fetch JWT.
-    String clientSecret = "PASSWORD";
-
-    // Client-ID of MinIO service on IDP.
-    String idpClientId = "MINIO-CLIENT-ID";
+    String clientSecret = "CLIENT-SECRET";
```
balamurugana, line 76: The client ID/secret are account information of IDP service. Either you leave it as is or change `User-ID` to `Username`

examples/MinioClientWithClientGrantsProvider.java
```diff
@@ -72,42 +70,20 @@ public static void main(String[] args) throws Exception {
...
 
     // STS endpoint usually point to MinIO server.
     String stsEndpoint = "http://STS-HOST:STS-PORT/";
 
-    // Policy to Credentials.
-    String policy =
```
balamurugana, line 87: Why do we remove this? Doesn't minio server support it?

examples/MinioClientWithClientGrantsProvider.java
kanagarajkm: Since they are passed as `client_id` and `client_secret` to IDP, I can name them as `IDP-CLIENT-ID` and `IDP-CLIENT-SECRET`. Calling it as USER-ID will create confusion.

examples/MinioClientWithClientGrantsProvider.java
kanagarajkm: Inline policy is supported and its optional in the API. I can bring it back.

harshavardhana:
Closing this PR in favor of #1101 due to urgency.

PR #1102, opened 2020-11-07
Add support to empty prefix value in XML messages

balamurugana:


api/src/main/java/io/minio/messages/PrefixConverter.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.simpleframework.xml.convert.Converter;
+import org.simpleframework.xml.stream.InputNode;
+import org.simpleframework.xml.stream.OutputNode;
+
+/**
+ * XML converter class for prefix due to SimpleXML limitation in converting empty tag
+ * <Prefix></Prefix> to empty string.
+ */
+public class PrefixConverter implements Converter<String> {
```
anjalshireesh, line 27: Since the logic is not very specific to `prefix` as such and can be useful wherever such a functionality is required (de-serialize empty tag to empty string), maybe the name of this class can be something more generic? Say `StringConverter` ?

functional/FunctionalTest.java
```diff
@@ -2799,7 +2789,17 @@ public static void setBucketLifecycle() throws Exception {
...
+            bucketName,
+            new LifecycleRule(
+                Status.ENABLED,
+                null,
+                new Expiration((ZonedDateTime) null, 365, null),
+                new RuleFilter("logs/"),
+                "rule2",
+                null,
+                null,
+                null));
```
anjalshireesh, line 2802: Since the exact same code is being used at multiple places to create this lifecycle rule, maybe it can be defined as a method or a static final class level variable?

anjalshireesh (approved):
LGTM, couple of minor suggestions

sinhaashish (approved):

api/src/main/java/io/minio/messages/PrefixConverter.java
balamurugana: Currently `prefix` is only supported empty. Surely change when I find another case.

functional/FunctionalTest.java
balamurugana: Yep. I will take care in next refactoring functional tests.

PR #1142, opened 2020-12-24
Honor extra headers argument in getPresignedObjectUrl API.

### LINKED ISSUE #1140: Headers are not included in signature calculated for PresignedUrl
When requesting a PresignedUrl with 
```java
Map<String,String> metadatas = Map.of("x-amz-meta-anything","any-value");

GetPresignedObjectUrlArgs args = GetPresignedObjectUrlArgs.builder()
                                                          .method(method)
                                                          .bucket(bucketName)
                                                          .object(objectUUID)
                                                          .expiry(expiry)
                                                          .extraHeaders(metadatas)
                                                          .build();

String presignedObjectUrl = minioClient.getPresignedObjectUrl(args);
 ```
 the url generated does not take care of extraHeaders, they should be added to the signature calculation and then to the x-amz-signedheaders. (only extraQueryParams are added to the presignedURL)
 
 And I see no way to add any extraHeaders to the signature calculation, the queryParam x-amz-signedheaders keep it's default value: "host". 
 
 This way, anyone using the presignedUrl can put any value in headers (for example metadata headers : x-amz-meta-*) without breaking the signature. And we cannot force a header (with a specific value) to be added to the request using the presignedUrl.
 
 There should be a way to include extraHeaders in the calculation of the signature and then in the header x-amz-signedheaders.
 
 By the way, the GetPresignedObjectUrlArgs object accepts a `extraHeaders` field from the `BaseArgs` super class but doesn't use it. This field is quite misleading then.

 Using AWS S3 Sdk, it's a bit more complicated than using the minio-java client but I can get the intended behaviour: 
 
 ```java
Map<String,String> metadatas = Map.of("x-amz-meta-anything","any-value");
PutObjectRequest objectRequest = PutObjectRequest.builder()
                                                  .bucket(bucketName)
                                                  .key(objectUUID)
                                                  .metadata(extraMetadatas)
                                                  .build();
PutObjectPresignRequest presignRequest = PutObjectPresignRequest.builder()
                                                                 .signatureDuration(Duration.ofMinutes(10))
                                                                 .putObjectRequest(objectRequest)
                                                                 .build();


PresignedPutObjectRequest presignedRequest = s3Presigner.presignPutObject(presignRequest);
System.out.println("signed headers : " + presignedRequest.signedHeaders());
s3Presigner.close();
System.out.println(presignedRequest.url().toString());
  ```
  
  Here are the results from those 2 snippets: 
  
  first one using minioclient: 
```
http://localhost:9000/afnic/68799667-f12b-43c8-b3a8-0072e0e1b069?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=phoenix%2F20201223%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20201223T124706Z&X-Amz-Expires=30&X-Amz-SignedHeaders=host&X-Amz-Signature=75f9e3a1c4c75404b7f49cfc0157f588a2a186a54a29af66e2074633453e330c
```

second one using aws s3 sdk:
```
signed headers : {x-amz-meta-anything=[any-value], host=[localhost:9000]}

http://localhost:9000/afnic/68799667-f12b-43c8-b3a8-0072e0e1b069?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Date=20201223T124706Z&X-Amz-SignedHeaders=host%3Bx-amz-meta-anything&X-Amz-Expires=600&X-Amz-Credential=phoenix%2F20201223%2Feu-central-1%2Fs3%2Faws4_request&X-Amz-Signature=9c203457adfd759b4b2a3f046f9a1ccf0d275afa1af16422510fa2758088275c
```
the main visible difference beeing the value of the queryParam `X-Amz-SignedHeaders` and of course the process of signature calculation which allow to send an error if the x-amz-meta-anything header is not added with value "any-value" while using the presignedUrl

ps: using the extraQueryParams with minioClient as shown [here](https://github.com/minio/minio-java/blob/master/examples/PresignedPutObject.java) does not solve the problem or I'm misunderstanding it.

balamurugana:
It is user responsibility to provide respective signed headers and values when
executing returned pre-signed URL.

Fixes #1140

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -1545,22 +1545,24 @@ public class MinioClient {
           XmlParserException, ServerException {
     checkArgs(args);
 
-    byte[] body = null;
-    if (args.method() == Method.PUT || args.method() == Method.POST) {
-      body = EMPTY_BODY;
-    }
+    byte[] body = (args.method() == Method.PUT || args.method() == Method.POST) ? EMPTY_BODY : null;
 
     Multimap<String, String> queryParams = newMultimap(args.extraQueryParams());
     if (args.versionId() != null) queryParams.put("versionId", args.versionId());
 
     String region = getRegion(args.bucket(), args.region());
     HttpUrl url = buildUrl(args.method(), args.bucket(), args.object(), region, queryParams);
-    if (provider == null) {
-      return url.toString();
-    }
+    if (provider == null) return url.toString();
 
     Credentials creds = provider.fetch();
-    Request request = createRequest(url, args.method(), null, body, 0, creds);
+    Request request =
+        createRequest(
+            url,
+            args.method(),
+            args.extraHeaders() == null ? null : httpHeaders(args.extraHeaders()),
+            body,
+            0,
+            creds);
     url = Signer.presignV4(request, region, creds.accessKey(), creds.secretKey(), args.expiry());
     return url.toString();
   }
--- a/api/src/main/java/io/minio/Signer.java
+++ b/api/src/main/java/io/minio/Signer.java
@@ -76,6 +76,15 @@ public class Signer {
     IGNORED_HEADERS.add("user-agent");
   }
 
+  private static final Set<String> PRESIGN_IGNORED_HEADERS = new HashSet<>();
+
+  static {
+    PRESIGN_IGNORED_HEADERS.addAll(IGNORED_HEADERS);
+    PRESIGN_IGNORED_HEADERS.add("content-md5");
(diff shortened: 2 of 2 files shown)


BigUstad (approved):
LGTM

ebozduman (approved):
LGTM

ebozduman:
I see some issues with the existing example script, `examples/PresignedPutObject.java` while I was testing this fix, but it looks like the issue is not related to this fix. I'll confirm that the issue is a legit problem and file and issue for it if confirmed. I saw similar issue with `examples/PresignedPutObject.java` when I add parameters, `extraHeaders()` and `extraQueryParams()`.

PR #1156, opened 2021-02-02
Add preload data support for putObject() API.

balamurugana:
When this flag is enabled, entire part/object data is loaded into memory to
enable connection retry on network failure in the middle of upload.

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -76,6 +76,7 @@ import io.minio.messages.Tags;
 import io.minio.messages.VersioningConfiguration;
 import io.minio.org.apache.commons.validator.routines.InetAddressValidator;
 import java.io.BufferedInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
@@ -626,9 +627,8 @@ public class MinioClient {
     traceBuilder.append("\n");
 
     OkHttpClient httpClient = this.httpClient;
-    if (method == Method.PUT || method == Method.POST) {
-      // Issue #924: disable connection retry for PUT and POST methods. Its safe to do
-      // retry for other methods.
+    if (!(body instanceof byte[]) && (method == Method.PUT || method == Method.POST)) {
+      // Issue #924: disable connection retry for PUT and POST methods for other than byte array.
       httpClient = this.httpClient.newBuilder().retryOnConnectionFailure(false).build();
     }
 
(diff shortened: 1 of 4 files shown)


harshavardhana (approved):

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3807,6 +3815,25 @@ private long getAvailableSize(Object data, long expectedReadSize)
...
+      long bytesToRead = size - totalBytesRead;
+      if (bytesToRead > buf.length) bytesToRead = buf.length;
+      int bytesRead = -1;
+      if (data instanceof RandomAccessFile) {
+        bytesRead = ((RandomAccessFile) data).read(buf, 0, (int) bytesToRead);
+      } else if (data instanceof BufferedInputStream) {
+        bytesRead = ((BufferedInputStream) data).read(buf, 0, (int) bytesToRead);
+      }
+      if (bytesRead < 0) throw new IOException("EOF on reading data");
+      buffer.write(buf, 0, bytesRead);
```
anjalshireesh, line 3832: If the offset is always passed as `0`, will it not keep overwriting the data written in the previous iteration?

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -3807,6 +3815,25 @@ private long getAvailableSize(Object data, long expectedReadSize)
     return totalBytesRead;
   }
 
+  private void fillData(ByteArrayOutputStream buffer, Object data, long size) throws IOException {
```
anjalshireesh, line 3818: Since this looks like a generic method to fill a given output stream, can it be moved to some utility class outside `MinioClient`?

balamurugana: It is not generic method, it is specifically designed and used to `MinioClient`

api/src/main/java/io/minio/MinioClient.java
balamurugana: No

api/src/main/java/io/minio/MinioClient.java
anjalshireesh: We can still have something like `MinioClientUtil`. The `MinioClient.java` is too huge for comfort (5000+ lines)

balamurugana: This is different issue, needs to be addressed in different PR.

anjalshireesh: Your call. IMHO it's a good idea to start straight away, especially for new methods that are being introduced.

anjalshireesh (approved):

PR #1161, opened 2021-02-17
Use generic logic to find Item is a DeleteMarker

harshavardhana:


api/src/main/java/io/minio/messages/Item.java
```diff
@@ -128,6 +129,6 @@ public boolean isDir() {
 
   /** Returns whether this item is a delete marker or not. */
   public boolean isDeleteMarker() {
-    return (etag == null && size == 0 && storageClass == null && versionId != null);
+    return (etag == null && size == 0 && versionId != null);
```
balamurugana, line 132: Like AWS SDK Java, you could just do
```java
return this instanceof DeleteMarker;
```

api/src/main/java/io/minio/messages/Item.java
```diff
@@ -46,7 +46,8 @@
   private long size; // except DeleteMarker
 
   @Element(name = "StorageClass", required = false)
-  private String storageClass; // except DeleteMarker
+  // except DeleteMarker, but will be populated in case of MinIO server
+  private String storageClass;
```
balamurugana, line 50: You could add inline comment like
```java
private String storageClass; // except DeleteMarker, not in case of MinIO server.
```

balamurugana (approved):

kannappanr (approved):
LGTM

PR #1170, opened 2021-03-15
Add support for continuation token for listObjectsV2

lem21h:
Add support for listObjetsV2 to provide and use continuationToken to continue listing of the selected bucket.
Also match the user metadata keys returned from statObject metod with listObjects (remove the "protocol" X-AWS-Meta prefix).

### DIFF
--- a/api/src/main/java/io/minio/ListObjectsArgs.java
+++ b/api/src/main/java/io/minio/ListObjectsArgs.java
@@ -38,7 +38,12 @@ public class ListObjectsArgs extends BucketArgs {
       return "";
     }
 
-    return (delimiter.isEmpty() ? "/" : delimiter);
+    if (continuationToken != null && !continuationToken.isEmpty()) {
+      // for continuation token, the delimiter has to be empty
+      return null;
+    } else {
+      return (delimiter.isEmpty() ? "/" : delimiter);
+    }
   }
 
   public boolean useUrlEncodingType() {
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -1868,7 +1868,10 @@ public class MinioClient {
     return executeListObjectsV2(args);
   }
 
-  public ListBucketResultV2 listObjectsV2(ListObjectsArgs args) throws InvalidResponseException, NoSuchAlgorithmException, InvalidKeyException, ServerException, InternalException, XmlParserException, IOException, InsufficientDataException, ErrorResponseException {
+  public ListBucketResultV2 listObjectsV2(ListObjectsArgs args)
+      throws InvalidResponseException, NoSuchAlgorithmException, InvalidKeyException,
+          ServerException, InternalException, XmlParserException, IOException,
+          InsufficientDataException, ErrorResponseException {
     return executeListObjectsV2(
             args.bucket(),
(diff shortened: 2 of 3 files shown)


balamurugana (changes requested):
For custom object listing, extend `MinioClient` and use appropriate `protected` methods.

lem21h:
It's not a custom method - I want to have a support for the continuation token which is defined in AWS S3 interface and also MIN.IO supports it as well - therefore this should be supported

aleksanderlech:
> For custom object listing, extend `MinioClient` and use appropriate `protected` methods.

Don't you think that its a bad practice to call a protected method through the reflection just to support a standard, common S3 API case? It could be a temporary solution (hack) but not a permanent.

balamurugana:
> It's not a custom method - I want to have a support for the continuation token which is defined in AWS S3 interface and also MIN.IO supports it as well - therefore this should be supported

For low level API access, `MinioClient` provides access to such APIs by `protected` methods. All public methods are meant to higher level logic. I recommend to use low level APIs for your need.

aleksanderlech:
> > It's not a custom method - I want to have a support for the continuation token which is defined in AWS S3 interface and also MIN.IO supports it as well - therefore this should be supported
> 
> For low level API access, `MinioClient` provides access to such APIs by `protected` methods. All public methods are meant to higher level logic. I recommend to use low level APIs for your need.

In Java protected methods are not meant to be used as API.

balamurugana:
@aleksanderlech `ListObjectsArgs` is meant to higher level public `listObjects()` which is meant for three kinds of listing. You have to use [listObjectsV2()](https://minio-java.min.io/io/minio/MinioClient.html#listObjectsV2-java.lang.String-java.lang.String-java.lang.String-java.lang.String-java.lang.String-java.lang.Integer-java.lang.String-java.lang.String-boolean-boolean-com.google.common.collect.Multimap-com.google.common.collect.Multimap-) API for your custom need and there is no reason to add public API `listObjectsV2()`

Also you are free to clone `minio-java` and have your own ideas.

lem21h:
> > It's not a custom method - I want to have a support for the continuation token which is defined in AWS S3 interface and also MIN.IO supports it as well - therefore this should be supported
> 
> For low level API access, `MinioClient` provides access to such APIs by `protected` methods. All public methods are meant to higher level logic. I recommend to use low level APIs for your need.

What I have added here is NOT low level API.
The main purpose is to UTILIZE the CONTINUATION TOKEN, which can be provided here https://minio-java.min.io/io/minio/ListObjectsArgs.html#

and continue the listing.

Currently I don't see a way to get the list and continue it other than providing "startAfter" parameter.

Also this PR unifies the way the METADATA is provided to match with `statObject` - so the metadata starts without Amazon S3 prefix.

To sum up - this is far from "CUSTOM" list method and version V1 does not support continuation tokens - therefore there is no need to add it there.

balamurugana:
1. Go through [ListObjectV2 S3 API](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html) carefully. It is not possible to guess/compute continuation token to the first call unless you have fetched it from the server previously. This means this this PR is trying to solve your custom problem in a public method.
2. For any other issues, send separate PR.

PR #1193, opened 2021-06-07
fixed content type with PutObject

eesha-pamula:
Fixed content type by checking for the object type. Before this was set to default(would always default to application/octet-stream.)

### DIFF
--- a/api/src/main/java/io/minio/PutObjectArgs.java
+++ b/api/src/main/java/io/minio/PutObjectArgs.java
@@ -40,10 +40,10 @@ public class PutObjectArgs extends PutObjectBaseArgs {
     if (contentType != null) {
       return contentType;
     }
-
     contentType = Files.probeContentType(Paths.get(super.object()));
-    return (contentType != null && !contentType.isEmpty()) ? contentType : "application/octet-stream";
-
+    return (contentType != null && !contentType.isEmpty())
+        ? contentType
+        : "application/octet-stream";
   }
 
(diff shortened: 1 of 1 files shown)


api/src/main/java/io/minio/PutObjectArgs.java
```diff
@@ -35,7 +37,13 @@ public BufferedInputStream stream() {
    */
   public String contentType() throws IOException {
     String contentType = super.contentType();
-    return (contentType != null) ? contentType : "application/octet-stream";
+    if (contentType != null) {
+      return contentType;
+    }
+    contentType = Files.probeContentType(Paths.get(super.object()));
```
balamurugana, line 43: `putObject()` method is to upload `stream` to an object in S3 service. It doesn't matter whether a local file has object name or not. If your requirement is to upload a file to an object in S3 service, use `uploadObject()` method. This method detects content-type of the file.

balamurugana (changes requested):

api/src/main/java/io/minio/PutObjectArgs.java
eesha-pamula: then it seems that application/octet-stream is the correct Content-Type when using putObject

balamurugana: Yep. If content-type is not provided, it defaults to `application/octet-stream`.

PR #1180, opened 2021-04-07
Mint - Add environment variable RUN_ON_FAIL

iternity-dotcom:
Refering to https://github.com/minio/minio/issues/11996

## Description
By setting the variable `RUN_ON_FAIL` all tests from minio-java are executed independent of failures.

 ## Motivation and Context
A single failing test from the test package causes all the remaining tests to be skipped.

## How to test this PR?
`RUN_ON_FAIL=1`
At least two tests must fail.

`RUN_ON_FAIL=0`
The behavior is not changed if the environment variable is not set or equals 0.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [ ] Documentation updated
- [ ] Unit tests added/updated

PR #1201, opened 2021-06-15
Mint - Avoid invalid JSON output

iternity-dotcom:
## Description

The console outputs have been removed.

 

## Motivation and Context

Testing with Mint Suite generates a log.json file. We want to process the file, but MinIO Java generates invalid JSON.

 

## How to test this PR?

Running the tests against a `minio server data{1...4}` should now return a valid JSON.

 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)

- [x] New feature (non-breaking change which adds functionality)

- [ ] Optimization (provides speedup with no functional changes)

- [ ] Breaking change (fix or feature that would cause existing functionality to change)

 

## Checklist:

- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)

- [ ] Documentation updated

- [ ] Unit tests added/updated

### DIFF
--- a/functional/FunctionalTest.java
+++ b/functional/FunctionalTest.java
@@ -3738,8 +3738,6 @@ public class FunctionalTest {
       return true;
     }
 
-    System.out.println("downloading " + MINIO_BINARY + " binary");
-
     Request.Builder requestBuilder = new Request.Builder();
     Request request = requestBuilder.url(HttpUrl.parse(url)).method("GET", null).build();
     OkHttpClient transport =
@@ -3791,7 +3789,6 @@ public class FunctionalTest {
     pb.redirectErrorStream(true);
     pb.redirectOutput(ProcessBuilder.Redirect.to(new File(MINIO_BINARY + ".log")));
 
-    System.out.println("starting minio server");
     Process p = pb.start();
     Thread.sleep(10 * 1000); // wait for 10 seconds to do real start.
     return p;


balamurugana:
Run functional test like below instead of changing current behavior
```
FunctionalTest <ENDPOINT> <ACCESSKEY> <SECRETKEY> <REGION>
```

vadmeste (approved):
LGTM

balamurugana (changes requested):
Run functional test like below instead of changing current behavior
```
FunctionalTest <ENDPOINT> <ACCESSKEY> <SECRETKEY> <REGION>
```

vadmeste:
> Run functional test like below instead of changing current behavior
> 
> ```
> FunctionalTest <ENDPOINT> <ACCESSKEY> <SECRETKEY> <REGION>
> ```

Oh downloadMinio() is not executed during functional tests ?

balamurugana:
> > Run functional test like below instead of changing current behavior
> > ```
> > FunctionalTest <ENDPOINT> <ACCESSKEY> <SECRETKEY> <REGION>
> > ```
> 
> Oh downloadMinio() is not executed during functional tests ?

It is called if not those four arguments are passed.

iternity-dotcom:
@balamurugana: So this call has to be changed somewhere else? I don‘t quite understand.

balamurugana:
@iternity-dotcom The functional test can be run in two ways
1. Without any arguments
  - it downloads `minio` server and runs with fixed configuration for functional tests to run
2. With arguments
  - functional tests run against on provided S3 endpoint.

If you are using functional tests on `mint`, you would need to run your own s3 server and run functional tests against that.

iternity-dotcom:
Thanks @balamurugana for your explanation.
We run the mint tests against our own Minio instance and observe the above behavior.
The functional minio-java tests are implicitly called by mint. So maybe this is a problem on how functional tests are called by mint?

@andrebruch Can you also share your thoughts on this?

iternity-dotcom:
@andrebruch Maybe we are missing the <REGION> param?

iternity-dotcom:
@balamurugana Another question: Why does it make sense to write the removed console outputs to log.json in case the functional tests are called without params (thus downloading the Minio automatically)? Wouldn't this also result in an invalid log.json?

andrebruch:
@balamurugana For us the 2nd way is relevant.
With this way no invalid JSON is actually generated.

@iternity-dotcom The 1st way was called by a change on my part in `mint`:
I passed the `RUN_ON_FAIL` variable as an environment variable and additionally (that was the mistake) as a parameter to minio-java.

I corrected the RUN_ON_FAIL-PR: https://github.com/minio/minio-java/pull/1180.
So it is correct that this PR is closed.

balamurugana:
@iternity-dotcom 

> @balamurugana Another question: Why does it make sense to write the removed console outputs to log.json in case the functional tests are called without params (thus downloading the Minio automatically)? Wouldn't this also result in an invalid log.json?

Result as JSON is outputted only if `MINT_MODE` env is set i.e. JSON output is relevant to mint only. May be you could send a fix to error out if four arguments are not passed if `MINT_MODE` env is set.

iternity-dotcom:
@balamurugana OK. Thanks for concluding the conversation here.

PR #1211, opened 2021-07-16
Fail if unsupported okhttp3 is found.

balamurugana:
As `okhttp` 4.x and 3.x use same package name `okhttp3`, `minio-java`
fails at middle of runtime if `okhttp` 3.x is used. This patch fails
at beginning if unsupported okhttp3 is found.

Signed-off-by: Bala.FA <bala.gluster@gmail.com>

### DIFF
--- a/api/src/main/java/io/minio/S3Base.java
+++ b/api/src/main/java/io/minio/S3Base.java
@@ -100,6 +100,14 @@ import okhttp3.ResponseBody;
 
 /** Core S3 API client. */
 public abstract class S3Base {
+  static {
+    try {
+      RequestBody.create(new byte[] {}, null);
+    } catch (NoSuchMethodError ex) {
+      throw new RuntimeException("Unsupported OkHttp library found. Must use okhttp >= 4.8.1", ex);
+    }
+  }
+
   protected static final String NO_SUCH_BUCKET_MESSAGE = "Bucket does not exist";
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

odin568:
As a consequence using this libary against SpringBoot 2.5.3 (latest) causes (expected) exceptions:
`Unsupported OkHttp library found. Must use okhttp >= 4.8.1`

Dependency Versions: https://docs.spring.io/spring-boot/docs/current/reference/html/dependency-versions.html#

Upgrade to okhttp4 was rejected till now, don't know for when it is planned.
https://github.com/spring-projects/spring-boot/issues/18506

Not sure if being so strict is a good idea. What is the reasoning behind?

balamurugana:
> As a consequence using this libary against SpringBoot 2.5.3 (latest) causes (expected) exceptions:
> `Unsupported OkHttp library found. Must use okhttp >= 4.8.1`
> 
> Dependency Versions: https://docs.spring.io/spring-boot/docs/current/reference/html/dependency-versions.html#
> 
> Upgrade to okhttp4 was rejected till now, don't know for when it is planned.
> [spring-projects/spring-boot#18506](https://github.com/spring-projects/spring-boot/issues/18506)
> 
> Not sure if being so strict is a good idea. What is the reasoning behind?

As mentioned in the description of this PR, the failure happens if `okhttp v3` is used.  With this PR, you get an error in prior by custom error.

PR #1221, opened 2021-08-19
Add MinIO admin client functionality

Sam-Kramer:
Adds the following calls for MinIO Admin functionality:
- Add/Set/List/Delete policies
- Add/Set/List/Delete users

I had to add a dependency on bouncy castle and implement sio-go's encryption methodology in-order to encrypt/decrypt UserInfo payloads. 

Seeing as there is no integration test, I tested against my local MinIO server running on Minikube. I commented out the tests to show how/which I tested each endpoints.


harshavardhana:
> Seeing as there is no integration test, I tested against my local MinIO server running on Minikube. I commented out the tests to show how/which I tested each endpoints.

The integration tests are here functional/*  - you can add them there.

Sam-Kramer:
> > Seeing as there is no integration test, I tested against my local MinIO server running on Minikube. I commented out the tests to show how/which I tested each endpoints.
> 
> The integration tests are here functional/* - you can add them there.

Excellent, I'll add them there! Will re-push some time tomorrow (I live in the U.K., so it's slightly late for me here 😄 )

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -17,40 +17,19 @@
 
 package io.minio;
 
+import com.fasterxml.jackson.databind.type.MapType;
```
balamurugana, line 20: You could have a separate subproject called `MinioAdmin` instead of adding this to `MinioClient` which is meant to do S3 APIs.

balamurugana (changes requested):
Keep MinIO Admin functionality as a separate package.

harshavardhana:
> Keep MinIO Admin functionality as a separate package i.e. gradle subproject.

Agreed this is the clean approach.

build.gradle
```diff
@@ -146,11 +147,131 @@ project(':api') {
...
+        repositories {
+            sonatype {
+                username = project.properties['nexusUsername']
+                password = project.properties['nexusPassword']
+            }
+        }
+    }
+}
+
+project(':minio-admin') {
```
Sam-Kramer, line 245: I just copied and pasted what `:api` had

minio-admin/src/main/java/io/minio/admin/MinioAdminClient.java
```diff
@@ -0,0 +1,523 @@
...
+    }
+
+    return response;
+  }
+
+  public static MinioAdminClient.Builder builder() {
+    return new MinioAdminClient.Builder();
+  }
+
+  public static final class Builder {
```
Sam-Kramer, line 311: A lot of this is copied from MinioClient -- I'm happy to throw some of the validation methods into a shared utility class? Could also maybe create a parent class called (MinioClientBuilder) or something as well.

minio-admin/src/main/java/io/minio/admin/MinioAdminClient.java
```diff
@@ -0,0 +1,523 @@
...
+          try {
+            this.httpClient = enableExternalCertificates(this.httpClient, filename);
+          } catch (GeneralSecurityException | IOException e) {
+            throw new RuntimeException(e);
+          }
+        }
+      }
+
+      return new MinioAdminClient(
+          baseUrl, (region != null) ? region : US_EAST_1, provider, httpClient);
```
Sam-Kramer, line 520: sanity check: region is never supplied in MinIO urls?

harshavardhana: Region for Admin API is useless @Sam-Kramer - although region can be used it is usually not necessary with MinIO. 

It can be left as is with MinioClient implementation.

build.gradle
balamurugana: I guess, `Gradle` might have a way to use common configurations. You could avoid repeated configurations.

functional/FunctionalTest.java
```diff
@@ -83,6 +83,11 @@
 import io.minio.Time;
 import io.minio.UploadObjectArgs;
 import io.minio.Xml;
+import io.minio.admin.AddPolicyArgs;
```
balamurugana, line 86: Make a separate class for testing `MinioAdminClient` than mixing here.

minio-admin/src/main/java/io/minio/admin/MinioAdminClient.java
balamurugana: 1. As this class is `MinIO` specific, you can trim lot of code here e.g. we don't need to check whether given endpoint is AWS S3 etc.
2. Feel free to make `protected` class/methods in `MinioClient` and use it here.

balamurugana (commented):
Please remove `gradle/wrapper/gradle-wrapper.jar` file in this PR.

build.gradle
Sam-Kramer: I gave this a shot but I'm a bit of a gradle n00b -- I tried defining a function, but the problem is it uses the global namespace (vs the task). Thus it breaks, as some things get re-used, when they're not expected to. Most notably, around the maven publish jar plugin.

minio-admin/src/main/java/io/minio/admin/MinioAdminClient.java
Sam-Kramer: What I ended up doing here, which greatly reduced the amount of code, is just made a `BaseBuilder<T>` in the S3Base.

One hard issue with using the protected methods as you suggested, is that the class is `static`, so it does not have a reference to the current object (as it does not exist). I guess the other alternative is to make the functions `protected static`?

build.gradle
balamurugana: Got it

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +0,0 @@
-distributionBase=GRADLE_USER_HOME
```
balamurugana, line 1: Why do we need to remove `gradle-wrapper.jar` and `gradle-wrapper.properties`?

.gitignore
```diff
@@ -15,6 +15,7 @@ hs_err_pid*
 # gradle working files
 .gradle/
 build/
+gradle/
```
balamurugana, line 18: It's not required to be ignored.

api/src/main/java/io/minio/S3Base.java
```diff
@@ -57,46 +58,23 @@
...
-import javax.net.ssl.X509TrustManager;
-import okhttp3.Headers;
-import okhttp3.HttpUrl;
-import okhttp3.OkHttpClient;
-import okhttp3.Request;
-import okhttp3.RequestBody;
-import okhttp3.Response;
-import okhttp3.ResponseBody;
+import javax.net.ssl.*;
+import okhttp3.*;
```
balamurugana, line 77: Import whatever needed instead of `*`

functional/FunctionalTest.java
```diff
@@ -407,7 +409,7 @@ public static void skipStream(InputStream stream, int len) throws Exception {
     }
   }
 
-  private static void handleException(String methodName, String args, long startTime, Exception e)
+  static void handleException(String methodName, String args, long startTime, Exception e)
```
balamurugana, line 412: Please restore this ineffective diff.

balamurugana (changes requested):
Minor comments; otherwise LGTM

Please address Github workflow error

gradle/wrapper/gradle-wrapper.properties
Sam-Kramer: Sorry, this was a misunderstanding on my side! I've restored both of these to what develop has, which should fix the codeql / circecli failures... as they have no gradle wrapper.

build.gradle
```diff
@@ -262,6 +383,7 @@ project(':functional') {
     dependencies {
         compile "junit:junit:4.13"
         compile project(':api')
+        compile project(':minio-admin')
     }
```
harshavardhana, line 387: Since this is already `minio-java` does it make sense to just call as this `:admin` or `:admin-api` ?  @balamurugana

functional/TestMinioAdminClient.java
```diff
@@ -0,0 +1,138 @@
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
```
harshavardhana, line 1: Add missing copyright header

harshavardhana (changes requested):
Please add copy right headers to all the newly added files.

build.gradle
balamurugana: @harshavardhana Makes sense. We could just have `admin`

functional/FunctionalTest.java
Sam-Kramer: @balamurugana given the failing tests, I realized this was needed as we use this method in `TestMinioAdminClient.java`. Any opinions on being able to use this function there?

build.gradle
Sam-Kramer: Agreed, I think that's a lot cleaner, will change to `admin`.

functional/FunctionalTest.java
balamurugana: Does `public` instead of `private` work for you?

Sam-Kramer: Yup!

harshavardhana:
strange the same test is failing on `windows` @Sam-Kramer are you using randomized usernames?

Sam-Kramer:
> strange the same test is failing on `windows` @Sam-Kramer are you using randomized usernames?

Hmmm, this is a bit strange, I shouldn't be: https://github.com/minio/minio-java/pull/1221/files#diff-4d85893fe8f0da0e54f6e37e5bc7ae66ae7ed183606f30081ce5b48ca8670789R31-R33

And I am just using the same accessKey/secretKey as the regular client. It's an unauthorized message, so I wonder if we are signing the request differently on a windows host?

balamurugana:
> > strange the same test is failing on `windows` @Sam-Kramer are you using randomized usernames?
> 
> Hmmm, this is a bit strange, I shouldn't be: https://github.com/minio/minio-java/pull/1221/files#diff-4d85893fe8f0da0e54f6e37e5bc7ae66ae7ed183606f30081ce5b48ca8670789R31-R33
> 
> And I am just using the same accessKey/secretKey as the regular client. It's an unauthorized message, so I wonder if we are signing the request differently on a windows host?

Github workflow on windows uses `play.minio.io`. You would need to create random names for the testing.

harshavardhana (dismissed):

Sam-Kramer:
@harshavardhana / @balamurugana , does the user for `play.minio.io` have admin permissions?

Sam-Kramer:
> @harshavardhana / @balamurugana , does the user for `play.minio.io` have admin permissions?

Friendly bump on this ^

harshavardhana:
> > @harshavardhana / @balamurugana , does the user for `play.minio.io` have admin permissions?
> 
> Friendly bump on this ^

Which user are you using? @Sam-Kramer

Sam-Kramer:
> > > @harshavardhana / @balamurugana , does the user for `play.minio.io` have admin permissions?
> > 
> > 
> > Friendly bump on this ^
> 
> Which user are you using? @Sam-Kramer

I'm simply re-using whatever the previous functional tests used for their user: https://github.com/minio/minio-java/blob/62d5c05a4eeb23df17e74dd2497e0b75b900778d/functional/FunctionalTest.java#L3863-L3865.

 I'd imagine the credentials being used are stored in the circleci test, which I don't believe I cannot see.

balamurugana:
@harshavardhana The credential information is available at https://github.com/minio/minio-java/blob/master/build.gradle#L284

harshavardhana:
> I'm simply re-using whatever the previous functional tests used for their user:

Let me check locally what is failing @Sam-Kramer

harshavardhana:
Looks like all tests run fine locally 

```
createUser()
addCannedPolicy()
setPolicy()
listUsers()
listCannedPolicies()
deleteUser()
removeCannedPolicy()

Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/6.6.1/userguide/command_line_interface.html#sec:command_line_warnings

BUILD SUCCESSFUL in 1m 31s
6 actionable tasks: 6 executed
```

balamurugana:
@harshavardhana Yep. It fails only with `play.min.io` as showed in Github workflow.

Sam-Kramer:
> Looks like all tests run fine locally
> 
> ```
> createUser()
> addCannedPolicy()
> setPolicy()
> listUsers()
> listCannedPolicies()
> deleteUser()
> removeCannedPolicy()
> 
> Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
> Use '--warning-mode all' to show the individual deprecation warnings.
> See https://docs.gradle.org/6.6.1/userguide/command_line_interface.html#sec:command_line_warnings
> 
> BUILD SUCCESSFUL in 1m 31s
> 6 actionable tasks: 6 executed
> ```

Was this ran against a local minio or against play.min.io w/ this user (https://github.com/minio/minio-java/blob/master/build.gradle#L284)? Seeing as it's an Access Denied exception, I'd imagine there's nothing wrong with the message signing itself, and rather, a permission issue?

harshavardhana:
> Was this ran against a local minio or against play.min.io w/ this user (https://github.com/minio/minio-java/blob/master/build.gradle#L284)? Seeing as it's an Access Denied exception, I'd imagine there's nothing wrong with the message signing itself, and rather, a permission issue?

This is a root user - it can never return "Access Denied" @Sam-Kramer that's why I am perplexed.

harshavardhana:
Found the problem - it is the usage of AdminClient using the signature v4 style when TLS is enabled it ends up passing UNSIGNED-PAYLOAD for `x-amz-content-sha256` this is not supported with Admin API on MinIO server. 

i.e MinIO server expects a proper x-amz-content-sha256 value not a special string even with TLS. This is unlike S3 API behavior.  If you fix that it will work fine. 

Additionally I am also going to allow x-amz-content-sha256 -> UNSIGNED-PAYLOAD value.

Sam-Kramer:
> Found the problem - it is the usage of AdminClient using the signature v4 style when TLS is enabled it ends up passing UNSIGNED-PAYLOAD for `x-amz-content-sha256` this is not supported with Admin API on MinIO server.
> 
> i.e MinIO server expects a proper x-amz-content-sha256 value not a special string even with TLS. This is unlike S3 API behavior. If you fix that it will work fine.
> 
> Additionally I am also going to allow x-amz-content-sha256 -> UNSIGNED-PAYLOAD value.

Thanks for digging into this! Excellent, I'll fix this and push

harshavardhana:
> Thanks for digging into this! Excellent, I'll fix this and push

Take a look at the PR #1231 - these fixes should be incorporated in your build.

harshavardhana:
I also took the opportunity to fix the MinIO build service windows runs on CI/CD - now MinIO on windows runs a local instance, instead of talking to https://play.min.io 

It is also doing TLS tests locally as well.

harshavardhana (dismissed):

functional/FunctionalTest.java
```diff
@@ -3857,7 +3889,19 @@ public static void main(String[] args) throws Exception {
 
     int exitValue = 0;
     try {
-      client = MinioClient.builder().endpoint(endpoint).credentials(accessKey, secretKey).build();
+      client =
+          MinioClient.builder()
+              .endpoint(endpoint)
+              .credentials(accessKey, secretKey)
+              .httpClient(getUnsafeOkHttpClient())
```
balamurugana, line 3896: Just use [ignoreCertCheck()](https://minio-java.min.io/io/minio/S3Base.html#ignoreCertCheck--) immediately after create `MinioClient`

functional/FunctionalTest.java
```diff
@@ -3815,9 +3847,9 @@ public static void main(String[] args) throws Exception {
 
     Process minioProcess = null;
 
-    String kmsKeyName = "my-minio-key";
+    String kmsKeyName = null;
     if (args.length != 4) {
-      endpoint = "http://localhost:9000";
+      endpoint = "https://localhost:9000";
```
balamurugana, line 3852: This tests should also run on non TLS setup to ensure the functionality.

api/src/main/java/io/minio/MinioProperties.java
```diff
@@ -26,7 +26,7 @@
 import java.util.logging.Logger;
 
 /** Identifies and stores version information of minio-java package at run time. */
-enum MinioProperties {
+public enum MinioProperties {
```
balamurugana, line 29: Is it required to be `public`?

balamurugana (approved):

api/src/main/java/io/minio/MinioProperties.java
harshavardhana: yes for adding USER_AGENT inside MinioAdminClient

functional/FunctionalTest.java
harshavardhana: non-TLS is not necessary @balamurugana - production workloads are TLS based so just testing these should cover these cases. 

Since TLS based are specially SSE-C ones and SSE-KMS ones rest of them do not behave any differently either with TLS or without TLS. 

So I don't see we should have non-TLS based setup as well to be tested.

harshavardhana (approved):

functional/FunctionalTest.java
balamurugana: NonTLS tests ensure code quality and correct functionality of `minio-java`. The code flow of TLS and NonTLS is different and we should not get any surprises in future. The server is in production or not is out of question for `minio-java`

harshavardhana: If you want to be very specific for certain features like sha256 calculation etc then we can enable it after this PR is merged. I will send a change @balamurugana

harshavardhana: Let's try to get this in atleast, AFAIK the functionality itself is complete in this PR and working fine with the latest MinIO releases.

balamurugana (approved):

PR #1231, opened 2021-09-14
test PR with SSL enabled

harshavardhana:


harshavardhana:
@Sam-Kramer please take the changes in this PR because these are needed to make sure that CI/CD works fine with your PR and also to circumvent one bug present on MinIO with SSE-KMS setup.

PR #1233, opened 2021-09-16
add TLS and non-TLS tests

harshavardhana:


functional/FunctionalTest.java
```diff
@@ -3880,7 +3909,6 @@ public static void main(String[] args) throws Exception {
       region = args[3];
     }
 
-    isSecureEndpoint = endpoint.toLowerCase(Locale.US).contains("https://");
```
balamurugana, line 3883: If `endpoint` is a passed argument, we should have this.

functional/FunctionalTest.java
```diff
@@ -3823,12 +3836,112 @@ public static Process runMinio() throws Exception {
...
+              .endpoint(endpoint)
+              .credentials(accessKey, secretKey)
+              .region(region)
+              .build();
+      adminClientTests = new TestMinioAdminClient(adminClient, mintEnv);
+      FunctionalTest.runTests();
+    }
+  }
+
+  public static void runCustomEndpointTests() throws Exception {
```
balamurugana, line 3903: This method is exactly duplicate of above with endpoint change. We should have one test runner irrespective of endpoint.

functional/FunctionalTest.java
```diff
@@ -3823,12 +3836,112 @@ public static Process runMinio() throws Exception {
...
+    isSecureEndpoint = endpoint.toLowerCase(Locale.US).contains("https://");
+    client = MinioClient.builder().endpoint(endpoint).credentials(accessKey, secretKey).build();
+    client.ignoreCertCheck();
```
balamurugana, line 3906: No need to do this for endpoint argument passed. User knows how to use functional test in this case.

functional/FunctionalTest.java
balamurugana: The actual fix is not removing this line and ignore using `endpointTLS` down for this case.

.github/workflows/gradle.yml
```diff
@@ -35,6 +35,7 @@ jobs:
           ./gradlew runFunctionalTest
           javac -cp api/build/libs/minio-${DEV_VERSION}-all.jar functional/TestUserAgent.java
           java -Dversion=${DEV_VERSION} -cp api/build/libs/minio-${DEV_VERSION}-all.jar:functional TestUserAgent
+          ./gradlew clean
```
balamurugana, line 38: What problem did you see to add this line?

harshavardhana: This is mainly a problem that happens sporadically causing TestUserAgent.java build to fail.

functional/FunctionalTest.java
```diff
@@ -3823,12 +3836,70 @@ public static Process runMinio() throws Exception {
...
+    if (automated) {
+      // Run TLS tests on TLS endpoint
+      client =
+          MinioClient.builder().endpoint(endpointTLS).credentials(accessKey, secretKey).build();
+      client.ignoreCertCheck();
+      adminClient =
+          MinioAdminClient.builder()
+              .endpoint(endpointTLS)
+              .credentials(accessKey, secretKey)
+              .httpClient(getUnsafeOkHttpClient())
```
balamurugana, line 3868: You could remove this and use `adminClient.ignoreCertCheck()`

balamurugana (commented):
Minor correction

functional/FunctionalTest.java
harshavardhana: That doesn't work @balamurugana

balamurugana: It seems `MinioAdminClient` has implementation problem. I will fix the issue in different PR

balamurugana (approved):

vadmeste (approved):
LGTM

Issue #1257, opened 2021-11-10
There is a defect in the SetBucketNotification example


23557544:
``https://github.com/minio/minio-java/blob/master/examples/SetBucketNotification.java

config.queueConfigurationList()
return
Collections.unmodifiableList

queueConfigurationList.add(queueConfiguration);
will throw UnsupportedOperationException

===========================================================
Registered individual monitoring events cannot be seen in the console

Code:
`String bucket = "test1109", queue = "arn:minio:sqs::redis_dev:redis";

        List<EventType> eventTypes = Arrays.asList(EventType.OBJECT_CREATED_PUT);

        // 定义事件通知配置对象
        NotificationConfiguration config = null;
        config = new NotificationConfiguration();

        // 定义监听配置参数
        QueueConfiguration queueConfiguration = null;
        queueConfiguration = new QueueConfiguration();
        // 设置Minio Server的监听配置的arn参数，配置监听后重启minio服务获取对应的arn值
        queueConfiguration.setQueue(queue);
        // 设置监听事件
        queueConfiguration.setEvents(eventTypes);

        // config.queueConfigurationList()返回的是只读集合
        List<QueueConfiguration> queueConfigurationList = new ArrayList<>();
        queueConfigurationList.add(queueConfiguration);
        // 设置监听队列配置
        config.setQueueConfigurationList(queueConfigurationList);

        try {
            minioClient.setBucketNotification(SetBucketNotificationArgs.builder()
                    .bucket(bucket)
                    .config(config)
                    .build()
            );
        } catch (Throwable e) {
            log.error("注册存储桶监听事件出错 bucket[{}] queue[{}] eventTypes[{}]", bucket, queue, eventTypes.toString());
            throw new CestcMinioException("注册存储桶监听事件出错", e);
        }`
![1636536385(1)](https://user-images.githubusercontent.com/26159770/141086611-7a7b9c27-18b0-47c4-96c6-4bbd4bc29b03.png)

balamurugana:
Feel free to send a fix

PR #1258, opened 2021-11-10
issues #1257 fix UnsupportedOperationException

23557544:
fix UnsupportedOperationException for SetBucketNotification.java example

### DIFF
--- a/examples/SetBucketNotification.java
+++ b/examples/SetBucketNotification.java
@@ -48,7 +48,7 @@ public class SetBucketNotification {
       NotificationConfiguration config = new NotificationConfiguration();
 
       // Add a new SQS configuration.
-      List<QueueConfiguration> queueConfigurationList = config.queueConfigurationList();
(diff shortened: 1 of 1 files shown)


examples/SetBucketNotification.java
```diff
@@ -48,7 +48,7 @@ public static void main(String[] args)
       NotificationConfiguration config = new NotificationConfiguration();
 
       // Add a new SQS configuration.
-      List<QueueConfiguration> queueConfigurationList = config.queueConfigurationList();
+      List<QueueConfiguration> queueConfigurationList = new LinkedList<>(config.queueConfigurationList());
```
balamurugana, line 51: just do `List<QueueConfiguration> queueConfigurationList = new LinkedList<>();`

balamurugana (changes requested):

examples/SetBucketNotification.java
23557544: copy that

PR #1260, opened 2021-11-10
fix SetBucketNotification example for UnsupportedOperationException

23557544:
fix UnsupportedOperationException for SetBucketNotification.java example


balamurugana (approved):
Thanks for the fix

PR #1261, opened 2021-11-10
Use standard Objects class instead of Guava

Emkas:
Minimal Java version for the project seems to be 8 (`source\targetCompatibility = JavaVersion.VERSION_1_8`). There's no need to use Guava for `equals` and `hashCode` methods. I've changed `com.google.common.base.Objects` to `java.util.Objects` and this reduced number of imports from Guava by half.


Emkas:
All `equals` and `hashcode` looks automatically generated... So I'd be nice to ensure `com.google.common.base.Objects` woudn't be use in future, but I don't know how to achieve (I don't think that in Spotless we can achieve that). Assuming my pull request will be approved, of course :-)

balamurugana:
Even though, it is not possible to completely get rid of Guava, the PR looks good to me. Methods `hashCode()` and `equals()` were actually written manually. This means it is possible to use `import com.google.common.base.Objects;` in future. Can you add a check in Github Workflow if `com.google.common.base.Objects` is used?

Emkas:
> Can you add a check in Github Workflow if com.google.common.base.Objects is used

I have zero idea what Github Workflow is (yet :-)), but I'll try.

balamurugana:
> > Can you add a check in Github Workflow if com.google.common.base.Objects is used
> 
> I have zero idea what Github Workflow is (yet :-)), but I'll try.

Just add `grep` check here https://github.com/minio/minio-java/blob/master/.github/workflows/gradle.yml#L23

Emkas:
@balamurugana Tell me if this is what you wanted. I had to 'negate' grep returning code, so I've made simple if based on https://github.com/actions/checkout/blob/main/.github/workflows/check-dist.yml#L40

.github/workflows/gradle.yml
```diff
@@ -45,3 +45,6 @@ jobs:
           $Env:Path += ';.'
           gradlew.bat build
           gradlew.bat runFunctionalTest
+      - name: Use java.util.Objects not Guava
+        run: |
+          ! grep -q -r --include \*.java com.google.common.base.Objects
```
balamurugana, line 50: 1. Move this check before `Setup java` so that it fails earlier.
2. The check should be like below
```yaml
- name: Check limited Guava usage
  run: |
    if grep -r com.google.common.base.Objects .; then echo "Error: use java.util.Objects instead of com.google.common.base.Objects" && false; fi
```

harshavardhana:
> Minimal Java version for the project seems to be 8 (`source\targetCompatibility = JavaVersion.VERSION_1_8`). There's no need to use Guava for `equals` and `hashCode` methods. I've changed `com.google.common.base.Objects` to `java.util.Objects` and this reduced number of imports from Guava by half.

Does this reduce the SDK size in some way?

.github/workflows/gradle.yml
Emkas: I've done something like this on my first try and then I thought I will make it "better" :-)

Emkas:
> Does this reduce the SDK size in some way?

Guava is still in use, so not really.

.github/workflows/gradle.yml
```diff
@@ -20,6 +20,12 @@ jobs:
 
     steps:
       - uses: actions/checkout@master
+      - name: Check limited Guava usage
+        run: |
+          if grep -q -r --include \*.java com.google.common.base.Objects; then
```
balamurugana, line 25: 1. We should  do `grep -r com.google.common.base.Objects .` in order to see affected files. I don't know why `--include` is needed.
2. This check should run only on `ubuntu`

balamurugana (changes requested):

.github/workflows/gradle.yml
Emkas: > 1. We should  do `grep -r com.google.common.base.Objects .` in order to see affected files. I don't know why `--include` is needed.

OK, I didn't know if we want any output of this grep or not.
Well, to skip files other than .java, so this is not needed, but I think this do no charm. I you want, I will remove this.

> 2. This check should run only on `ubuntu`

OK.

Emkas: I've added ``--include`` so i.e. if someone will add this in some docs or README there will be no error.

balamurugana: > I've added `--include` so i.e. if someone will add this in some docs or README there will be no error.

This PR is meant to not use Guava's `Objects` class including code and docs. I see no meaning to refer Guava's `Objects` class in docs. Applying the check to entire project is more meaningful.

Emkas: Well, as you wish. On my local machine grep always finds gradle.yml, but if it's not a case than OK. I'll change it exactly as you wrote it.

balamurugana: You are right. We would need to exclude `.github` directory to skip `.github/workflow/gradle.yml` file not to be matched. I tried in my local clone with `grep -r --exclude-dir=.github com.google.common.base.Objects .` worked as expected.

balamurugana (approved):

PR #1276, opened 2021-12-03
fix: presignedUrl docs response-content-type is only meant for GET/HEAD

harshavardhana:


docs/API.md
```diff
@@ -1315,42 +1315,34 @@ Gets presigned URL of an object for HTTP method, expiry time and custom request
...
-            .method(Method.DELETE)
-            .bucket("my-bucketname")
-            .object("my-objectname")
-            .expiry(24 * 60 * 60)
-            .build());
-System.out.println(url);
-
-// Get presigned URL string to upload 'my-objectname' in 'my-bucketname' 
-// with response-content-type as application/json and life time as one day.
+// Get presigned URL string to download 'my-objectname' in 'my-bucketname'
```
balamurugana, line 1318: If you are trying to add new examples, just add them.
If you are trying to fix existing examples, just fix relevant example and don't delete other examples.

harshavardhana: I added meaningful examples, DELETE examples are meaningless and we shouldn't presigned URLs for those.

I added in this PR what an end-user expects.

balamurugana: Do you say presigned DELETE does not work?

harshavardhana: This PR is meant to emphanize GET/PUT instead of DELETE and DELETE shouldn't be the first example.

Pre-signed URLs are meant for GET/PUT.  80% of the time it's GET so we should mention GET first and then PUT.

Commonly no one ever generates pre-signed URLs for DELETEs however they work in S3 API.  Whoever has that niche need will figure it out for themselves doesn't need a mention in the docs.

balamurugana: You could add GET and PUT as first examples and have HEAD or DELETE as the last example for API completion.

harshavardhana: Okay I will add HEAD @balamurugana

harshavardhana: updated

balamurugana (approved):

sinhaashish (approved):

PR #1277, opened 2021-12-03
Add uploadSnowballObjects() method.

balamurugana:

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -51,6 +51,9 @@ import io.minio.messages.SelectObjectContentRequest;
 import io.minio.messages.SseConfiguration;
 import io.minio.messages.Tags;
 import io.minio.messages.VersioningConfiguration;
+import java.io.BufferedOutputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -64,6 +67,7 @@ import java.nio.file.StandardCopyOption;
 import java.nio.file.StandardOpenOption;
 import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
+import java.util.Date;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
@@ -73,6 +77,9 @@ import okhttp3.HttpUrl;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Response;
+import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
+import org.apache.commons.compress.archivers.tar.TarArchiveOutputStream;
+import org.xerial.snappy.SnappyFramedOutputStream;
 
 /**
  * Simple Storage Service (aka S3) client to perform bucket and object operations.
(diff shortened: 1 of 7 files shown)


harshavardhana (approved):

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2591,6 +2598,132 @@ public void deleteObjectTags(DeleteObjectTagsArgs args)
...
+      tarOutputStream = new TarArchiveOutputStream(os);
+      for (SnowballObject object : args.objects()) {
+        if (object.filename() != null) {
+          Path filePath = Paths.get(object.filename());
+          TarArchiveEntry entry = new TarArchiveEntry(filePath.toFile(), object.name());
+          tarOutputStream.putArchiveEntry(entry);
+          Files.copy(filePath, tarOutputStream);
+        } else {
+          TarArchiveEntry entry = new TarArchiveEntry(object.name());
+          entry.setNames("minio", "minio");
```
klauspost, line 2672: This case should return an error instead. Too magic + doesn't work with multiple files anyway.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2591,6 +2598,132 @@ public void deleteObjectTags(DeleteObjectTagsArgs args)
...
+      if (args.compression()) {
+        sos = new SnappyFramedOutputStream(os);
+        os = sos;
+      }
+
+      tarOutputStream = new TarArchiveOutputStream(os);
+      for (SnowballObject object : args.objects()) {
+        if (object.filename() != null) {
+          Path filePath = Paths.get(object.filename());
+          TarArchiveEntry entry = new TarArchiveEntry(filePath.toFile(), object.name());
```
klauspost, line 2667: I am wondering it we should evaluate object names. `\` shouldn't be in object name for example - I don't know if the `Paths.get` handles that.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2591,6 +2598,132 @@ public void deleteObjectTags(DeleteObjectTagsArgs args)
...
+   * Uploads multiple objects in a single put call. It is done by creating intermediate TAR file
+   * optionally compressed which is uploaded to S3 service.
+   *
+   * <pre>Example:{@code
+   * // Upload snowball objects.
+   * List<SnowballObject> objects = new ArrayList<SnowballObject>();
+   * objects.add(
+   *     new SnowballObject(
+   *         "my-object-one",
+   *         new ByteArrayInputStream("hello".getBytes(StandardCharsets.UTF_8)),
```
klauspost, line 2611: Having to provide all streams upfront makes it seem un-scalable, either having everything in memory or as FDs seem like a big limitation.

Would it be possible to do it similar to the go version where you can send them one at the time?

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2591,6 +2598,132 @@ public void deleteObjectTags(DeleteObjectTagsArgs args)
...
+          data,
+          data.length,
+          headers,
+          args.extraQueryParams());
+    }
+
+    long length = Paths.get(args.stagingFilename()).toFile().length();
+    if (length > ObjectWriteArgs.MAX_PART_SIZE) {
+      throw new IllegalArgumentException(
+          "tarball size " + length + " is more than maximum allowed 5GiB");
```
klauspost, line 2713: The limit is 5TiB.

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2591,6 +2598,132 @@ public void deleteObjectTags(DeleteObjectTagsArgs args)
...
+    FileOutputStream fos = null;
+    BufferedOutputStream bos = null;
+    SnappyFramedOutputStream sos = null;
+    ByteArrayOutputStream baos = null;
+    TarArchiveOutputStream tarOutputStream = null;
+
+    try {
+      OutputStream os = null;
+      if (args.stagingFilename() != null) {
+        fos = new FileOutputStream(args.stagingFilename());
```
klauspost, line 2650: Why are you writing to a temp file? Can't you just stream directly to the destination?

Note that multipart upload is *not* supported, so you cannot use that anyway.

api/src/main/java/io/minio/MinioClient.java
balamurugana: `SnowballObject` class accepts one of the options
1. `InputStream` (like `io.Reader` in golang), size and optional modification date
2. just a filename

In the example, I just used a bytes as `InputStream` for simplicity.

api/src/main/java/io/minio/MinioClient.java
balamurugana: `stagingFilename` is optional. If provided, it is used. This option is available to avoid memory usage.

api/src/main/java/io/minio/MinioClient.java
balamurugana: According to S3 specification, anything can be object name. `Paths.get()` or `TarArchiveEntry` might raise an exception for some characters in the filename. But I haven't tested that.

api/src/main/java/io/minio/MinioClient.java
balamurugana: I just added user/group name as `minio`/`minio` for each tar entry in case of stream as source. Do you say this is not needed?

api/src/main/java/io/minio/MinioClient.java
balamurugana: Fixed

api/src/main/java/io/minio/MinioClient.java
klauspost: @balamurugana But you can stream directly to the destination. There is no need for the TAR file to be "done" before it is uploaded.

api/src/main/java/io/minio/MinioClient.java
klauspost: Ah. So this is user/group, not object key. No need to add that AFAICT.

klauspost:
Ah yeah, now I remember that it wasn't actually made streaming, but required size upfront: https://github.com/minio/minio/blob/be34fc9134156caa59efbaa69ea7ca8dae102d56/cmd/object-handlers.go#L1927

Makes it so much less useful.

api/src/main/java/io/minio/MinioClient.java
klauspost: So the destination key is `object.name()`? What does the `filePath.toFile()` do?

balamurugana: `filePath.toFile()` is used in `TarArchiveEntry` to get size, user/group info, modification time etc.

api/src/main/java/io/minio/MinioClient.java
balamurugana: done

klauspost (commented):
LGMT, but it has literally been about 20 years since I stopped writing java.

harshavardhana (approved):

minio-trusted (approved):

Issue #397, opened 2016-06-13
Java async client for minio


meggarr:
Would you please add **async** support in Java API for all operations, including _buckets_, _objects_, and all other client operations?

Note: **Future** is the Java's way to do async, compared to NodeJS's callback handler. If it can get a **Future** in the implementations, the API will be compatible with the JDK9 flow in the future, and also compatible with existing Java async libraries, e.g. Vert.X, RxJava etc.

balamurugana:
@meggarr is this https://github.com/balamurugana/minio-java/commit/bdf052316f4ffa5604fb9d46e24502aebfef26d2 patch good enough to solve this issue?

meggarr:
@balamurugana I think this is only a simple wrapper around synchronous call with thread pool. It is good that we have API that returns `Future`, but on the other hand, can it do the **true async** in a lower level? E.g. without providing a thread pool, can it register a listener/handler that can be called when it has a result?

If it cannot provide a way in listener/handler, I think the user can simply create his/her utility to execute the code in thread pool and then notify a handler. For example, in Vert.X, it has its way to wrap this blocking/sync code, e.g. [HazelcastAsyncMap::put](https://github.com/vert-x3/vertx-hazelcast/blob/master/src/main/java/io/vertx/spi/cluster/hazelcast/impl/HazelcastAsyncMap.java#L53-56). In this case, we don't need a new API, the existing blocking/sync API is already there.

BTW, can I ask what is the implementation inside the MinioClient ? Is it HTTP with `minio` protocol implementation? Since I saw several HTTP words in the source code.

If `minio`'s driver is based on HTTP, how about we divide the `minio`'s driver into two parts? 1. The protocol parser; 2. The transport. The protocol parser is much more standalone, and the transport is more about using HTTP client to talk to server.. Because many framework has its own HTTP client framework, e.g. Netty has its HTTP client, Apache HTTP has its http client, Vert.X also has a HTTP client (based on Netty) which is **true async**. If you provide a `minio` protocol parser, I'd rather write my own version Minio client using any of these HTTP Client.

Just my two cents.

GarimaKapoor:
@balamurugana is this issue resolved?

balamurugana:
@GarimaKapoor No. We haven't made the conclusion to made all calls using async.  This requires major rewrite of minio-java.

deekoder:
We have to discuss this and decide what we want to do about this. Assigning it to myself till then.

TopSpoofer:
[...]

PR #1286, opened 2022-01-12
Java async client for minio #397

mmadoo:
A first version of async upload (#397) 
If the direction is good, I can do some improvement to use CompletableFuture instead of OnCloseCallback for instance and add documentation.

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -67,6 +67,7 @@ import java.nio.file.StandardCopyOption;
 import java.nio.file.StandardOpenOption;
 import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
+import java.util.*;
 import java.util.Date;
 import java.util.Iterator;
 import java.util.LinkedList;
@@ -126,6 +127,7 @@ import org.xerial.snappy.SnappyFramedOutputStream;
  * }</pre>
  */
 public class MinioClient extends S3Base {
+
   private MinioClient(
       HttpUrl baseUrl,
       String region,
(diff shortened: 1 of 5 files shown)


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -119,6 +123,9 @@
  * }</pre>
  */
 public class MinioClient extends S3Base {
+	
+  private ExecutorService executor;
```
balamurugana, line 127: My recommendation is to keep async as separate client i.e. `AsyncMinioClient`. I am not sure this is the one we wanted in async support. I did something similar long back https://github.com/balamurugana/minio-java/commit/d6f5878c7d4100eafb0d7ae59506e1f0e7b6e0b1 but we would need to have full support from bottom to top i.e. from `OkHttpClient` to `MinioClient`.

mmadoo:
Hello,
Is it possible to have a feedback on this pull request ?
It have handle all TODO.

harshavardhana:
@mmadoo you can run `./gradlew check` to fix all the things in one shot before you push to this PR.  I see that you are fixing things after you submit the PR and see failure.

mmadoo:
> @mmadoo you can run `./gradlew check` to fix all the things in one shot before you push to this PR. I see that you are fixing things after you submit the PR and see failure.

Sorry,
It should be OK.
```
./gradlew check
./gradlew build
./gradlew runFunctionalTest
```
are succesfull on my computer.

mmadoo:
I have just squashed and rebased.
Test and build are localy successful.

balamurugana:
It is better to have async support like AWS SDK Java. I am looking into that and will assess how much changes in the current code.

minio-trusted:
merged in #1308

PR #1306, opened 2022-02-22
Remove resume support in downloadObject() API.

balamurugana:
In downloadObject() API expects undocumented `Range` header and to
make a reliable trusted object download, the resume support is
removed. With this change, `IllegalArgumentException` is thrown if
destination file exists and the API always downloads given object
freshly.

### DIFF
--- a/api/src/main/java/io/minio/DownloadObjectArgs.java
+++ b/api/src/main/java/io/minio/DownloadObjectArgs.java
@@ -17,7 +17,6 @@
 package io.minio;
 
 import java.nio.file.Files;
-import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Objects;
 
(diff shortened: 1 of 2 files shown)


harshavardhana (approved):

kannappanr (approved):
LGTM

geraldo-netto:
Hi,

I'm facing an issue when using this API because it throws me a No such file exception when I use the full path on file name
e.g.: 

```java
        minioClient.downloadObject(
            DownloadObjectArgs.builder()
                .bucket("my-bucketname")
                .object("my-objectname")
                .filename("/tmp/my-filename.txt") // my-filename.txt does not exist
                .build());
```

On version 8.3.6 it used to work as-is


Thank you very much/Kind Regards,

PR #1321, opened 2022-04-13
Add overwrite flag to downloadObject() API.

balamurugana:
When setting overwrite flag to true for `DownloadObjectArgs` makes overwriting destination file if it exists.

### DIFF
--- a/api/src/main/java/io/minio/DownloadObjectArgs.java
+++ b/api/src/main/java/io/minio/DownloadObjectArgs.java
@@ -16,16 +16,19 @@
 
 package io.minio;
 
-import java.nio.file.Files;
-import java.nio.file.Paths;
 import java.util.Objects;
 
 /** Argument class of {@link MinioClient#downloadObject}. */
 public class DownloadObjectArgs extends ObjectReadArgs {
   private String filename;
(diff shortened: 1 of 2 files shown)


harshavardhana (changes requested):
When we overwrite do we automatically truncate the file?

balamurugana:
> When we overwrite do we automatically truncate the file?

No. `downloadObject()` downloads object's data into a temporary file named `String tempFilename = filename + "." + S3Escaper.encode(stat.etag()) + ".part.minio";` (If this temporary file exists, it is removed before downloading to it) and renamed to actual destination file.

harshavardhana:
> > When we overwrite do we automatically truncate the file?
> 
> No. `downloadObject()` downloads object's data into a temporary file named `String tempFilename = filename + "." + S3Escaper.encode(stat.etag()) + ".part.minio";` (If this temporary file exists, it is removed before downloading to it) and renamed to actual destination file.

👍🏽

harshavardhana (approved):

harshavardhana (approved):

harshavardhana (approved):

PR #1323, opened 2022-04-21
Add getUserInfo to admin client

murphywill:
This PR adds the `getUserInfo` function to the Admin Client allowing users to get the UserInfo for a single specified MinIO user (AccessToken). This allows fetching of UserInfo for not only MinIO local users but also for MinIO LDAP Users, which are not returned by the `listUsers` function.

e.g. `mcAdmin.getUserInfo("cn=Ldap User,cn=Users,dc=company,dc=io");` will return a UserInfo object for the MinIO LDAP account `cn=Ldap User,cn=Users,dc=company,dc=io` if the user exists. This UserInfo is not returned by the `mcAdmin.listUsers()` function. 

As MinIO LDAP Users do not get a `Status` by default, a new Status enum value of `unknown` is also needed in the UserInfo class. This will be assigned to any User who does not return a status of `enable` or `disabled`.

### DIFF
--- a/adminapi/src/main/java/io/minio/admin/UserInfo.java
+++ b/adminapi/src/main/java/io/minio/admin/UserInfo.java
@@ -70,8 +70,7 @@ public class UserInfo {
 
   public static enum Status {
     ENABLED("enabled"),
-    DISABLED("disabled"),
-    UNKNOWN("unknown");
+    DISABLED("disabled");
 
     private final String value;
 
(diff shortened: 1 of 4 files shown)


adminapi/src/main/java/io/minio/admin/UserInfo.java
```diff
@@ -93,7 +94,7 @@ public static Status fromString(String statusString) {
         return DISABLED;
       }
 
-      throw new IllegalArgumentException("Unknown status " + statusString);
+      return UNKNOWN;
```
balamurugana, line 97: There is no unknown state from MinIO server. It is better to throw an error so that we can address actual value instead of treating it as `Unknown`

balamurugana (approved):

harshavardhana (approved):

PR #1339, opened 2022-06-14
Allow null values for UserInfo status

murphywill:
LDAD users in Minio do not have a `status` value - it is returned as an empty string. We need to update the UserInfo class to allow for these empty string values. so we can fetch info about Minio Ldap Users. 

When a user does not have a status assigned, we will get the following error: 

`com.fasterxml.jackson.databind.exc.ValueInstantiationException: Cannot construct instance of 'io.minio.admin.UserInfo', problem: Status must be provided`

By allowing null values for the status field in the UserInfo model class, this issue is resolved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,19 +12,19 @@ Java 1.8 or above.
 <dependency>
     <groupId>io.minio</groupId>
     <artifactId>minio</artifactId>
-    <version>8.4.0</version>
+    <version>8.4.2</version>
 </dependency>
 ```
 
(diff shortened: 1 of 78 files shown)


balamurugana (changes requested):
LGTM. But send single commit PR by squashing all 15 commits.

murphywill:
Replaced by https://github.com/minio/minio-java/pull/1340 - single commit PR as requested.

PR #1374, opened 2022-10-07
fix: responseDate for retention can have single 'S'Z

harshavardhana:
Continuation of #907, we need to cover all 3
scenarios here.

### DIFF
--- a/api/src/main/java/io/minio/messages/ResponseDate.java
+++ b/api/src/main/java/io/minio/messages/ResponseDate.java
@@ -35,6 +35,9 @@ public class ResponseDate {
   public static final DateTimeFormatter MINIO_RESPONSE_DATE_FORMAT =
       DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH':'mm':'ss'Z'", Locale.US).withZone(Time.UTC);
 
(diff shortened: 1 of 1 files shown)


poornas (dismissed):

api/src/main/java/io/minio/messages/ResponseDate.java
```diff
@@ -35,6 +35,9 @@ public class ResponseDate {
   public static final DateTimeFormatter MINIO_RESPONSE_DATE_FORMAT =
       DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH':'mm':'ss'Z'", Locale.US).withZone(Time.UTC);
 
+  public static final DateTimeFormatter MINIO_RESPONSE_DATE_FORMAT_SHORT =
```
balamurugana, line 38: Why don't you comply with AWS date format like "yyyy-MM-dd'T'HH':'mm':'ss'.'SSS'Z'" ?

harshavardhana: SSS'Z allows .S'Z , .SS'Z, .SSS'Z - this is allowed in Go time.Parse() 

I am actually fixing something else here we need to support two formats for retainUntilDate 

Reading aws-sdk-java will update the PR.

harshavardhana:
Actually found the server bug, closing this.

PR #1382, opened 2022-11-08
AdminClient: add getDataUsageInfo API

### LINKED ISSUE #1381: Missing DataUsageInfo admin operation
Current MinIO admin API exposes various bucket information, related to data usage (mostly related to estimated size, how many objects are there, etc.) via `/minio/api/v3/datausageinfo` endpoint - https://github.com/minio/minio/blob/RELEASE.2022-11-08T05-27-07Z/cmd/admin-handlers.go#L489 (wrongly named endpoint, correct one is https://github.com/minio/minio/blob/76d822bf1e0c7e47cc1952edb6d3c8b64593863f/cmd/admin-router.go#L68 )

The existence of this operation in `minio-admin` Java library would be very helpful, as currently you're able to manage Quota operations (e.g. fetch current Quota settings), but you don't actually know how big the bucket is.

lukasraska:
This PR implements full support for `DataUsageInfo` object (https://github.com/minio/minio/blob/RELEASE.2022-11-08T05-27-07Z/cmd/data-usage-utils.go ), which can be retrieved via admin API.

Fixes #1381


adminapi/src/main/java/io/minio/admin/DataUsageInfo.java
```diff
@@ -0,0 +1,85 @@
...
+
+package io.minio.admin;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Map;
+
+/** Represents Data usage of the current object APi. */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class DataUsageInfo {
```
balamurugana, line 26: Add all missing objects and make it complete than partial support.

balamurugana (changes requested):

adminapi/src/main/java/io/minio/admin/DataUsageInfo.java
lukasraska: @balamurugana Done. As there are few model classes instead of just those two small ones, I've created new package to hold these. I've also included simple test that the full object can actually be deserialized via Jackson

adminapi/src/test/java/io/minio/admin/DataUsageInfoTest.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * MinIO Java SDK for Amazon S3 Compatible Cloud Storage,
+ * (C) 2021 MinIO, Inc.
```
balamurugana, line 3: fix year to 2022

adminapi/src/test/resources/json/datausageinfo.json
```diff
@@ -0,0 +1,43 @@
+{
```
balamurugana, line 1: You could embed this as a string in tests.

adminapi/src/main/java/io/minio/admin/model/TierStats.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Per-tier stats of a remote tier.
+ *
+ * @see https://github.com/minio/minio/blob/master/cmd/data-usage-cache.go#L102
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class TierStats {
+  @JsonProperty("TotalSize")
+  private long totalSize;
```
balamurugana, line 31: You could make this fields as `public` to avoid getters.

adminapi/src/main/java/io/minio/admin/model/DataUsageInfo.java
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * Represents data usage stats of the current object APi.
+ *
+ * @see https://github.com/minio/minio/blob/master/cmd/data-usage-utils.go#L69
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class DataUsageInfo {
+  @JsonProperty("lastUpdate")
+  private Date lastUpdate;
```
balamurugana, line 33: You could make this fields as `public` to avoid getters.

adminapi/src/main/java/io/minio/admin/model/BucketUsageInfo.java
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Represents bucket usage stats of the current object APi.
+ *
+ * @see https://github.com/minio/minio/blob/master/cmd/data-usage-utils.go#L47
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class BucketUsageInfo {
+  @JsonProperty("size")
+  private long size;
```
balamurugana, line 32: You could make this fields as `public` to avoid getters.

adminapi/src/main/java/io/minio/admin/model/BucketTargetUsageInfo.java
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Represents bucket replica stats of the current object APi.
+ *
+ * @see https://github.com/minio/minio/blob/master/cmd/data-usage-utils.go#L34
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class BucketTargetUsageInfo {
+  @JsonProperty("objectsPendingReplicationTotalSize")
+  private long objectsPendingReplicationTotalSize;
```
balamurugana, line 31: You could make this fields as `public` to avoid getters.

adminapi/src/main/java/io/minio/admin/model/AllTierStats.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Collection of per-tier stats across all configured remote tiers
+ *
+ * @see https://github.com/minio/minio/blob/master/cmd/data-usage-cache.go#L63
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class AllTierStats {
+  @JsonProperty("Tiers")
+  private Map<String, TierStats> tiers;
```
balamurugana, line 32: You could make this fields as `public` to avoid getters.

adminapi/src/main/java/io/minio/admin/MinioAdminClient.java
```diff
@@ -29,6 +29,7 @@
 import io.minio.S3Escaper;
 import io.minio.Signer;
 import io.minio.Time;
+import io.minio.admin.model.DataUsageInfo;
```
balamurugana, line 32: You could have `messages` instead of `model` like `io.minio.messages`

adminapi/src/test/resources/json/datausageinfo.json
lukasraska: I could, but given the string would be pretty large and this project is still supporting Java 8 (text blocks are Java 15+), I would still prefer to have it in resources

adminapi/src/main/java/io/minio/admin/model/TierStats.java
lukasraska: I want to avoid this approach as much as I can. Not only it's against all best practices in Java, but it would make the whole minio-java API inconsistent. For example https://github.com/minio/minio-java/blob/master/api/src/main/java/io/minio/messages/Stats.java doesn't expose any literals directly, only via getters. Same with the other object types - only retrievable via getters and there is no option to actually change them.

I've changed the `getTotalSize()` and all other getters in all messages into the same structure as it's used in the `api` module -> `totalSize()` (so the `get` part is omitted), also I've made all collections returned from those getters unmodifiable, to be consistent with the approach already set here. 

Let me know if it's ok this way or not. Thanks

lukasraska:
One thing that's questionable currently is the usage of `Date` for `DataUsageInfo.lastUpdate` field. While I would want to switch it to newer API (such as `ZonedDateTime` used in `api` module), it would mean introduction of new `com.fasterxml.jackson.datatype:jackson-datatype-jsr310` dependency, so it that's ok, I would include it there as well. Thanks

adminapi/src/test/resources/json/datausageinfo.json
balamurugana: > One thing that's questionable currently is the usage of `Date` for `DataUsageInfo.lastUpdate` field. While I would want to switch it to newer API (such as `ZonedDateTime` used in `api` module), it would mean introduction of new `com.fasterxml.jackson.datatype:jackson-datatype-jsr310` dependency, so it that's ok, I would include it there as well. Thanks

Please go with `ZonedDateTime`

lukasraska: @balamurugana done

balamurugana (approved):

PR #1390, opened 2022-12-07
API to track the bucket information

jithendra319:
I have added API which tracks the bucket information and store bucket information into a text file


balamurugana:
Do not send of your custom requirements.

PR #1477, opened 2023-07-28
Add a temporary connection function that can specify the generated ba…

CtrlAndZ:
Add a temporary connection function that can specify the generated base URL. Because in most cases, Java programs connect to Minio through an internal network IP, but the generated temporary access link needs to be accessed through a public network.

### DIFF
--- a/api/src/main/java/io/minio/GetPresignedObjectUrlArgs.java
+++ b/api/src/main/java/io/minio/GetPresignedObjectUrlArgs.java
@@ -27,9 +27,13 @@ import java.util.concurrent.TimeUnit;
 public class GetPresignedObjectUrlArgs extends ObjectVersionArgs {
   // default expiration for a presigned URL is 7 days in seconds
   public static final int DEFAULT_EXPIRY_TIME = (int) TimeUnit.DAYS.toSeconds(7);
-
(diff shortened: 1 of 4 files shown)


balamurugana:
Any user specific logic should be outside of minio-java. Inherit required class your side.

CtrlAndZ:
> Any user specific logic should be outside of minio-java. Inherit required class your side.

This should not be the logic of user characteristics. Minio-java should provide the ability to generate temporary links that can access specified proxy servers. [After all, most of the scenarios for using temporary links are not in the intranet environment.](https://blog.csdn.net/jianzhang11/article/details/105672261)

balamurugana:
If you want to use proxy, pass custom http client as per https://minio-java.min.io/io/minio/MinioAsyncClient.Builder.html#httpClient-okhttp3.OkHttpClient-

CtrlAndZ:
> If you want to use proxy, pass custom http client as per https://minio-java.min.io/io/minio/MinioAsyncClient.Builder.html#httpClient-okhttp3.OkHttpClient-

The program is deployed in an intranet environment and cannot connect to the external network. It is not possible to use an external network proxy address to build MinioAsyncClient.

Issue #1522, opened 2024-01-15
Add support for specifying port of custom endpoint in IamAwsProvider


jgolda:
Currently IamAwsProvider accepts providing a custom endpoint of IMDS service. It's great for testing, but in my opinion could be further improved by allowing the class'es client to provide a custom port as well. Currently in order to setup a local testing environment the user needs to set up a mock endpoint listening on port 80, which is sometimes not acceptable (eg. collides with other services).

Let me know if you're interested in such a feature. If so, i'll prepare a PR

jgolda:
Not really, as the url provided via `customEndpoint` is later parsed and the url part is ignored, only scheme, hostname is used, eg in io.minio.credentials.IamAwsProvider#fetchImdsToken. 
If the user passes eg. http://localhost:8180 then minio client attempts to connect to http://localhost:80 (as 80 is default). There is no way around it

balamurugana:
Feel free to send a fix WRT `customEndpoint` usage

PR #1523, opened 2024-01-15
IamAwsProvider: fix custom endpoint having non-default port

jgolda:
Follow-up to https://github.com/minio/minio-java/issues/1522

### DIFF
--- a/api/src/main/java/io/minio/credentials/IamAwsProvider.java
+++ b/api/src/main/java/io/minio/credentials/IamAwsProvider.java
@@ -140,6 +140,7 @@ public class IamAwsProvider extends EnvironmentProvider {
           new HttpUrl.Builder()
               .scheme(url.scheme())
               .host(url.host())
+              .port(url.port())
               .addPathSegments("latest/api/token")
               .build();
     }
@@ -188,6 +189,7 @@ public class IamAwsProvider extends EnvironmentProvider {
           new HttpUrl.Builder()
               .scheme(url.scheme())
               .host(url.host())
+              .port(url.port())
               .addPathSegments("latest/meta-data/iam/security-credentials/")
               .build();
     }
--- a/api/src/main/java/io/minio/credentials/IamAwsProvider.java
+++ b/api/src/main/java/io/minio/credentials/IamAwsProvider.java
@@ -45,11 +45,6 @@ import okhttp3.Response;
  * for Amazon EC2</a>.
  */
 public class IamAwsProvider extends EnvironmentProvider {
-
-  private static final int DEFAULT_PORT_HTTP = 80;
-  private static final int DEFAULT_PORT_HTTPS = 443;
-  private static final String SCHEME_HTTP = "http";
-  private static final String SCHEME_HTTPS = "https";
   // Custom endpoint to fetch IAM role credentials.
   private final HttpUrl customEndpoint;
   private final OkHttpClient httpClient;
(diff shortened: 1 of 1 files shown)


jgolda:
As to the pipeline errors they are most likely unrelated to my changes - same error occurs on master branch and the class i touched is not used during the testes, so it should not affect them. I manually tested the built library in an example app based on the FileUploader from project's README, just to authenticate and call an API method. 

As far as i recall the tests fail because min.io instance expected to run TLS is not running TLS, hence giving this error. 
I'll try to research it in the evening - if you have any ideas as to how to fix it, i'll also appreciate it

api/src/main/java/io/minio/credentials/IamAwsProvider.java
```diff
@@ -140,6 +140,7 @@ private String fetchImdsToken() {
           new HttpUrl.Builder()
               .scheme(url.scheme())
               .host(url.host())
+              .port(url.port())
```
balamurugana, line 143: Add port only if port and service do not match i.e. ignore port when port and service matches i.e HTTP -> 80, HTTPS -> 443

balamurugana (changes requested):

api/src/main/java/io/minio/credentials/IamAwsProvider.java
jgolda: Is there any argument to do it this way? To my best knowledge applying the default port doesn't affect the call in any negative way. Any additional logic makes this feature more complicated to understand and maintain for future developers. 

Ofc if you insist, i'll change it, but i'd love to see a rationale behind this

balamurugana: It is required for Signature v4 computation.

jgolda: Are IAM requests signed? I cannot recall seeing any signing in IamAwsProvider and cannot recall this in AWS docs. To my best knowledge these are plain HTTP requests

balamurugana: For `IamAwsProvider`, the request is not signed. However it is better to add port only if it is not standard port for correctness.

jgolda: As you wish. Uploaded new commit with the fix

api/src/main/java/io/minio/credentials/IamAwsProvider.java
```diff
@@ -184,12 +201,9 @@ private HttpUrl getIamRoleNamedUrl(String token) {
...
     } else {
-      url =
-          new HttpUrl.Builder()
-              .scheme(url.scheme())
-              .host(url.host())
-              .addPathSegments("latest/meta-data/iam/security-credentials/")
-              .build();
+      HttpUrl.Builder urlBuilder = new HttpUrl.Builder().scheme(url.scheme()).host(url.host());
+      setPortIfRequired(urlBuilder, url);
+      url = urlBuilder.addPathSegments("latest/meta-data/iam/security-credentials/").build();
```
balamurugana, line 206:
```suggestion
      url =
          new HttpUrl.Builder()
              .scheme(url.scheme())
              .host(url.host())
              .port(url.port())
              .addPathSegments("latest/meta-data/iam/security-credentials/")
              .build();
```

api/src/main/java/io/minio/credentials/IamAwsProvider.java
```diff
@@ -136,12 +141,9 @@ private String fetchImdsToken() {
...
     } else {
-      url =
-          new HttpUrl.Builder()
-              .scheme(url.scheme())
-              .host(url.host())
-              .addPathSegments("latest/api/token")
-              .build();
+      HttpUrl.Builder urlBuilder = new HttpUrl.Builder().scheme(url.scheme()).host(url.host());
+      setPortIfRequired(urlBuilder, url);
+      url = urlBuilder.addPathSegments("latest/api/token").build();
```
balamurugana, line 146:
```suggestion
      url =
          new HttpUrl.Builder()
              .scheme(url.scheme())
              .host(url.host())
              .port(url.port())
              .addPathSegments("latest/api/token")
              .build();
```

balamurugana (changes requested):
As `okhttp3.HttpUrl` is taking care of default port for `http` and `https`, this commit should be reverted

balamurugana (approved):

minio-trusted (approved):

PR #1546, opened 2024-04-11
Close HTTP client owned by MinioClient

findepi:
MinioClient builder and MinioAsyncClient builder allocate a new OkHttpClient instance. OkHttpClient manages internal resources such as dispatcher thread pool for executing HTTP requests and should be closed to dispose of these resources.

### DIFF
--- a/api/src/main/java/io/minio/S3Base.java
+++ b/api/src/main/java/io/minio/S3Base.java
@@ -99,7 +99,7 @@ import okhttp3.Response;
 import okhttp3.ResponseBody;
 
 /** Core S3 API client. */
-public abstract class S3Base {
+public abstract class S3Base implements AutoCloseable {
   static {
     try {
       RequestBody.create(new byte[] {}, null);
@@ -136,7 +136,10 @@ public abstract class S3Base {
   protected String region;
   protected Provider provider;
   protected OkHttpClient httpClient;
+  protected boolean closeHttpClient;
 
+  /** @deprecated This method is no longer supported. */
+  @Deprecated
   protected S3Base(
       HttpUrl baseUrl,
       String awsS3Prefix,
@@ -146,6 +149,28 @@ public abstract class S3Base {
       String region,
       Provider provider,
       OkHttpClient httpClient) {
+    this(
+        baseUrl,
+        awsS3Prefix,
+        awsDomainSuffix,
+        awsDualstack,
+        useVirtualStyle,
+        region,
+        provider,
+        httpClient,
+        false);
+  }
+
+  protected S3Base(
+      HttpUrl baseUrl,
+      String awsS3Prefix,
+      String awsDomainSuffix,
+      boolean awsDualstack,
+      boolean useVirtualStyle,
+      String region,
+      Provider provider,
+      OkHttpClient httpClient,
+      boolean closeHttpClient) {
     this.baseUrl = baseUrl;
     this.awsS3Prefix = awsS3Prefix;
     this.awsDomainSuffix = awsDomainSuffix;
@@ -154,6 +179,7 @@ public abstract class S3Base {
     this.region = region;
     this.provider = provider;
     this.httpClient = httpClient;
+    this.closeHttpClient = closeHttpClient;
   }
 
   /** @deprecated This method is no longer supported. */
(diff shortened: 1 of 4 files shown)


api/src/main/java/io/minio/S3Base.java
```diff
@@ -145,7 +146,8 @@ protected S3Base(
       boolean useVirtualStyle,
       String region,
       Provider provider,
-      OkHttpClient httpClient) {
+      OkHttpClient httpClient,
+      boolean closeHttpClient) {
```
balamurugana, line 150: This breaks backward compatibility.

You would need to create another constructor and have this `closeHttpClient` argument and call the newly added constructor in this constructor.

Finally deprecate this constructor.

api/src/main/java/io/minio/S3Base.java
```diff
@@ -193,6 +197,7 @@ protected S3Base(S3Base client) {
     this.region = client.region;
     this.provider = client.provider;
     this.httpClient = client.httpClient;
+    this.closeHttpClient = false;
```
balamurugana, line 200:
```suggestion
    this.closeHttpClient = client.closeHttpClient;
```

balamurugana (changes requested):

api/src/main/java/io/minio/S3Base.java
findepi: good catch, fixed!

api/src/main/java/io/minio/S3Base.java
findepi: That depends on the purpose of this "copy" constructor, i.e. what semantics we want,
If the caller may use both instances of S3Base, only one should close the HTTP client (and then the other is not usable anymore).
If the caller is supposed to use only the new instance of S3Base, yes, we should copy the flag.

For now, I will copy the flag.

balamurugana:
Please fix the CI failure

findepi:
I see this failure on the CI
```
Error: Exception in thread "main" io.minio.errors.ServerException: server failed with HTTP status code 500
	at io.minio.S3Base$1.onResponse(S3Base.java:753)
	at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:519)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:750)
```

I am able to reproduce the failure locally with (following the CI script https://github.com/minio/minio-java/blob/fcbfa54e48214ab5a1c34b44b5f52bddca364d24/.github/workflows/gradle.yml#L34-L44)
```sh
./gradlew clean build && 
DEV_VERSION="$(./gradlew properties | awk '/^version:/ { print $2 }')" && 
javac -cp api/build/libs/minio-${DEV_VERSION}-all.jar functional/TestUserAgent.java &&
java -Dversion=${DEV_VERSION} -cp api/build/libs/minio-${DEV_VERSION}-all.jar:functional TestUserAgent
```

however, I think I am doing something wrong still, as the above fails also when I run it on master.
@balamurugana please help me what should be the right command to reproduce the CI failure

balamurugana:
@findepi `TestUserAgent.java` fails due to `example.org` server failure.  Could you apply below diff to this PR?

```diff
diff --git a/functional/TestUserAgent.java b/functional/TestUserAgent.java
index fca96cfb..51180f99 100644
--- a/functional/TestUserAgent.java
+++ b/functional/TestUserAgent.java
@@ -23,7 +23,7 @@ import java.util.Scanner;
 
 public class TestUserAgent {
   public static void main(String[] args) throws Exception {
-    MinioClient client = MinioClient.builder().endpoint("http://example.org").build();
+    MinioClient client = MinioClient.builder().endpoint("http://httpbin.org").build();
     ByteArrayOutputStream baos = new ByteArrayOutputStream();
     client.traceOn(baos);
     client.bucketExists(BucketExistsArgs.builder().bucket("any-bucket-name-works").build());
```

findepi:
@balamurugana done!

balamurugana (approved):

harshavardhana (approved):

findepi:
@balamurugana thanks for review! 
@harshavardhana thanks for the merge!

PR #1573, opened 2024-07-02
remove bucket support force

dormanze:
As the title.


api/src/main/java/io/minio/MinioAsyncClient.java
```diff
@@ -1894,6 +1894,10 @@ public CompletableFuture<Void> removeBucket(RemoveBucketArgs args)
       throws InsufficientDataException, InternalException, InvalidKeyException, IOException,
           NoSuchAlgorithmException, XmlParserException {
     checkArgs(args);
+    if (args.isForce) {
+      return executeDeleteAsync(args, newMultimap("x-minio-force-delete", "true"), null)
```
balamurugana, line 1898: This MinIO server specific header is used to delete bucket or prefix forcefully. As it is two different kind of thing, either you use `extraHeaders` builder option or complete this PR only to `removeBucket` with appropriate builder methods.

balamurugana (changes requested):

Issue #1564, opened 2024-05-26
Adding tolerant reader pattern feature


janwesterkamp:
Bucket names following a naming convention - which is good.
In some cases, object store implementations do not follow this naming convention, i.e. Ceph allows to set a flag to be able to ignoring it.
As a user, I would still be able to access such a (misconfigured) object store, at least with read access - meaning having the reader to be tolerant to this.

**In Detail:**

The naming convention for AWS S3 compatible APIs does not allow uppercase letters to be used on S3 compatible object stores:

    https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html
    https://docs.ceph.com/en/latest/radosgw/s3/bucketops/#constraints

To make these names possible in Ceph, a separate flag (see note in link above about rgw_relaxed_s3_bucket_names flag) need to be set, to allow violating against this convention, i.e. allowing longer names or using uppercase letters for buckets.

In my concrete example I would like to access Copernicus Data Space Environment (CDSE) S3 Access service, which offers public access to earth observation data: https://documentation.dataspace.copernicus.eu/APIs/S3.html

Unfortunately it looks like CloudFerro as provider of this service used the Ceph flag mentioned above and uppercase letters for the bucket names and it is unclear if or when this will be fixed.

The S3 Access API offers these public available buckets:

    EODATA
    DIAS
 
At the moment, MinIO Java SDK and MinIO Client (mc) will not be able to access this object store instance (one of the biggest Ceph instances in the world, following CloudFerro).
While there are alternatives on the CLI (s3cmd, AWS CLI) that already implement the tolerant reader pattern, with Java it's only the AWS Java SDK (using the uppercase letters, but with other inconveniences for like accessing non-AWS resources), as far as I know.
However, listing the buckets with naming convention violation is possible with MinIO tooling, but not accessing objects inside.

So I would appreciate having the tolerant reader pattern implemented in the MinIO tooling to have a more lightweight access to data in environments like this.

As save way to achieve this might be to keep the strict setting by default and having a (tolerantReader) flag to be set to be able to access data violating the naming convention.

janwesterkamp:
It looks like this behaviour was under discussion in mc already before and has been implemented in MinIO Go already (but not in mc yet):

- https://github.com/minio/mc/issues/2157
- https://github.com/minio/mc/issues/2229
- https://github.com/minio/mc/issues/2134
- https://github.com/minio/mc/issues/1715
- https://github.com/minio/mc/issues/1664
[...]

PR #1576, opened 2024-08-05
Add skip bucket/region/object name validation as special usecase

rudgal:
Hello,

the minio-java client is my preferred way to access S3 buckets within java projects.

Unfortunately in my current project I need to access an on-premises S3 bucket. 
For some historic reasons the bucketname contains underscores and as such does not adhere to the official [naming rules](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html). 
Sadly I have no means to correct this bucketname easily without causing severe efforts throughout the organization.

For this reason I would appreciate to have a possibility to disable the validation in such cases.
With this PR I'd like to propose a feasible solution.

I researched passt issues/change-requests, this might also be related to #1564 .

### DIFF
--- a/api/src/main/java/io/minio/BucketArgs.java
+++ b/api/src/main/java/io/minio/BucketArgs.java
@@ -39,9 +39,13 @@ public abstract class BucketArgs extends BaseArgs {
       extends BaseArgs.Builder<B, A> {
     private static final Pattern BUCKET_NAME_REGEX =
         Pattern.compile("^[a-z0-9][a-z0-9\\.\\-]{1,61}[a-z0-9]$");
+    private boolean validateBucket = true;
 
     protected void validateBucketName(String name) {
       validateNotNull(name, "bucket name");
+      if (!validateBucket) {
+        return;
+      }
 
       if (!BUCKET_NAME_REGEX.matcher(name).find()) {
         throw new IllegalArgumentException(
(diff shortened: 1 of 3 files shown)


api/src/main/java/io/minio/BucketArgs.java
```diff
@@ -39,9 +39,13 @@ public abstract static class Builder<B extends Builder<B, A>, A extends BucketAr
       extends BaseArgs.Builder<B, A> {
     private static final Pattern BUCKET_NAME_REGEX =
         Pattern.compile("^[a-z0-9][a-z0-9\\.\\-]{1,61}[a-z0-9]$");
+    private boolean validateBucket = true;
```
balamurugana, line 42: 1. rename this member as `private boolean skipValidation = false;` and fix it accordingly
2. Add skip validation check to `ObjectArgs` as well.

api/src/test/java/io/minio/MakeBucketArgsTest.java
```diff
@@ -51,6 +51,11 @@ public void testEmptyRegionBuild() {
     Assert.fail("exception should be thrown");
   }
 
+  @Test()
+  public void testInvalidBucketNameWithoutValidation() {
```
balamurugana, line 55: Remove this obvious test

balamurugana (changes requested):

rudgal:
Thanks for your feedback, adapted it accordingly.

balamurugana (approved):

harshavardhana (approved):

Issue #1586, opened 2024-10-10
Add closeHttpClient to MinioAsyncClient.Builder


SerhiiKryvokin:
In PR #1546 the logic was added to close `httpClient` when closing minio client. 
However, if a custom `httpClient` is passed to the builder then `closeHttpClient` is set to `false`

Here is the code from `MinioAsyncClient.java`
```
boolean closeHttpClient = false;
  if (this.httpClient == null) {
    this.httpClient =
            HttpUtils.newDefaultHttpClient(
                    DEFAULT_CONNECTION_TIMEOUT, DEFAULT_CONNECTION_TIMEOUT, DEFAULT_CONNECTION_TIMEOUT);
    closeHttpClient = true;
}
```

So we set `closeHttpClient` to `true` only when `httpClient` is not passed to the builder and we create a default one  

My proposal is to allow to pass `closeHttpClient` to the builder and even make it `true` by default
I can create PR for this if the idea is approved

balamurugana:
Yes. For custom http client, we do not set to close because it is user's responsibility when to close it.  You could add a new builder method like `public Builder httpClient(OkHttpClient httpClient, boolean close)`.

SerhiiKryvokin:
Solved in PR #1587

PR #1587, opened 2024-10-11
add httpclient close option to minio client builder

SerhiiKryvokin:
Issue reference #1586

### DIFF
--- a/api/src/test/java/io/minio/MinioClientTest.java
+++ b/api/src/test/java/io/minio/MinioClientTest.java
@@ -27,6 +27,7 @@ import java.security.NoSuchAlgorithmException;
 import java.util.HashMap;
 import java.util.Map;
 import javax.crypto.KeyGenerator;
+import okhttp3.OkHttpClient;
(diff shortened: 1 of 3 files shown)


api/src/test/java/io/minio/MinioClientTest.java
```diff
@@ -362,6 +363,25 @@ public void testAwsEndpoints()
         "https://s3-accelerate.dualstack.amazonaws.com.cn/mybucket/myobject", url.split("\\?")[0]);
   }
 
+  @Test
```
balamurugana, line 366: Please remove this obvious test.

SerhiiKryvokin: Why do you prefer not to cover it with a unit test? What if somebody goes to S3Base.java and modifies `close()` method so http client is not closed anymore. Without a unit test it will only be possible to spot by code review

balamurugana (approved):

harshavardhana (approved):

PR #1605, opened 2024-12-21
Handle null value in io.minio.messages.Item.lastModified() method

hwaite:
Avert exception:

`java.lang.NullPointerException: Cannot invoke "io.minio.messages.ResponseDate.zonedDateTime()" because "this.lastModified" is null`.

Null values occur when doing a non-recursive, prefixed 'listObjects' call.

### DIFF
--- a/api/src/main/java/io/minio/messages/Item.java
+++ b/api/src/main/java/io/minio/messages/Item.java
@@ -21,6 +21,7 @@ import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
 import java.time.ZonedDateTime;
 import java.util.Map;
+import java.util.Optional;
 import org.simpleframework.xml.Element;
 
 /**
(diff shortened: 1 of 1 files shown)


api/src/main/java/io/minio/messages/Item.java
```diff
@@ -86,7 +87,9 @@ public String objectName() {
 
   /** Returns last modified time of the object. */
   public ZonedDateTime lastModified() {
-    return lastModified.zonedDateTime();
+    return Optional.ofNullable(lastModified)
```
balamurugana, line 90: @hwaite I suspect a server bug. Please share the reproducer along with your object storage server information.

balamurugana (changes requested):

api/src/main/java/io/minio/messages/Item.java
hwaite: See #1606.

api/src/main/java/io/minio/messages/Item.java
```diff
@@ -86,7 +87,9 @@ public String objectName() {
...
+      .map(ResponseDate::zonedDateTime)
+      .orElse(null);
```
balamurugana, line 92: Keep it simple

```suggestion
    return (lastModified == null) ? null : lastModified.zonedDateTime();
```

balamurugana (changes requested):

harshavardhana (approved):

PR #1655, opened 2025-07-16
feat: upgrade okhttp 5.1.0

livk-cloud:
Okhttp has released [5.1.0](https://square.github.io/okhttp/changelogs/changelog/) I think minio should be upgraded to 5.x at the same time


balamurugana:
addressed in https://github.com/minio/minio-java/pull/1657

PR #1685, opened 2025-12-25
Allow creation of synchronous MinioClient using already created MinioAsyncClient

jgolda:
Changes introduced in (or around) #1665 make it quite easy and convenient to provide a different Executor for executing asynchronous code in MinioAsyncClient. 
Unfortunatelly it's not possible to do the same trick for MinioClient, as it's not possible to create a MinioClient providing external MinioAsyncClient (or it's subclass, which would provide a different Executor). 

The change in this pull request changes the visibility of the constructor in MinioClient to protected. It should allow creation of MinioClient with external MinioAsyncClient.

### DIFF
--- a/api/src/main/java/io/minio/MinioClient.java
+++ b/api/src/main/java/io/minio/MinioClient.java
@@ -100,7 +100,7 @@ import okhttp3.OkHttpClient;
 public class MinioClient implements AutoCloseable {
   private MinioAsyncClient asyncClient = null;
 
-  private MinioClient(MinioAsyncClient asyncClient) {
+  protected MinioClient(MinioAsyncClient asyncClient) {
     this.asyncClient = asyncClient;
   }
 


api/src/main/java/io/minio/MinioClient.java
```diff
@@ -88,7 +100,7 @@
 public class MinioClient implements AutoCloseable {
   private MinioAsyncClient asyncClient = null;
 
-  private MinioClient(MinioAsyncClient asyncClient) {
+  protected MinioClient(MinioAsyncClient asyncClient) {
```
balamurugana: If you are creating a different `MinioAsyncClient` and inherit `MinioClient`, why not have your own class inheriting `MinioAsyncClient`?

balamurugana (changes requested):

api/src/main/java/io/minio/MinioClient.java
jgolda: I intend to use a class inheriting MinioAsyncClient - say CustomMinioAsyncClient. There I override the supplyAsync method, providing my own executor. It works fine. 

The problem is I cannot create an instance of MinioClient nor it's subclass, which would be configured with a CustomMinioAsyncClient. The only constructor in MinioClient accepting MinioAsyncClient is private. This blocks me from crating a synchronous client (relatively easily), which will use a custom Executor

jgolda: In practise I only need the sync client. I override the async client only to provide a custom Executor, which executes everything on the calling thread (Runnable::run).

With the private constructor in MinioClient I can't reuse the already existing error handling logic in MinioClient and have to copy-paste this boilerplate to my project

balamurugana: `MinioClient` just does
```java
    try {
      return asyncClient.API(args).join();
    } catch (CompletionException e) {
      asyncClient.throwMinioException(e);
      return null;
    }
```

What do you mean by `error handling logic in MinioClient`?  I do not think this PR adds value.

jgolda: Well I do try/catch for every method I use from async client, as I need to do more or less everything the sync client does. Now I'm copying parts of it to my code, which gives me 100-150 lines of boilerplate to support and merge with future min.io versions. 

As far as I recall master didn't change much in terms of error handling, so aligning code to changes in master wouldn't improve much I guess. It won't remove the boilerplate, it might make it different.

You're clearly negative about this change. Does it break anything or would you like it done differently? I'm open to some different approach which will allow me to avoid the boilerplate, but I would need some directions on what should be improved

PR #1693, opened 2026-03-24
fix versionId null check in getPresignedObjectUrl properly

### LINKED ISSUE #1692: Regression: getPresignedObjectUrl adds empty versionId when null
Hello,

I think i have managed to pinpoint a bug in the latest version `9.0.0`. There appears to be a regression in `getPresignedObjectUrl` method in `MinioClient` class. It looks like a simple invalid (inverted) condition in the source code.

Current code `9.0.0`
```java
Http.QueryParameters queryParams = new Http.QueryParameters();
if (args.versionId() == null) queryParams.put("versionId", args.versionId());
```

Previous version `8.6.0`
```java
Multimap<String, String> queryParams = newMultimap(args.extraQueryParams());
if (args.versionId() != null) queryParams.put("versionId", args.versionId());
```

Impact:
- When versionId is null or not set, the SDK generates `versionId=` in query params.
- When versionId is set, the SDK does not add `versionId` to query params.
- This results in malformed presigned URLs.
- Can lead to signature mismatch errors with S3-compatible backends.

Expected behavior:
- versionId should only be added if non-null.

Workaround (applicable only if you do not use `versionId` param):
- Explicitly set versionId to any non-null value.
```java
minioClient.getPresignedObjectUrl(
  GetPresignedObjectUrlArgs.builder()
  .versionId("dummy")
  // other params
  .build()
)
```

This seems like a simple regression with potentially broad impact, but it looks like a fix is fairly simple.

Paw565pl:
Fixes #1692

### DIFF
--- a/functional/TestMinioClient.java
+++ b/functional/TestMinioClient.java
@@ -1270,6 +1270,20 @@ public class TestMinioClient extends TestArgs {
                 .build(),
             expectedChecksum);
 
+        testTags = "[GET, versionId]";
+        String versionId = "dummy-version";
+        String urlString =
+            client.getPresignedObjectUrl(
+                GetPresignedObjectUrlArgs.builder()
+                    .method(Http.Method.GET)
+                    .bucket(bucketName)
+                    .object(objectName)
+                    .versionId(versionId)
+                    .build());
+        Assertions.assertTrue(
+            urlString.contains("versionId=" + versionId),
+            "versionId not found in the generated presigned URL");
+
         testTags = "[GET, expiry, query params]";
         Map<String, String> queryParams = new HashMap<>();
         queryParams.put("response-content-type", "application/json");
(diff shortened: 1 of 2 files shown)


Paw565pl:
I have added a small assertion to prevent this regression from happening in the future. If you would like any changes let me know 🙂

functional/TestMinioClient.java
```diff
@@ -1270,6 +1270,20 @@ public void testGetPresignedObjectUrlForGet() throws Exception {
...
+        String urlString =
+            client.getPresignedObjectUrl(
+                GetPresignedObjectUrlArgs.builder()
+                    .method(Http.Method.GET)
+                    .bucket(bucketName)
+                    .object(objectName)
+                    .versionId(versionId)
+                    .build());
+        Assertions.assertTrue(
+            urlString.contains("versionId=" + versionId),
```
balamurugana: Parse the generated url using `okhttp3.HttpUrl.parse()` and validate `versionId` query param for concrete check.

balamurugana (changes requested):

functional/TestMinioClient.java
Paw565pl: should be ok now

balamurugana (approved):

programista10kpln:
merge pls

PR #1694, opened 2026-04-03
fix: skip functional tests on HTTP 405/501 responses

harshavardhana:
Use HTTP status codes directly instead of checking error
code strings, so tests are properly skipped when APIs
return 405 (Method Not Allowed) or 501 (Not Implemented).

### DIFF
--- a/functional/TestArgs.java
+++ b/functional/TestArgs.java
@@ -323,7 +323,8 @@ public class TestArgs {
   public static void handleException(String methodName, String args, long startTime, Exception e)
       throws Exception {
     if (e instanceof ErrorResponseException) {
-      if (((ErrorResponseException) e).errorResponse().code().equals("NotImplemented")) {
+      int code = ((ErrorResponseException) e).response().code();
+      if (code == 405 || code == 501) {
         mintIgnoredLog(methodName, args, startTime);
         return;
       }
--- a/functional/TestMinioClient.java
+++ b/functional/TestMinioClient.java
@@ -394,7 +394,8 @@ public class TestMinioClient extends TestArgs {
           MakeBucketArgs.builder().bucket(bucketNameWithLock).objectLock(true).build());
     } catch (Exception e) {
       if (e instanceof ErrorResponseException) {
-        if (((ErrorResponseException) e).errorResponse().code().equals("NotImplemented")) {
+        int code = ((ErrorResponseException) e).response().code();
+        if (code == 405 || code == 501) {
           bucketNameWithLock = null;
           return;
         }


functional/FunctionalTest.java
```diff
@@ -1,117 +1,4252 @@
...
+        totalBytesRead += bytesRead;
+      }
+    }
+  }
+
+  public static void handleException(String methodName, String args, long startTime, Exception e)
+      throws Exception {
+    if (e instanceof ErrorResponseException) {
+      int code = ((ErrorResponseException) e).response().code();
+      if (code == 405 || code == 501) {
```
balamurugana: `MinioClient` already returns `MethodNotAllowed` S3 error code for `405` and `501` HTTP status code.
```suggestion
      if (((ErrorResponseException) e).errorResponse().code().equals("NotImplemented") || (ErrorResponseException) e).errorResponse().code().equals("MethodNotAllowed")) {
```

functional/FunctionalTest.java
```diff
@@ -1,117 +1,4252 @@
...
+      handleException("makeBucket()", null, startTime, e);
+    }
+
+    try {
+      client.makeBucket(
+          MakeBucketArgs.builder().bucket(bucketNameWithLock).objectLock(true).build());
+    } catch (Exception e) {
+      if (e instanceof ErrorResponseException) {
+        int code = ((ErrorResponseException) e).response().code();
+        if (code == 405 || code == 501) {
```
balamurugana: `MinioClient` already returns `MethodNotAllowed` S3 error code for `405` and `501` HTTP status code.
```suggestion
      if (((ErrorResponseException) e).errorResponse().code().equals("NotImplemented") || (ErrorResponseException) e).errorResponse().code().equals("MethodNotAllowed")) {
```

harshavardhana:
As mentioned in minio-py, there is a purpose of decoupling from a Code string to an actual HTTP status code, which makes it easy when many variations of S3 responses come in from different vendors.

balamurugana:
@harshavardhana 
> As mentioned in minio-py, there is a purpose of decoupling from a Code string to an actual HTTP status code, which makes it easy when many variations of S3 responses come in from different vendors.

My question is what if `NotImplemented` is returned when HTTP status code is not `405`/`501`. We don't need to care out that?

harshavardhana:
> @harshavardhana
> 
> > As mentioned in minio-py, there is a purpose of decoupling from a Code string to an actual HTTP status code, which makes it easy when many variations of S3 responses come in from different vendors.
> 
> My question is what if `NotImplemented` is returned when HTTP status code is not `405`/`501`. We don't need to care out that?

We don't have to, those codes can change anytime. It is more safe to use http status code directly since that won't change even if error code changes.

balamurugana (approved):

PR #1697, opened 2026-04-21
Handle multiple directories in SSL_CERT_DIR env variable

iiljkic:
### Description

Currently, the `SSL_CERT_DIR` environment variable is treated as a single directory path. However, this variable often contains multiple paths separated by the OS path separator. See the OpenSSL manual:

> If any directories are named on the command line, then those are processed in turn. If not, then the `SSL_CERT_DIR` environment variable is consulted; this should be a colon-separated list of directories, like the Unix` PATH` variable. If that is not set then the default directory (installation-specific but often `/usr/local/ssl/certs`) is processed.
Reference: https://docs.openssl.org/3.1/man1/openssl-rehash/#synopsis

This PR updates the certificate loading logic to support multiple directories, ensuring broader compatibility with enterprise and containerized environments.

### Changes Proposed

- Split paths safely: The `SSL_CERT_DIR` string is now split using the OS-specific path separator (`File.pathSeparator`).
- Process independently: Each directory is processed individually to load the certificates.

### Testing Done
- Verified that the code correctly parses multiple directories using OS-specific delimiters.
- Verified that the code still correctly processes single directory variable values.
- Verified that invalid or empty directory strings are gracefully ignored.
- Ran `./gradlew :api:spotlessApply`, `./gradlew build`, and `./gradlew runFunctionalTest` and confirmed everything still green.

### DIFF
--- a/api/src/main/java/io/minio/Http.java
+++ b/api/src/main/java/io/minio/Http.java
@@ -20,6 +20,7 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Multimap;
 import io.minio.credentials.Credentials;
 import io.minio.errors.MinioException;
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.RandomAccessFile;
@@ -423,15 +424,15 @@ public class Http {
   }
 
   private static int setCertificateEntry(
-      CertificateFactory cf, KeyStore ks, Path file, String namePrefix)
+      CertificateFactory cf, KeyStore ks, Path certPath, String namePrefix)
       throws CertificateException, IOException, KeyStoreException {
-    try (InputStream in = Files.newInputStream(file)) {
-      int index = 0;
+    try (InputStream in = Files.newInputStream(certPath)) {
+      int certsInFile = 0;
       while (in.available() > 0) {
         X509Certificate cert = (X509Certificate) cf.generateCertificate(in);
-        ks.setCertificateEntry(namePrefix + (index++), cert);
+        ks.setCertificateEntry(namePrefix + (certsInFile++), cert);
       }
-      return index;
+      return certsInFile;
     }
   }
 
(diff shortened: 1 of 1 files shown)


api/src/main/java/io/minio/Http.java
```diff
@@ -423,15 +424,15 @@ private static X509TrustManager buildTrustManagerFromKeyStore(KeyStore ks)
...
-      int index = 0;
+    try (InputStream in = Files.newInputStream(certPath)) {
+      int certsInFile = 0;
       while (in.available() > 0) {
         X509Certificate cert = (X509Certificate) cf.generateCertificate(in);
-        ks.setCertificateEntry(namePrefix + (index++), cert);
+        ks.setCertificateEntry(namePrefix + (certsInFile++), cert);
       }
-      return index;
+      return certsInFile;
```
balamurugana, line 435: You are just renaming variables here. I don't think it is needed.

api/src/main/java/io/minio/Http.java
```diff
@@ -444,30 +445,42 @@ private static X509TrustManager getTrustManagerFromFile(String filePath)
     return buildTrustManagerFromKeyStore(ks);
   }
 
-  private static X509TrustManager getTrustManagerFromDir(String dirPath)
+  private static X509TrustManager getTrustManagerFromDirs(String dirPaths)
```
balamurugana, line 448: I don't think we need to say `paths`. Just treat `dirPath` like `PATH` env var i.e. colon separated list of path. This applicable everywhere.

balamurugana (changes requested):

iiljkic:
Reverted the renamings to keep the diff as small as possible

api/src/main/java/io/minio/Http.java
```diff
@@ -468,6 +471,15 @@ private static X509TrustManager getTrustManagerFromDir(String dirPath)
     return buildTrustManagerFromKeyStore(ks);
   }
 
+  private static List<Path> getDirectories(String dirPath) {
+    return Stream.of(dirPath.split(File.pathSeparator))
```
balamurugana: Move the logic into `getTrustManagerFromFile()`

balamurugana (changes requested):

iiljkic:
I removed the `getDirectories(`) helper and inlined the path-splitting logic

Note: I assumed you meant to inline it inside `getTrustManagerFromDir()` rather than `getTrustManagerFromFile()`

balamurugana (approved):

harshavardhana (approved):

PR #1700, opened 2026-04-27
fix: retry transient per-object delete failures in removeObjects() cleanup

allanrogerr:
## Problem
The `removeObjects()` cleanup helper in `TestMinioClient.java` **silently discarded all deletion errors** via `ignore(r.get())`. On transient server-side errors (e.g. `SlowDown`, `InternalError`) during a batch delete of 1050 objects, the failing objects remain in the bucket. The subsequent `removeBucket()` call then returns `409 BucketNotEmpty`, causing the test to log a false FAIL from the cleanup path — even when the test logic itself passed.

This failure was observed in 5 distinct CI runs across the week ending 2026-04-24, always in `testListObjects` with 1050 objects, always as the sole minio-java failure while 13 of 14 other mint suites passed. Tracked in miniohq/eos#4608.

**CI evidence (all show the same `BucketNotEmpty` on `listObjects()` / 1050 objects):**
- https://ns-3.min.dev/self-hosted-github-runner-logs-bucket/miniohq_eos/2026/04/21/24709147300/72269330732.log
- https://ns-3.min.dev/self-hosted-github-runner-logs-bucket/miniohq_eos/2026/04/18/24608558011/71959373972.log
- https://ns-3.min.dev/self-hosted-github-runner-logs-bucket/miniohq_eos/2026/04/20/24664997228/72122064001.log
- https://ns-3.min.dev/self-hosted-github-runner-logs-bucket/miniohq_eos/2026/04/21/24744053117/72392971599.log
- https://ns-3.min.dev/self-hosted-github-runner-logs-bucket/miniohq_eos/2026/04/20/24664870144/72119918649.log

Repeated error fragment across all logs:
```
"name":"minio-java","function":"listObjects()","args":"[bucket, prefix, recursive, 1050 objects]",
"status":"FAIL","error":"S3 operation failed; ErrorResponseException{errorResponse=
ErrorResponse{code='BucketNotEmpty', message='The bucket you tried to delete is not empty',
bucketName='minio-java-test-<random>'}}`
``

## Fix

`removeObjects()` now **detects deletion errors and retries transient ones**:

- **Read errors**: Process each `DeleteResult.Error` instead of ignoring via `ignore(r.get())`
- **Non-transient errors** (`!TRANSIENT_DELETE_CODES`): throw `IOException` immediately (preserves actual failures)
- **Transient errors** (`InternalError`, `RequestTimeout`, `ServiceUnavailable`, `SlowDown`): retry the full set up to `MAX_DELETE_ATTEMPTS` (5) times with exponential backoff (1s / 2s / 4s / 8s)
- **Success**: break early if no errors returned

The `testRemoveObjects` wrapper's `finally` block is guarded by a `succeeded` flag so the cleanup is only re-attempted on test failure, preventing spurious timeouts on success.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   build:
-    name: Test on java-version ${{ matrix.java-version }} and ${{ matrix.os }}
+    name: Test on Java ${{ matrix.java-version }} in ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   build:
-    name: Test on Java ${{ matrix.java-version }} in ${{ matrix.os }}
+    name: Test on java-version ${{ matrix.java-version }} and ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
--- a/functional/TestMinioClient.java
+++ b/functional/TestMinioClient.java
@@ -1250,6 +1250,10 @@ public class TestMinioClient extends TestArgs {
         RemoveObjectArgs.builder().bucket(bucketName).object(getRandomName()).build());
   }
 
+  @edu.umd.cs.findbugs.annotations.SuppressFBWarnings(
+      value = "DE_MIGHT_IGNORE",
+      justification =
+          "Cleanup exception suppressed intentionally so original test failure propagates")
   public void testRemoveObjects(String testTags, List<ObjectWriteResponse> results)
       throws Exception {
     String methodName = "removeObjects()";
--- a/functional/TestMinioClient.java
+++ b/functional/TestMinioClient.java
@@ -1040,17 +1040,12 @@ public class TestMinioClient extends TestArgs {
   }
 
   public void removeObjects(String bucketName, List<ObjectWriteResponse> results) throws Exception {
-    // DeleteResult.Error has no versionId; keyed by name to rebuild versioned retry batches.
-    Map<String, List<ObjectWriteResponse>> byName = new HashMap<>();
-    for (ObjectWriteResponse res : results) {
-      byName.computeIfAbsent(res.object(), k -> new ArrayList<>()).add(res);
-    }
-    List<DeleteRequest.Object> toDelete =
+    List<DeleteRequest.Object> objects =
         results.stream()
             .map(r -> new DeleteRequest.Object(r.object(), r.versionId()))
             .collect(Collectors.toList());
-    Set<String> failedNames = Collections.emptySet();
-    for (int attempt = 0; attempt < MAX_DELETE_ATTEMPTS && !toDelete.isEmpty(); attempt++) {
+    boolean anyTransient = false;
+    for (int attempt = 0; attempt < MAX_DELETE_ATTEMPTS; attempt++) {
       if (attempt > 0) {
         try {
           Thread.sleep(500L << attempt); // 1s / 2s / 4s / 8s
@@ -1059,11 +1054,11 @@ public class TestMinioClient extends TestArgs {
           throw ie;
         }
       }
-      Set<String> retryNames = new HashSet<>();
+      anyTransient = false;
       IOException nonTransientErr = null;
-      for (int i = 0; i < toDelete.size(); i += DELETE_BATCH_SIZE) {
+      for (int i = 0; i < objects.size(); i += DELETE_BATCH_SIZE) {
         List<DeleteRequest.Object> chunk =
-            toDelete.subList(i, Math.min(i + DELETE_BATCH_SIZE, toDelete.size()));
+            objects.subList(i, Math.min(i + DELETE_BATCH_SIZE, objects.size()));
         for (Result<DeleteResult.Error> r :
             client.removeObjects(
                 RemoveObjectsArgs.builder().bucket(bucketName).objects(chunk).build())) {
@@ -1084,29 +1079,19 @@ public class TestMinioClient extends TestArgs {
             }
             continue; // drain current chunk's response before throwing
           }
-          retryNames.add(err.objectName());
+          anyTransient = true;
         }
         if (nonTransientErr != null) throw nonTransientErr;
       }
-      // All versions re-queued because DeleteResult.Error lacks versionId;
-      // already-deleted versions return NoSuchVersion, filtered upstream by MinioAsyncClient.
-      failedNames = retryNames;
-      toDelete = new ArrayList<>();
-      for (String name : retryNames) {
-        for (ObjectWriteResponse res : byName.getOrDefault(name, Collections.emptyList())) {
-          toDelete.add(new DeleteRequest.Object(res.object(), res.versionId()));
-        }
-      }
+      if (!anyTransient) break;
     }
-    if (!toDelete.isEmpty()) {
+    if (anyTransient) {
       throw new IOException(
-          toDelete.size()
+          results.size()
               + " object(s) not deleted after "
               + MAX_DELETE_ATTEMPTS
               + " attempts in bucket "
-              + bucketName
-              + "; sample: "
-              + failedNames.stream().limit(5).collect(Collectors.toList()));
+              + bucketName);
     }
   }
 
--- a/functional/TestMinioClient.java
+++ b/functional/TestMinioClient.java
@@ -1262,7 +1262,13 @@ public class TestMinioClient extends TestArgs {
     } catch (Exception e) {
       handleException(methodName, testTags, startTime, e);
     } finally {
-      if (!succeeded) removeObjects(bucketName, results);
+      if (!succeeded) {
+        try {
+          removeObjects(bucketName, results);
+        } catch (Exception ignored) {
+          // suppress so the original test exception propagates unmasked
+        }
+      }
     }
   }
 
--- a/functional/TestMinioClient.java
+++ b/functional/TestMinioClient.java
@@ -113,7 +113,6 @@ import io.minio.messages.Status;
 import io.minio.messages.Tags;
 import io.minio.messages.VersioningConfiguration;
 import java.io.ByteArrayInputStream;
-import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
@@ -1058,7 +1057,7 @@ public class TestMinioClient extends TestArgs {
         DeleteResult.Error err = r.get();
         String code = err.code();
         if (!TRANSIENT_DELETE_CODES.contains(code)) {
-          throw new IOException(
+          throw new Exception(
               "non-transient delete error '"
                   + code
                   + "': "
@@ -1073,7 +1072,7 @@ public class TestMinioClient extends TestArgs {
       if (!anyTransient) break;
     }
     if (anyTransient) {
-      throw new IOException(
+      throw new Exception(
           results.size()
               + " object(s) not deleted after "
               + MAX_DELETE_ATTEMPTS
@@ -1222,29 +1221,15 @@ public class TestMinioClient extends TestArgs {
         RemoveObjectArgs.builder().bucket(bucketName).object(getRandomName()).build());
   }
 
-  @edu.umd.cs.findbugs.annotations.SuppressFBWarnings(
-      value = "DE_MIGHT_IGNORE",
-      justification =
-          "Cleanup exception suppressed intentionally so original test failure propagates")
   public void testRemoveObjects(String testTags, List<ObjectWriteResponse> results)
       throws Exception {
     String methodName = "removeObjects()";
     long startTime = System.currentTimeMillis();
-    boolean succeeded = false;
     try {
       removeObjects(bucketName, results);
       mintSuccessLog(methodName, testTags, startTime);
-      succeeded = true;
     } catch (Exception e) {
       handleException(methodName, testTags, startTime, e);
-    } finally {
-      if (!succeeded) {
-        try {
-          removeObjects(bucketName, results);
-        } catch (Exception ignored) {
-          // suppress so the original test exception propagates unmasked
-        }
-      }
     }
   }
 
--- a/functional/TestMinioClient.java
+++ b/functional/TestMinioClient.java
@@ -1079,12 +1079,12 @@ public class TestMinioClient extends TestArgs {
                           + " in bucket "
                           + bucketName);
             }
-            continue; // drain remaining response before throwing
+            continue; // drain current chunk's response before throwing
           }
           retryNames.add(err.objectName());
         }
+        if (nonTransientErr != null) throw nonTransientErr;
       }
-      if (nonTransientErr != null) throw nonTransientErr;
       // All versions re-queued because DeleteResult.Error lacks versionId;
       // already-deleted versions return NoSuchVersion, filtered upstream by MinioAsyncClient.
       toDelete = new ArrayList<>();
--- a/functional/TestMinioClient.java
+++ b/functional/TestMinioClient.java
@@ -141,6 +141,9 @@ import org.junit.jupiter.api.Assertions;
     value = {"THROWS_METHOD_THROWS_CLAUSE_BASIC_EXCEPTION", "REC_CATCH_EXCEPTION"})
 public class TestMinioClient extends TestArgs {
   private static final int MAX_DELETE_RETRIES = 5;
+  // Matches the SDK's internal chunk size: MinioAsyncClient sets completed=true on the
+  // first chunk that returns errors, dropping remaining chunks in the same call.
+  private static final int DELETE_BATCH_SIZE = 1000;
   private static final Set<String> TRANSIENT_DELETE_CODES =
       Collections.unmodifiableSet(
           new HashSet<>(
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: functional/TestMinioClient.java [hunk 1]
```diff
@@ -137,6 +140,13 @@
     value = "REC",
     justification = "Allow catching super class Exception since it's tests")
 public class TestMinioClient extends TestArgs {
+  private static final int MAX_DELETE_RETRIES = 5;
+  private static final Set<String> IGNORABLE_DELETE_CODES =
+      new HashSet<>(Arrays.asList("NoSuchKey", "NoSuchVersion"));
```
balamurugana: 1. `minio-java` handles `NoSuchVersion` already https://github.com/minio/minio-java/blob/master/api/src/main/java/io/minio/MinioAsyncClient.java#L1058
2. Previously or AFAIK S3 doesn't return `NoSuchKey` non-existing object. If this is `MinIO` server recent behavior, we would need to handle it at https://github.com/minio/minio-java/blob/master/api/src/main/java/io/minio/MinioAsyncClient.java#L1058

functional/TestMinioClient.java
```diff
@@ -1027,18 +1037,44 @@ public List<ObjectWriteResponse> createObjects(String bucketName, int count, int
...
+      } catch (ErrorResponseException e) {
+        String code = e.errorResponse().code();
+        if (IGNORABLE_DELETE_CODES.contains(code)) return;
+        if (attempt == MAX_DELETE_RETRIES - 1 || !TRANSIENT_DELETE_CODES.contains(code)) throw e;
+      }
+    }
+  }
+
   public void removeObjects(String bucketName, List<ObjectWriteResponse> results) throws Exception {
+    // DeleteResult.Error only exposes objectName(); index all responses for retry lookup.
```
balamurugana: We don't need this complexity. Below is good enough
```difff
diff --git a/functional/TestMinioClient.java b/functional/TestMinioClient.java
index ed37ac18..85481869 100644
--- a/functional/TestMinioClient.java
+++ b/functional/TestMinioClient.java
@@ -92,6 +92,7 @@ import io.minio.messages.AccessControlList;
 import io.minio.messages.AccessControlPolicy;
 import io.minio.messages.CORSConfiguration;
 import io.minio.messages.DeleteRequest;
+import io.minio.messages.DeleteResult;
 import io.minio.messages.ErrorResponse;
 import io.minio.messages.EventType;
 import io.minio.messages.Filter;
@@ -1027,17 +1028,24 @@ public class TestMinioClient extends TestArgs {
   }
 
   public void removeObjects(String bucketName, List<ObjectWriteResponse> results) throws Exception {
-    List<DeleteRequest.Object> objects =
-        results.stream()
-            .map(
-                result -> {
-                  return new DeleteRequest.Object(result.object(), result.versionId());
-                })
-            .collect(Collectors.toList());
-    for (Result<?> r :
-        client.removeObjects(
-            RemoveObjectsArgs.builder().bucket(bucketName).objects(objects).build())) {
-      ignore(r.get());
+    List<String> codes =
+        Arrays.asList("InternalError", "RequestTimeout", "ServiceUnavailable", "SlowDown");
+    boolean failed = true;
+    for (int attempt = 0; attempt < 5 && failed; attempt++) {
+      if (attempt != 0) Thread.sleep(500L << attempt);
+      List<DeleteRequest.Object> objects =
+          results.stream()
+              .map(
+                  result -> {
+                    return new DeleteRequest.Object(result.object(), result.versionId());
+                  })
+              .collect(Collectors.toList());
+      for (Result<DeleteResult.Error> r :
+          client.removeObjects(
+              RemoveObjectsArgs.builder().bucket(bucketName).objects(objects).build())) {
+
+        if ((failed = codes.contains(r.get().code()))) break;
+      }
     }
   }
 
```

allanrogerr: > We don't need this complexity. Below is good enough
> 
> ```
> diff --git a/functional/TestMinioClient.java b/functional/TestMinioClient.java
> index ed37ac18..85481869 100644
> --- a/functional/TestMinioClient.java
> +++ b/functional/TestMinioClient.java
> @@ -92,6 +92,7 @@ import io.minio.messages.AccessControlList;
>  import io.minio.messages.AccessControlPolicy;
>  import io.minio.messages.CORSConfiguration;
>  import io.minio.messages.DeleteRequest;
> +import io.minio.messages.DeleteResult;
>  import io.minio.messages.ErrorResponse;
>  import io.minio.messages.EventType;
>  import io.minio.messages.Filter;
> @@ -1027,17 +1028,24 @@ public class TestMinioClient extends TestArgs {
>    }
>  
>    public void removeObjects(String bucketName, List<ObjectWriteResponse> results) throws Exception {
> -    List<DeleteRequest.Object> objects =
> -        results.stream()
> -            .map(
> -                result -> {
> -                  return new DeleteRequest.Object(result.object(), result.versionId());
> -                })
> -            .collect(Collectors.toList());
> -    for (Result<?> r :
> -        client.removeObjects(
> -            RemoveObjectsArgs.builder().bucket(bucketName).objects(objects).build())) {
> -      ignore(r.get());
> +    List<String> codes =
> +        Arrays.asList("InternalError", "RequestTimeout", "ServiceUnavailable", "SlowDown");
> +    boolean failed = true;
> +    for (int attempt = 0; attempt < 5 && failed; attempt++) {
> +      if (attempt != 0) Thread.sleep(500L << attempt);
> +      List<DeleteRequest.Object> objects =
> +          results.stream()
> +              .map(
> +                  result -> {
> +                    return new DeleteRequest.Object(result.object(), result.versionId());
> +                  })
> +              .collect(Collectors.toList());
> +      for (Result<DeleteResult.Error> r :
> +          client.removeObjects(
> +              RemoveObjectsArgs.builder().bucket(bucketName).objects(objects).build())) {
> +
> +        if ((failed = codes.contains(r.get().code()))) break;
> +      }
>      }
>    }
>  
> ```

I'll incorporate the essence of this - there are a couple of observations to make it work correctly. Thanks!

functional/TestMinioClient.java
allanrogerr: Addressed

functional/TestMinioClient.java
```diff
@@ -1028,17 +1038,63 @@ public List<ObjectWriteResponse> createObjects(String bucketName, int count, int
...
+          Thread.currentThread().interrupt();
+          throw ie;
+        }
+      }
+      Set<String> retryNames = new HashSet<>();
+      IOException nonTransientErr = null;
+      for (Result<DeleteResult.Error> r :
+          client.removeObjects(
+              RemoveObjectsArgs.builder().bucket(bucketName).objects(toDelete).build())) {
+        DeleteResult.Error err = r.get();
```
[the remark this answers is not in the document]
allanrogerr: Fixed in fb4e8d61. Confirmed via MinioAsyncClient.java L1098: `completed=true` is set when any chunk returns errors, dropping subsequent chunks in that call. `removeObjects()` now iterates `toDelete` in `DELETE_BATCH_SIZE` (1000) slices, issuing a separate `client.removeObjects()` call per slice. Each slice is fully processed independently, so a transient error in one slice cannot silently drop objects from another.

functional/TestMinioClient.java
```diff
@@ -1028,17 +1038,63 @@ public List<ObjectWriteResponse> createObjects(String bucketName, int count, int
...
             .collect(Collectors.toList());
-    for (Result<?> r :
-        client.removeObjects(
-            RemoveObjectsArgs.builder().bucket(bucketName).objects(objects).build())) {
-      ignore(r.get());
+    for (int attempt = 0; attempt < MAX_DELETE_RETRIES && !toDelete.isEmpty(); attempt++) {
+      if (attempt > 0) {
+        try {
+          Thread.sleep(500L << attempt); // 1s / 2s / 4s / 8s
+        } catch (InterruptedException ie) {
```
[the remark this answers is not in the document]
allanrogerr: The code comment `// 1s / 2s / 4s / 8s` already documents the actual schedule accurately. The PR description previously described the old `retryRemoveObject()` helper which had a different backoff starting at 500 ms; the description has been updated in this cycle to reflect the current batch-retry implementation.

functional/TestMinioClient.java
```diff
@@ -1028,17 +1038,63 @@ public List<ObjectWriteResponse> createObjects(String bucketName, int count, int
...
+      Set<String> retryNames = new HashSet<>();
+      IOException nonTransientErr = null;
+      for (Result<DeleteResult.Error> r :
+          client.removeObjects(
+              RemoveObjectsArgs.builder().bucket(bucketName).objects(toDelete).build())) {
+        DeleteResult.Error err = r.get();
+        String code = err.code();
+        if (!TRANSIENT_DELETE_CODES.contains(code)) {
+          if (nonTransientErr == null) {
+            nonTransientErr =
```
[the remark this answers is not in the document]
allanrogerr: The PR description has been rewritten to match the current implementation: batch retry loop with manual `DELETE_BATCH_SIZE` chunking, `TRANSIENT_DELETE_CODES` set for retry classification, immediate throw on non-transient errors, and the `byName` index for versioned-delete retry semantics. References to `retryRemoveObject()`, per-object retry, `NoSuchKey`/`NoSuchVersion` ignorable codes, and the 500 ms backoff description have been removed.

functional/TestMinioClient.java
```diff
@@ -1028,17 +1041,67 @@ public List<ObjectWriteResponse> createObjects(String bucketName, int count, int
...
+        for (Result<DeleteResult.Error> r :
+            client.removeObjects(
+                RemoveObjectsArgs.builder().bucket(bucketName).objects(chunk).build())) {
+          DeleteResult.Error err = r.get();
+          String code = err.code();
+          if (!TRANSIENT_DELETE_CODES.contains(code)) {
+            if (nonTransientErr == null) {
+              nonTransientErr =
+                  new IOException(
+                      "non-transient delete error '"
```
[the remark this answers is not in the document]
allanrogerr: Fixed in 34820c7c. The `if (nonTransientErr != null) throw nonTransientErr;` check is now inside the chunk loop, immediately after the inner `for` loop drains the current chunk's response. No further `client.removeObjects()` calls are issued once a non-transient error is seen.

functional/TestMinioClient.java
```diff
@@ -137,6 +141,15 @@
     value = "REC",
     justification = "Allow catching super class Exception since it's tests")
 public class TestMinioClient extends TestArgs {
+  private static final int MAX_DELETE_RETRIES = 5;
```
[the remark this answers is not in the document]
allanrogerr: Fixed in 5efbf408.

functional/TestMinioClient.java
```diff
@@ -1028,17 +1041,67 @@ public List<ObjectWriteResponse> createObjects(String bucketName, int count, int
...
+          if (!TRANSIENT_DELETE_CODES.contains(code)) {
+            if (nonTransientErr == null) {
+              nonTransientErr =
+                  new IOException(
+                      "non-transient delete error '"
+                          + code
+                          + "' on "
+                          + err.objectName()
+                          + " in bucket "
+                          + bucketName);
```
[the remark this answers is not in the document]
allanrogerr: Fixed in 5efbf408.

functional/TestMinioClient.java
```diff
@@ -1028,17 +1041,67 @@ public List<ObjectWriteResponse> createObjects(String bucketName, int count, int
...
+        }
+      }
+    }
+    if (!toDelete.isEmpty()) {
+      throw new IOException(
+          toDelete.size()
+              + " object(s) not deleted after "
+              + MAX_DELETE_RETRIES
+              + " attempts in bucket "
+              + bucketName);
```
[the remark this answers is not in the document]
allanrogerr: Fixed in 5efbf408.

functional/TestMinioClient.java
```diff
@@ -1186,13 +1255,15 @@ public void testRemoveObjects(String testTags, List<ObjectWriteResponse> results
...
     try {
       removeObjects(bucketName, results);
       mintSuccessLog(methodName, testTags, startTime);
+      succeeded = true;
     } catch (Exception e) {
       handleException(methodName, testTags, startTime, e);
     } finally {
-      removeObjects(bucketName, results);
+      if (!succeeded) removeObjects(bucketName, results);
     }
```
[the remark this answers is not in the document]
allanrogerr: Fixed in 91890504. The cleanup call in `finally` is now wrapped in try/catch so the original test exception propagates unmasked.

functional/TestMinioClient.java
```diff
@@ -1028,17 +1041,73 @@ public List<ObjectWriteResponse> createObjects(String bucketName, int count, int
...
+              nonTransientErr =
+                  new IOException(
+                      "non-transient delete error '"
+                          + code
+                          + "': "
+                          + err.message()
+                          + " on "
+                          + err.objectName()
+                          + " in bucket "
+                          + bucketName);
```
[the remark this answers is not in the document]
allanrogerr: Not included intentionally. Per-object `<Error>` elements in the S3 bulk-delete response XML ([`DeleteObjects`](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html)) only carry `Key`, `VersionId`, `Code`, and `Message` — no `RequestId` or `HostId`. Those fields exist on the HTTP response envelope (header `x-amz-request-id`), not on individual error entries. `err.requestId()` and `err.hostId()` are always null here, so including them would print `requestId=null, hostId=null` with no diagnostic value.

functional/TestMinioClient.java
```diff
@@ -1028,17 +1041,73 @@ public List<ObjectWriteResponse> createObjects(String bucketName, int count, int
...
+    for (ObjectWriteResponse res : results) {
+      byName.computeIfAbsent(res.object(), k -> new ArrayList<>()).add(res);
+    }
+    List<DeleteRequest.Object> toDelete =
         results.stream()
-            .map(
-                result -> {
-                  return new DeleteRequest.Object(result.object(), result.versionId());
-                })
+            .map(r -> new DeleteRequest.Object(r.object(), r.versionId()))
```
balamurugana, line 1041: Why not retry entire `results` on failure?

allanrogerr: Retrying all `results` on any transient failure would re-send the full ~1050-object set each time — up to 4 extra full-batch passes — when in practice O(1–2) objects actually returned `SlowDown`/`InternalError`. The `byName` index exists specifically to enable cheap, targeted retries: narrow to just the failing names, re-queue all their `(name, versionId)` pairs (because `DeleteResult.Error` omits `versionId`), and only delete what still needs deleting.

A second reason: non-transient codes throw immediately after draining the current chunk, without entering the retry loop at all. Retrying the full set on any error would wrap a non-transient failure in the same retry path, burning all 5 attempts on 1050+ objects before surfacing it.

allanrogerr: Correction on the second point I raised — that argument doesn't hold. `throw nonTransientErr` (line 1090) fires immediately after the current chunk is drained, before `toDelete` is ever reassigned to `retryNames`. The retry loop never gets a second attempt when a non-transient error occurs regardless of whether `toDelete` holds the full set or the narrowed names. So non-transient fast-fail is identical either way.

The only real argument for narrowing is efficiency: avoiding re-sending already-deleted objects in the retry batch.

allanrogerr: > Why not retry entire `results` on failure?

1.- retrying the entire results re- sends ~1050 objects per attempt when only just a couple may have actually failed transiently. I would prefer to avoid this. Your thoughts @balamurugana ?

balamurugana: It doesn't matter for tests. 
WRT `public void removeObjectst(String bucketName, List<ObjectWriteResponse> results) throws Exception`,
1. `listObjects()` test is the only method creates 1050 objects to make `removeObjects()` on cleanup. Rest of the usages deal with fewer objects.
2. `removeObjects()` test sends 3005 objects on which 3004 objects are non-existent.

We just need plain/simple code.

allanrogerr:
@balamurugana PTAL

harshavardhana (dismissed):

functional/TestMinioClient.java
allanrogerr: Done

functional/TestMinioClient.java
```diff
@@ -1030,15 +1040,46 @@ public List<ObjectWriteResponse> createObjects(String bucketName, int count, int
...
+        }
+      }
+      anyTransient = false;
+      for (Result<DeleteResult.Error> r :
+          client.removeObjects(
+              RemoveObjectsArgs.builder().bucket(bucketName).objects(objects).build())) {
+        DeleteResult.Error err = r.get();
+        String code = err.code();
+        if (!TRANSIENT_DELETE_CODES.contains(code)) {
+          throw new IOException(
```
balamurugana: Like other test, you could use `Exception` instead of `IOException`

functional/TestMinioClient.java
```diff
@@ -1030,15 +1040,46 @@ public List<ObjectWriteResponse> createObjects(String bucketName, int count, int
...
+                  + err.objectName()
+                  + " in bucket "
+                  + bucketName);
+        }
+        anyTransient = true;
+      }
+      if (!anyTransient) break;
+    }
+    if (anyTransient) {
+      throw new IOException(
```
balamurugana: same as above

functional/TestMinioClient.java
```diff
@@ -1182,17 +1223,29 @@ public void removeObject() throws Exception {
...
     String methodName = "removeObjects()";
     long startTime = System.currentTimeMillis();
+    boolean succeeded = false;
     try {
       removeObjects(bucketName, results);
       mintSuccessLog(methodName, testTags, startTime);
+      succeeded = true;
     } catch (Exception e) {
       handleException(methodName, testTags, startTime, e);
     } finally {
```
balamurugana: As this PR fixes this `removeObjects()` with retries, we don't need `finally` block and `succeeded` handling.

functional/TestMinioClient.java
allanrogerr: Already addressed in 246c1a67 — `removeObjects()` now retries the full `results` set on every transient attempt rather than narrowing to failing names, matching your suggestion.

functional/TestMinioClient.java
allanrogerr: Fixed in 9b7c5587 — switched to `Exception` and dropped the now-unused `java.io.IOException` import.

functional/TestMinioClient.java
allanrogerr: Fixed in 9b7c5587 — same change as the previous occurrence: `IOException` → `Exception`.

functional/TestMinioClient.java
allanrogerr: Fixed in 9b7c5587 — removed the `finally` block, the `succeeded` flag, and the now-unneeded `DE_MIGHT_IGNORE` SpotBugs suppression. `removeObjects()` handles transient retries internally.

balamurugana (dismissed):

shtripat (dismissed):
lgtm

marktheunissen (dismissed):
looks good but my Java is rusty :D

cniackz (dismissed):

shtripat (dismissed):

.github/workflows/gradle.yml
```diff
@@ -10,7 +10,7 @@ on:
 
 jobs:
   build:
-    name: Test on Java ${{ matrix.java-version }} in ${{ matrix.os }}
+    name: Test on java-version ${{ matrix.java-version }} and ${{ matrix.os }}
```
balamurugana:
```suggestion
    name: Test on Java ${{ matrix.java-version }} in ${{ matrix.os }}
```

allanrogerr:
@balamurugana Your recent change caused 
```
Test on java-version 11 and windows-latestExpected — Waiting for status to be reported
```
if this is desired, could you merge this PR as it?

harshavardhana (approved):

balamurugana (approved):

shtripat (approved):
lgtm

PR #1701, opened 2026-05-04
feat: add automatic S3 request retry with exponential backoff

allanrogerr:
## Description

Adds SDK-level automatic retry on transient HTTP failures via a single OkHttp interceptor (`Http.RetryInterceptor`) installed on the SDK's default `OkHttpClient` (built by `Http.newDefaultClient()` / `Http.newDefaultClient(IntSupplier)`). The interceptor retries on retryable HTTP status codes and retryable transport-level `IOException`s, then returns the final response (or the terminal exception). The retry budget is exposed as `Builder.maxRetries(int)` / `setMaxRetries(int)` and defaults to **10** attempts.

**Caller-supplied clients are used verbatim.** A custom `OkHttpClient` passed via `Builder.httpClient(...)` is not modified: no interceptor injection, no `retryOnConnectionFailure` override, no rebuild via `newBuilder()`. To opt into the SDK's retry policy on a custom client, register `Http.RetryInterceptor` yourself and pair it with `retryOnConnectionFailure(false)`. `Builder.maxRetries(...)` and `MinioAsyncClient.setMaxRetries(...)` are no-ops on the caller-supplied path unless the caller has wired the interceptor to read from there.

On the SDK-default path, `OkHttpClient.retryOnConnectionFailure(false)` is set so the SDK's interceptor is the single source of retry policy — no double-retry layered on top. The interceptor's attempt budget is bound to the per-instance `BaseS3Client.maxRetries` field via an `IntSupplier` so `setMaxRetries(int)` takes effect on the very next request.

**Retryable HTTP status codes** (8): 408, 429, 499, 500, 502, 503, 504, 520.
**Retryable transport errors**: connection reset, EOF, socket timeout, idle-connection close, etc. SSL handshake / unknown-CA / cert-path errors and the HTTP-on-HTTPS protocol mismatch are explicitly **not** retried.
**Backoff**: full-jitter exponential, base 200 ms, per-attempt cap 1 s, exponent capped at 30 to avoid bit-shift overflow. Result range per attempt is `[1, min(cap, base · 2ⁿ)]`.
**Cancellation**: `chain.call().isCanceled()` is consulted before each attempt and after every IOException so cancelled calls do not burn the retry budget.
**Region/HEAD redirect**: untouched — `executeHeadAsync` region rediscovery still works.
**Public API**: `Retry` (helper class) is package-private; the public surface is `Http.RetryInterceptor`, `Http.newDefaultClient(...)`, and the builder/setter methods.

## Side-effect-free Checksum.update

`Checksum.update`'s `RandomAccessFile` branch was rewritten to use positional `FileChannel.read(ByteBuffer, long position)` so the method no longer mutates the file pointer. As a result, `BaseS3Client.createBody` and the two cousin patterns inside `MinioAsyncClient.putObject` no longer need to bracket their checksum reads with `getFilePointer()` / `seek()`. Net effect: ~25 lines deleted across three call sites, callers simpler, behaviour identical end-to-end.

## minio-go alignment

The retry contract is intentionally aligned with [`minio-go/retry.go`](https://github.com/minio/minio-go/blob/master/retry.go):

| Item | minio-go | Java |
| --- | --- | --- |
| `MaxRetry` | 10 | 10 |
| `DefaultRetryUnit` | 200 ms | 200 ms |
| `DefaultRetryCap` | 1 s | 1 s |
| `MaxJitter` | 1.0 | 1.0 |
| Retryable HTTP codes | 8 (above) | 8 (above) |
| IOException classifier | TLS handshake / unknown-CA / HTTPS-on-HTTP excluded | same |
| Backoff formula | `min(cap, unit·2ⁿ) − rand·…` | identical |

Two deliberate divergences:

1. **No S3 error-code body probing.** Per reviewer feedback, retry decisions are HTTP-status-only. Every body-level retryable code AWS S3 / MinIO emits in practice already rides on a retryable HTTP status (`SlowDown`/`InternalError`/`Throttling` → 503/500), so observable coverage is unchanged in real-world workloads. Removed: 5 MiB body peek, XML regex parse, the 12-element S3-code set.
2. **No request re-signing per attempt.** minio-go rebuilds and re-signs the request per attempt; this Java port reuses the signed `okhttp3.Request` because retries run inside the OkHttp interceptor layer. For the default attempt budget the worst-case wall-clock retry window is well under the 15-minute S3 signing window. Documented inline on `Http.RetryInterceptor`.

## Motivation and context

The SDK had no retry capability for transient HTTP failures. Throttling (503), gateway errors (502/504), request timeouts (408), and Cloudflare 520s propagated directly to callers, requiring every consumer to implement its own retry loop.

This PR went through several iterations driven by reviewer feedback:

1. **Initial design**: `CompletableFuture`-based retry loop with a dedicated scheduler thread.
2. **Per [@balamurugana review 4243115858](https://github.com/minio/minio-java/pull/1701#pullrequestreview-4243115858)**: stripped to a simple OkHttp interceptor; scheduler dropped.
3. **Per [review 4248622939](https://github.com/minio/minio-java/pull/1701#pullrequestreview-4248622939)** and **[review 4250022733](https://github.com/minio/minio-java/pull/1701#pullrequestreview-4250022733)** (multi-thread batches):
   - Caller-supplied clients are used verbatim ([r3205524097](https://github.com/minio/minio-java/pull/1701#discussion_r3205524097), [r3206746125](https://github.com/minio/minio-java/pull/1701#discussion_r3206746125)).
   - `BaseS3Client.httpClient` is `volatile` for safe publication across `setTimeout`/`ignoreCertCheck` reassignments ([r3205557447](https://github.com/minio/minio-java/pull/1701#discussion_r3205557447)).
   - `Checksum.update` made side-effect-free on the file pointer; `createBody` save/restore (and two cousins in `MinioAsyncClient.putObject`) dropped ([r3205580978](https://github.com/minio/minio-java/pull/1701#discussion_r3205580978)).
   - S3 error-code body probing removed; HTTP-status-only retry ([r3205604391](https://github.com/minio/minio-java/pull/1701#discussion_r3205604391)).
   - Reuse `BaseS3Client.RANDOM` for fan-out object-name suffix ([r3206792852](https://github.com/minio/minio-java/pull/1701#discussion_r3206792852)).

Does not address #1700: error code `BucketNotEmpty` (HTTP 409) is not natively retryable in minio-go and stays out of the retryable set here.

## How to test this PR

1. **Unit + integration + lint (Java 17):**
   ```
   ./gradlew :api:check
   ```
   28 `RetryTest` cases: each `Retry` predicate, the backoff formula, and MockWebServer integration tests covering retry per status code, no-retry on 4xx, retry exhaustion (both `InvalidResponseException` and `ErrorResponseException` paths), `maxRetries(1)` disables retry, `setMaxRetries` post-construction, builder validation (0 and -1), connection-drop IOException retry, caller-supplied client contracts (`testUserSuppliedClientWithoutInterceptorDoesNotRetry` and `testUserSuppliedClientWithInterceptorRetries`), cert-path SSLException not retryable, and a 10 s timeout-guarded test that fires `Call.cancel()` mid-retry-loop and asserts the loop short-circuits in <5 s.

2. **Functional tests against a real server:**
   ```
   ./gradlew :functional:runFunctionalTest \
     -Pendpoint=http://localhost:9000 \
     -PaccessKey=minioadmin -PsecretKey=minioadmin -Pregion=us-east-1
   ```

3. **Mint suite via Docker:**
   ```
   docker run --rm --network host \
     -e SERVER_ENDPOINT="127.0.0.1:9000" \
     -e ACCESS_KEY="minioadmin" -e SECRET_KEY="minioadmin" \
     -e ENABLE_HTTPS=0 -e MINT_MODE="full" \
     minio/mint:local minio-java
   ```

4. **Verify the caller-supplied-client contract** (no interceptor injection, `retryOnConnectionFailure` left alone):
   ```java
   OkHttpClient custom = new OkHttpClient.Builder()
       .retryOnConnectionFailure(true)  // intentionally true; preserved verbatim
       .build();
   MinioClient client = MinioClient.builder()
       .endpoint("https://…").credentials(…)
       .httpClient(custom)
       .maxRetries(3)                   // no-op on caller-supplied path
       .build();
   ```
   To opt into SDK retry on a custom client, register the interceptor explicitly:
   ```java
   OkHttpClient custom = new OkHttpClient.Builder()
       .retryOnConnectionFailure(false)
       .addInterceptor(new Http.RetryInterceptor())
       .build();
   ```
   Both behaviours are locked in by `RetryTest`.

## Validation performed

| Check | Result |
| --- | --- |
| `:api:check` (compile + spotless + spotbugs + 28 `RetryTest` cases, Java 17) | ✅ green |
| `:api:javadoc` | ✅ green |
| `./gradlew build` (api, adminapi, examples, functional) | ✅ green |
| `:functional:runFunctionalTest` against play.min.io | ✅ runs through bucket / object / encryption / lifecycle / policy coverage (a pre-existing reference-equality bug in `TestMinioClient.testBucketCors` on `play.min.io` is unrelated to this PR — `CORSConfiguration` lacks an `equals` override, and the branch does not touch CORS code) |
| Mint suite (erasure mode, JARs from this branch overlayed into the `registry.min.dev/public/mint:aistor` image) | ✅ minio-java tests run successfully against a 4-disk MinIO erasure cluster behind nginx-TLS |
| End-to-end repro for the side-effect-free `Checksum.update` change: `client.putObject(PutObjectAPIArgs)` with a `RandomAccessFile` body and no pre-set checksum headers | ✅ upload succeeds; bytes match (the prior "remove the lines" world fails with `EOFException` after 0 bytes — confirms the fix is load-bearing on this code path) |


api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -270,8 +296,47 @@ private String[] handleRedirectResponse(
     return new String[] {code, message};
   }
 
-  /** Execute HTTP request asynchronously for given parameters. */
+  /** Execute HTTP request asynchronously for given parameters, with automatic retry. */
   protected CompletableFuture<Response> executeAsync(Http.S3Request s3request, String region) {
+    // Non-seekable bodies (raw okhttp3 RequestBody) cannot be replayed — single attempt only.
+    Http.Body body = s3request.body();
+    int maxAttempts = (body != null && body.isHttpRequestBody()) ? 1 : this.maxRetries;
```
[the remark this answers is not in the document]
allanrogerr: Fixed in a79aa113. `Http.Body` now records `fileOffset = file.getFilePointer()` at construction time. `toRequestBody()` seeks back to `fileOffset` before building each `RequestBody`, so every attempt (including retries) reads from the correct offset. Additionally, `BaseS3Client.createBody()` now saves and restores the file pointer around checksum computation, ensuring `Http.Body` captures the true starting offset even when the checksum read has advanced the pointer.

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -270,8 +296,47 @@ private String[] handleRedirectResponse(
...
+              RETRY_SCHEDULER.schedule(
+                  () ->
+                      executeWithRetry(s3request, region, maxAttempts, attempt + 1)
+                          .whenComplete(
+                              (r, t) -> {
+                                if (t != null) retryFuture.completeExceptionally(t);
+                                else retryFuture.complete(r);
+                              }),
+                  delayMs,
+                  TimeUnit.MILLISECONDS);
```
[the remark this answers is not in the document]
allanrogerr: Fixed in a79aa113. The scheduler lambda now wraps the retry dispatch in `CompletableFuture.runAsync()`, so credential fetching and request building run on `ForkJoinPool.commonPool()` rather than the single scheduler thread. The scheduler thread is released immediately after submitting the work, eliminating head-of-line blocking for concurrent pending retries.

api/src/test/java/io/minio/RetryTest.java
```diff
@@ -0,0 +1,645 @@
...
+
+      client.putObject(
+          PutObjectArgs.builder()
+              .bucket("test-bucket")
+              .object("test-object")
+              .data(new byte[0], 0)
+              .build());
+
+      Assert.assertEquals(2, server.getRequestCount());
+    }
```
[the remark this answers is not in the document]
allanrogerr: Added `testRandomAccessFileBodyRetried` in a79aa113: writes a known payload to a temp file, enqueues a 500 then a 200 in MockWebServer, and asserts that both the failed and retried PUT requests carry byte-for-byte identical bodies. Regarding `ByteBuffer`: our custom `ByteBuffer.inputStream()` returns a fresh `ByteArrayInputStream` from the underlying `ByteArrayOutputStream` on every call (reading from index 0), so `ByteBuffer` bodies are already safely replayable without any change.

api/src/main/java/io/minio/Retry.java
```diff
@@ -0,0 +1,152 @@
...
+      ImmutableSet.of(
+          408, // Request Timeout
+          429, // Too Many Requests
+          499, // Client Closed Request (nginx)
+          500, // Internal Server Error
+          502, // Bad Gateway
+          503, // Service Unavailable
+          504, // Gateway Timeout
+          520 // Cloudflare unknown error
+          );
```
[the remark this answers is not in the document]
allanrogerr: Fixed in a79aa113. Moved the closing ``);`` onto the last element line (`520); // Cloudflare unknown error`), matching the `ImmutableSet.of()` style used in `Http.java` (`NON_EMPTY_HEADERS`) and `Signer.java` (`IGNORED_HEADERS`).

api/src/main/java/io/minio/MinioClient.java
```diff
@@ -2043,6 +2052,15 @@ public Builder httpClient(OkHttpClient httpClient, boolean close) {
...
     }
 
+    /**
+     * Sets the maximum number of retry attempts per request. Pass 1 to disable automatic retries.
+     * Defaults to {@link Retry#MAX_RETRY} (10).
+     */
+    public Builder maxRetries(int maxRetries) {
+      asyncClientBuilder.maxRetries(maxRetries);
+      return this;
+    }
```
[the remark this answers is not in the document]
allanrogerr: Fixed in a79aa113. Replaced `{@link Retry#MAX_RETRY}` with `{@code 10}` in `MinioClient.Builder.maxRetries()` Javadoc to avoid a doclint link to a package-private type from a public method.

api/src/main/java/io/minio/MinioAsyncClient.java
```diff
@@ -217,6 +218,16 @@ public Builder httpClient(OkHttpClient httpClient, boolean close) {
...
 
+    /**
+     * Sets the maximum number of retry attempts per request. Pass 1 to disable automatic retries.
+     * Defaults to {@link Retry#MAX_RETRY} (10).
+     */
+    public Builder maxRetries(int maxRetries) {
+      if (maxRetries < 1) throw new IllegalArgumentException("maxRetries must be >= 1");
+      this.maxRetries = maxRetries;
+      return this;
+    }
```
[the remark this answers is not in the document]
allanrogerr: Fixed in a79aa113. Replaced `{@link Retry#MAX_RETRY}` with `{@code 10}` in `MinioAsyncClient.Builder.maxRetries()` Javadoc to avoid a doclint link to a package-private type from a public method.

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -103,6 +106,15 @@ public abstract class BaseS3Client implements AutoCloseable {
...
   private static final Set<String> TRACE_QUERY_PARAMS =
       ImmutableSet.of("retention", "legal-hold", "tagging", UPLOAD_ID, "acl", "attributes");
+
+  private static final ScheduledExecutorService RETRY_SCHEDULER =
+      Executors.newSingleThreadScheduledExecutor(
+          r -> {
+            Thread t = new Thread(r, "minio-retry-scheduler");
+            t.setDaemon(true);
+            return t;
+          });
```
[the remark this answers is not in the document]
allanrogerr: The daemon flag on the scheduler thread is the intentional design: it ensures the thread does not block JVM shutdown, which is the primary lifecycle concern for a library. A static single-thread scheduler is also deliberate — one shared executor per JVM rather than one per client instance avoids thread proliferation when many clients are created. The MinIO Java SDK does not target OSGI/web-container classloader-isolation environments where a static executor would cause loader leaks; in those environments users are expected to manage client lifecycle via `close()`. No change.

api/src/main/java/io/minio/Http.java
```diff
@@ -713,6 +714,11 @@ public Body(
         String md5Hash) {
       if (length < 0) throw new IllegalArgumentException("valid length must be provided");
       this.file = file;
+      try {
+        this.fileOffset = file.getFilePointer();
+      } catch (IOException e) {
+        throw new IllegalStateException("failed to read file position", e);
```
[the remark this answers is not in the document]
allanrogerr: `RandomAccessFile.getFilePointer()` reads an in-memory `long` field maintained by the JVM; it cannot realistically throw `IOException` in practice. `IllegalStateException` is the correct Java idiom for "this should never happen, and if it does the object is in an unrecoverable state" — distinct from `IllegalArgumentException` (caller mistake) or `IOException`/`MinioException` (recoverable network/S3 error). Changing the constructor signature to `throws IOException` would break callers; wrapping as `MinioException` would be semantically wrong (it signals S3 server errors, not local file-handle state). No change.

api/src/test/java/io/minio/RetryTest.java
```diff
@@ -0,0 +1,690 @@
...
+      server.start();
+
+      MinioClient client =
+          MinioClient.builder().endpoint(server.url("").toString()).maxRetries(3).build();
+      try {
+        client.listBuckets();
+        Assert.fail("expected exception");
+      } catch (Exception e) {
+        Assert.assertNotNull("expected non-null exception", e);
+      }
```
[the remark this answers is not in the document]
allanrogerr: Fixed in beead6e7. `testNoRetryOn404` now asserts `e instanceof ErrorResponseException`, checks `errorResponse().code() == "NoSuchBucket"`, and checks `response().code() == 404`. While adding the assertion we also fixed a pre-existing NPE: `regionCache.remove(s3request.bucket())` was called unconditionally for `NoSuchBucket` errors, but `bucket` is `null` for `listBuckets()` — a null guard was added in `BaseS3Client` to skip the cache invalidation when there is no bucket.

api/src/test/java/io/minio/RetryTest.java
```diff
@@ -0,0 +1,690 @@
...
+              .setBody(new Buffer().writeUtf8(accessDeniedXml)));
+      server.start();
+
+      MinioClient client =
+          MinioClient.builder().endpoint(server.url("").toString()).maxRetries(3).build();
+      try {
+        client.listBuckets();
+        Assert.fail("expected exception");
+      } catch (Exception e) {
+        Assert.assertNotNull("expected non-null exception", e);
```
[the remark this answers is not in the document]
allanrogerr: Fixed in beead6e7. `testNoRetryOn403` now asserts `e instanceof ErrorResponseException`, checks `errorResponse().code() == "AccessDenied"`, and checks `response().code() == 403`.

api/src/test/java/io/minio/RetryTest.java
```diff
@@ -0,0 +1,690 @@
...
+      server.start();
+
+      MinioClient client =
+          MinioClient.builder().endpoint(server.url("").toString()).maxRetries(3).build();
+      try {
+        client.listBuckets();
+        Assert.fail("expected exception after exhausted retries");
+      } catch (Exception e) {
+        Assert.assertNotNull("expected non-null exception after exhausted retries", e);
+      }
```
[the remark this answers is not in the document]
allanrogerr: Fixed in beead6e7. `testRetryExhaustedThrowsLastError` now asserts `e instanceof InvalidResponseException` (the 500 HTML response produces a non-XML exception) and checks `responseCode() == 500`. This prevents false positives from unrelated exceptions.

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -285,11 +352,6 @@ protected CompletableFuture<Response> executeAsync(Http.S3Request s3request, Str
     if (traceStream != null) traceStream.print(request.httpTraces());
 
     OkHttpClient httpClient = this.httpClient;
```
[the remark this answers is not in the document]
allanrogerr: Fixed in 60aeaed4. `doExecuteAsync` now builds the call with `httpClient.newBuilder().retryOnConnectionFailure(false).build()` so OkHttp's internal connection-failure retry is disabled and `executeWithRetry` is the sole retry policy. OkHttp's `newBuilder()` shares the connection pool and dispatcher, so this is allocation-only with no per-request overhead on pooled connections.

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -270,8 +296,49 @@ private String[] handleRedirectResponse(
...
+              }
+              Throwable cause =
+                  (throwable instanceof CompletionException) ? throwable.getCause() : throwable;
+              if (cause == null) cause = throwable;
+              if (attempt + 1 >= maxAttempts || !Retry.isRetryable(cause)) {
+                return Utils.<Response>failedFuture(cause);
+              }
+              long delayMs = Retry.computeBackoffMs(attempt + 1, RANDOM);
+              CompletableFuture<Response> retryFuture = new CompletableFuture<>();
+              RETRY_SCHEDULER.schedule(
```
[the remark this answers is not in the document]
allanrogerr: Fixed in 60aeaed4. Replaced the shared static `RANDOM` field (and its `SecureRandom`-seeded construction) with `ThreadLocalRandom.current()` at the backoff call site in `BaseS3Client`. Also updated the fan-out object-name generator in `MinioAsyncClient` which was the only other consumer of `RANDOM`; both fields and their imports are removed.

api/src/test/java/io/minio/RetryTest.java
```diff
@@ -0,0 +1,692 @@
...
+          MinioClient.builder().endpoint(server.url("").toString()).maxRetries(1).build();
+      try {
+        client.listBuckets();
+        Assert.fail("expected exception");
+      } catch (Exception e) {
+        Assert.assertNotNull("expected non-null exception", e);
+      }
+      // maxRetries=1 means a single attempt; must not retry
+      Assert.assertEquals(1, server.getRequestCount());
+    }
```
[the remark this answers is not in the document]
allanrogerr: Fixed in 60aeaed4. `testMaxRetriesOneDisablesRetry` now uses `catch (InvalidResponseException e)` with `assertEquals(500, e.responseCode())` — matching the typed-catch pattern applied to the other no-retry tests in the previous commit.

api/src/test/java/io/minio/RetryTest.java
```diff
@@ -0,0 +1,692 @@
...
+          MinioClient.builder().endpoint(server.url("").toString()).maxRetries(2).build();
+      client.setMaxRetries(1); // override to disable retries
+      try {
+        client.listBuckets();
+        Assert.fail("expected exception");
+      } catch (Exception e) {
+        Assert.assertNotNull("expected non-null exception", e);
+      }
+      Assert.assertEquals(1, server.getRequestCount());
+    }
```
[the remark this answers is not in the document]
allanrogerr: Fixed in 60aeaed4. `testSetMaxRetriesPostConstruction` now uses `catch (InvalidResponseException e)` with `assertEquals(500, e.responseCode())` — same fix as `testMaxRetriesOneDisablesRetry` above.

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -270,8 +294,49 @@ private String[] handleRedirectResponse(
...
+                  () ->
+                      CompletableFuture.runAsync(
+                          () ->
+                              executeWithRetry(s3request, region, maxAttempts, attempt + 1)
+                                  .whenComplete(
+                                      (r, t) -> {
+                                        if (t != null) retryFuture.completeExceptionally(t);
+                                        else retryFuture.complete(r);
+                                      })),
```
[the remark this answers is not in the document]
allanrogerr: Fixed in 0d687eca. Added `.exceptionally()` on the `runAsync` future so any dispatch failure (e.g. `RejectedExecutionException` from a shutting-down common pool) is propagated directly into `retryFuture` instead of silently leaving it uncompleted.

api/src/main/java/io/minio/errors/InvalidResponseException.java
balamurugana, line 1: The feature can be simplified by
```diff
diff --git a/api/src/main/java/io/minio/Http.java b/api/src/main/java/io/minio/Http.java
index 5fa9bec2..23ef9378 100644
--- a/api/src/main/java/io/minio/Http.java
+++ b/api/src/main/java/io/minio/Http.java
@@ -52,6 +52,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 import java.util.stream.Collectors;
@@ -65,9 +66,12 @@ import javax.net.ssl.SSLSocketFactory;
 import javax.net.ssl.TrustManager;
 import javax.net.ssl.TrustManagerFactory;
 import javax.net.ssl.X509TrustManager;
+import okhttp3.Interceptor;
 import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
 import okhttp3.Protocol;
+import okhttp3.Request;
+import okhttp3.Response;
 import okio.BufferedSink;
 import okio.Okio;
 
@@ -608,6 +612,54 @@ public class Http {
         client, System.getenv("SSL_CERT_FILE"), System.getenv("SSL_CERT_DIR"));
   }
 
+  public static class StatusRetryInterceptor implements Interceptor {
+    private final int maxRetries;
+    private final Set<Integer> retryStatusCodes;
+    private final long delayMs; // This acts as our "base" delay (e.g., 100ms)
+
+    public StatusRetryInterceptor(int maxRetries, Set<Integer> retryStatusCodes, long delayMs) {
+      this.maxRetries = maxRetries;
+      this.retryStatusCodes = retryStatusCodes;
+      this.delayMs = delayMs;
+    }
+
+    @Override
+    public Response intercept(Chain chain) throws IOException {
+      okhttp3.Request request = chain.request();
+      okhttp3.Response response = chain.proceed(request);
+
+      int tryCount = 0;
+
+      while (retryStatusCodes.contains(response.code()) && tryCount < maxRetries) {
+        tryCount++;
+
+        // Critical step: close the response body to avoid socket leaks
+        response.close();
+
+        if (delayMs > 0) {
+          // 1. Calculate exponential backoff limit: base * 2^tryCount
+          // This doubles the maximum potential wait time with each retry
+          long maxBackoffLimit = delayMs * (1L << tryCount); // Using bit-shift for 2^tryCount
+
+          // 2. Apply random jitter: pick a random number between 0 and maxBackoffLimit
+          long jitteredDelay = ThreadLocalRandom.current().nextLong(0, maxBackoffLimit);
+
+          try {
+            Thread.sleep(jitteredDelay);
+          } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new IOException("Retry interrupted", e);
+          }
+        }
+
+        // Execute the request again
+        response = chain.proceed(request);
+      }
+
+      return response;
+    }
+  }
+
   /**
    * Creates new HTTP client with default timeout with additional TLS certificates from
    * SSL_CERT_FILE and SSL_CERT_DIR environment variables if present.
@@ -620,6 +672,19 @@ public class Http {
             .writeTimeout(DEFAULT_TIMEOUT, TimeUnit.MILLISECONDS)
             .readTimeout(DEFAULT_TIMEOUT, TimeUnit.MILLISECONDS)
             .protocols(Arrays.asList(Protocol.HTTP_1_1))
+            .addInterceptor(
+                new StatusRetryInterceptor(
+                    5,
+                    ImmutableSet.of(
+                        408, // Request Timeout
+                        429, // Too Many Requests
+                        499, // Client Closed Request (nginx)
+                        500, // Internal Server Error
+                        502, // Bad Gateway
+                        503, // Service Unavailable
+                        504, // Gateway Timeout
+                        520), // Cloudflare unknown error
+                    1000))
             .build();
     try {
       return enableExternalCertificatesFromEnv(client);
```

allanrogerr: Thank you for the suggestion — an OkHttp interceptor is an appealing simplification for the HTTP-status case. The reason the current implementation uses a `CompletableFuture`-based wrapper instead is that the interceptor approach has three gaps that would be regressions for this PR:

**1. S3 XML error codes are not visible at the interceptor layer.**
S3 errors like `SlowDown`, `ThrottlingException`, `InternalError`, `ExpiredToken` etc. arrive as HTTP 200 responses with an XML body — the status code is always 200. An interceptor keyed on `response.code()` never sees these; they must be classified after parsing the body in `throwMinioException()`, which is downstream of OkHttp.

**2. `Thread.sleep()` inside an interceptor blocks an OkHttp dispatcher thread.**
OkHttp's async dispatcher has a bounded thread pool (default 64). Sleeping inside `Callback.onResponse` / an interceptor for exponential backoff (up to ~31 s at 5 retries × 1 000 ms base) would hold a dispatcher thread for the full delay, starving other concurrent requests. The current design schedules the next attempt on `RETRY_SCHEDULER` and releases the dispatcher thread immediately.

**3. Non-seekable request bodies cannot be replayed via `chain.proceed(request)`.**
`RequestBody.writeTo()` is called once; OkHttp does not buffer or reset it. For multipart uploads or streamed PUT bodies, a second `chain.proceed` would send an empty or corrupt body. The current wrapper gates retry on the body's seekability before re-issuing the request.

If you'd prefer to move the HTTP-status subset into an interceptor and keep only S3-code + IOException handling in the wrapper, I'm happy to split it that way — but I wanted to flag the above before making that change.

balamurugana: > 1. S3 XML error codes are not visible at the interceptor layer.

No S3 error is returned with 2xx HTTP status code. If so, it is a bug at server side. As @harshavardhana mentioned previously we do not need to use S3 error code instead we need to use HTTP status code.

> 2. Thread.sleep() inside an interceptor blocks an OkHttp dispatcher thread.

This is fine. If the server is busy, why send new requests and wait in another thread. New request to the same server is blocked if `maxRequestsPerHost = 5` is hit due to retry sleep. Instead of 1s, we need to start the sleep with 100ms or lesser.

3. Non-seekable request bodies cannot be replayed via chain.proceed(request).

You don't need to. We have retry-able custom [RequestBody](https://minio-java.min.io/io/minio/Http.RequestBody.html). It works for `ByteBuffer`, `RandomAccessFile` and `byte[]`.

allanrogerr: Done in 34abd647 — retry is now an OkHttp interceptor (`Http.RetryInterceptor`) keyed on HTTP status, with IOException retry preserved. Dropped `RETRY_SCHEDULER`, `executeWithRetry`, and the `RETRYABLE_S3_CODES` set. Non-replayable raw `okhttp3.RequestBody` is sent once via `Http.RequestBody.isReplayable()`. Net: -196 lines.

balamurugana: Please revert unwanted changes

api/src/main/java/io/minio/Http.java
```diff
@@ -683,10 +686,57 @@ public static OkHttpClient setTimeout(
         .build();
   }
 
+  /**
+   * OkHttp interceptor that retries requests on retryable HTTP status codes with exponential
+   * backoff and full jitter.
+   */
+  public static class RetryInterceptor implements Interceptor {
```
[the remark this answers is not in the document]
allanrogerr: Addressed in 756e294. Kept `RetryInterceptor` public — it is intentionally part of the SDK's supported API surface so callers can register it on a stand-alone `OkHttpClient` (`BaseS3Client` already installs it on every supplied client via `wrapWithRetry`). The class Javadoc now declares this explicitly and drops the `{@link Retry#…}` references that pointed at package-private symbols, so the public Javadoc no longer leaks internal links.

api/src/main/java/io/minio/Http.java
```diff
@@ -683,10 +686,57 @@ public static OkHttpClient setTimeout(
...
+        // Cap exponent to avoid bit-shift overflow at high attempt counts.
+        int exp = Math.min(tryCount, 30);
+        long backoffCap = Math.min(MAX_DELAY_MS, BASE_DELAY_MS * (1L << exp));
+        long jittered = ThreadLocalRandom.current().nextLong(0, Math.max(1, backoffCap));
+        try {
+          Thread.sleep(jittered);
+        } catch (InterruptedException ie) {
+          Thread.currentThread().interrupt();
+          throw new IOException("retry interrupted", ie);
+        }
```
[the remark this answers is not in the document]
allanrogerr: Documented in 756e294. The synchronous-sleep design is deliberate: the previous iteration of this PR used a separate `ScheduledExecutorService` and was reverted to a plain OkHttp interceptor per balamurugana's proposal (see ba65312 → 6d39a12) precisely to keep the orchestration in one self-contained interceptor. Re-introducing an external scheduler would re-open that complexity. The `RetryInterceptor` Javadoc now carries an explicit `Threading` note that backoff sleeps on the OkHttp dispatcher slot and points high-concurrency callers at `Dispatcher.setMaxRequests` / `setMaxRequestsPerHost` for sizing — that is the supported escape hatch for retry-storm scenarios.

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -268,7 +266,10 @@ private String[] handleRedirectResponse(
     return new String[] {code, message};
   }
 
-  /** Execute HTTP request asynchronously for given parameters. */
+  /**
+   * Execute HTTP request asynchronously for given parameters. Retries on retryable HTTP status
+   * codes are handled by {@link Http.RetryInterceptor} installed on the default HTTP client.
```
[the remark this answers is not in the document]
allanrogerr: Reworded in 756e294. The Javadoc now correctly reflects `BaseS3Client.wrapWithRetry`: the interceptor is installed on every supplied `OkHttpClient` (default or caller-supplied), and the `maxRetries` supplier is read on every request. The "only on the default client" wording is gone.

api/src/test/java/io/minio/RetryTest.java
```diff
@@ -0,0 +1,89 @@
...
+      server.enqueue(retryableServerError());
+      server.enqueue(successResponse());
+      server.start();
+
+      MinioClient client = MinioClient.builder().endpoint(server.url("").toString()).build();
+      client.listBuckets();
+
+      Assert.assertEquals(2, server.getRequestCount());
+    }
+  }
```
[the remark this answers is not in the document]
allanrogerr: Strengthened in 756e294. `testRetryExhaustedReturnsLastResponse` now asserts the terminal `InvalidResponseException` carries `Response code: 500` rather than letting any exception type slip through. Added `testRetryExhaustedSurfacesXmlErrorResponse` for the XML/`ErrorResponseException` path, which asserts both `e.response().code() == 503` and `e.errorResponse().code() == "InternalError"` after retries are exhausted, plus the request count matches `maxRetries`.

api/src/main/java/io/minio/errors/InvalidResponseException.java
allanrogerr: Done. The overall aim was to implement retry as in minio-go

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -113,19 +111,56 @@ public abstract class BaseS3Client implements AutoCloseable {
...
+   */
+  private OkHttpClient wrapWithRetry(OkHttpClient client) {
+    OkHttpClient.Builder builder = client.newBuilder().retryOnConnectionFailure(false);
+    builder.interceptors().removeIf(i -> i instanceof Http.RetryInterceptor);
+    return builder.addInterceptor(new Http.RetryInterceptor(() -> this.maxRetries)).build();
+  }
+
+  /**
+   * Sets the maximum number of attempts for transient HTTP failures. Pass {@code 1} to disable
+   * automatic retries. Defaults to {@link Retry#MAX_RETRY}.
```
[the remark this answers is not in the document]
allanrogerr: Addressed in 7fe6e87. Replaced `{@link Retry#MAX_RETRY}` with the literal `{@code 10}` so generated public Javadoc no longer points at a package-private symbol.

api/src/main/java/io/minio/MinioAsyncClient.java
```diff
@@ -204,19 +206,39 @@ public Builder credentialsProvider(Provider provider) {
...
     public Builder httpClient(OkHttpClient httpClient, boolean close) {
       Utils.validateNotNull(httpClient, "http client");
       this.httpClient = httpClient;
       this.closeHttpClient = close;
       return this;
     }
 
+    /**
+     * Sets the maximum number of attempts per request. Pass {@code 1} to disable automatic retries.
+     * Defaults to {@link Retry#MAX_RETRY}.
```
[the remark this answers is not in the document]
allanrogerr: Addressed in 7fe6e87. Same fix as the `setMaxRetries` thread — `Builder.maxRetries` Javadoc now uses the literal `{@code 10}` instead of `{@link Retry#MAX_RETRY}`.

api/src/main/java/io/minio/MinioAsyncClient.java
```diff
@@ -204,19 +206,39 @@ public Builder credentialsProvider(Provider provider) {
       return this;
     }
 
+    /**
+     * Supplies a custom {@link OkHttpClient}. The SDK wraps this client to install {@link
+     * Http.RetryInterceptor} and forces {@code retryOnConnectionFailure(false)} so that {@link
+     * Http.RetryInterceptor} owns the entire retry policy. Any prior {@code RetryInterceptor} on
+     * the supplied client is replaced.
```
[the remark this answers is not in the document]
allanrogerr: Addressed in 7fe6e87. `wrapWithRetry` now also calls `removeIf` on `builder.networkInterceptors()`, so any prior `RetryInterceptor` is stripped from both chains before the new one is installed as an application interceptor. The Javadoc on `Builder.httpClient` and on `wrapWithRetry` was tightened to say so explicitly.

docs/API.md
```diff
@@ -71,6 +71,7 @@ MinIO Client Builder is used to create MinIO client. Builder has below methods t
 | `credentials()` | Accepts access key (aka user ID) and secret key (aka password) of an account in S3 service.                                                |
 | `region()`      | Accepts region name of S3 service. If specified, all operations use this region otherwise region is probed per bucket.                     |
 | `httpClient()`  | Custom HTTP client to override default.                                                                                                    |
+| `maxRetries()`  | Maximum number of attempts per request for transient HTTP failures (retryable status codes 408/429/499/500/502/503/504/520, retryable S3 codes such as `SlowDown`/`InternalError`/`ExpiredToken`, and retryable IOExceptions). Pass `1` to disable automatic retries. Defaults to `10`. |
```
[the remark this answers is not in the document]
allanrogerr: PR description updated to match the implementation: defaults to 10 attempts (not 5), public `maxRetries()` / `setMaxRetries()` API surface is documented, `retryOnConnectionFailure(false)` + SDK-level IOException classification is called out, and the retryable-S3-code list is included alongside the HTTP status set. Also added a minio-go alignment table to make the contract explicit. No code change here.

api/src/main/java/io/minio/Retry.java
```diff
@@ -0,0 +1,132 @@
...
+   *
+   * <p>With {@code MAX_JITTER == 1.0}, returns a uniform random value in {@code [0, min(cap, base *
+   * 2^attempt)]}.
+   */
+  static long exponentialBackoffMs(int attempt) {
+    int exp = Math.min(Math.max(attempt, 0), 30);
+    long sleep = DEFAULT_RETRY_UNIT_MS * (1L << exp);
+    if (sleep > DEFAULT_RETRY_CAP_MS) sleep = DEFAULT_RETRY_CAP_MS;
+    sleep -= (long) (ThreadLocalRandom.current().nextDouble() * (double) sleep * MAX_JITTER);
+    return Math.max(0L, sleep);
```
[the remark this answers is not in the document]
allanrogerr: Documented in 7fe6e87. The Javadoc on `exponentialBackoffMs` now states the actual range `[1, min(cap, base · 2^attempt)]` and explains why: `ThreadLocalRandom.nextDouble()` is in `[0.0, 1.0)` and the `(long)` cast truncates `rand * sleep` to at most `sleep - 1`, so the subtraction never reaches `sleep`. Kept the implementation as-is because minio-go's `exponentialBackoffWait` uses the same formula (Go's `Float64()` is also `[0.0, 1.0)`) — changing the formula to allow 0 ms would deviate from the minio-go contract this class explicitly mirrors. The existing tests already assert `delay >= 0`, which remains true.

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -113,19 +111,59 @@ public abstract class BaseS3Client implements AutoCloseable {
...
+   */
+  volatile int maxRetries = Retry.MAX_RETRY;
+
   protected BaseS3Client(
       Http.BaseUrl baseUrl, Provider provider, OkHttpClient httpClient, boolean closeHttpClient) {
     this.baseUrl = baseUrl;
     this.provider = provider;
-    this.httpClient = httpClient;
     this.closeHttpClient = closeHttpClient;
+    this.httpClient = wrapWithRetry(httpClient);
```
balamurugana: For custom httpclient, we should not `wrapWithRetry()`. It is user's choice.

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -113,19 +111,59 @@ public abstract class BaseS3Client implements AutoCloseable {
...
   }
 
   protected BaseS3Client(BaseS3Client client) {
     this.baseUrl = client.baseUrl;
     this.provider = client.provider;
-    this.httpClient = client.httpClient;
     this.closeHttpClient = client.closeHttpClient;
+    this.maxRetries = client.maxRetries;
+    this.httpClient = wrapWithRetry(client.httpClient);
```
balamurugana: Same as above

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -282,15 +331,9 @@ protected CompletableFuture<Response> executeAsync(Http.S3Request s3request, Str
     PrintWriter traceStream = this.traceStream;
     if (traceStream != null) traceStream.print(request.httpTraces());
 
-    OkHttpClient httpClient = this.httpClient;
```
balamurugana, line 285: `ignoreCertCheck()` and `setTimeout()`  modifies the current `httpClient`, you have to take a copy of current `httpClient` and work on it.

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -469,9 +512,10 @@ private void onResponse(final Response response) throws IOException {
                           response.header("x-amz-id-2"));
                 }
 
-                // invalidate region cache if needed
-                if (errorResponse.code().equals(NO_SUCH_BUCKET)
-                    || errorResponse.code().equals(RETRY_HEAD)) {
+                // invalidate region cache if needed (bucket may be null for e.g. listBuckets)
```
balamurugana: Why `listBuckets()` supposes to send `null` bucket? server side bug? This comment is misleading. Any S3 call to a bucket may have invalid region cache due to the bucket removal from one region and created in another region.

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -469,9 +512,10 @@ private void onResponse(final Response response) throws IOException {
...
+                if (s3request.bucket() != null
```
balamurugana: This is wrong. We would on based on error, not by input.

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -1223,6 +1267,15 @@ private Object[] createBody(PutObjectAPIBaseArgs args, MediaType contentType)
...
     String md5Hash = headers.getFirst(Http.Headers.CONTENT_MD5);
 
+    long fileStartPos = 0;
+    if (args.file() != null) {
+      try {
+        fileStartPos = args.file().getFilePointer();
+      } catch (IOException e) {
+        throw new MinioException(e);
+      }
+    }
```
balamurugana: Why do we need?

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -1278,6 +1331,14 @@ private Object[] createBody(PutObjectAPIBaseArgs args, MediaType contentType)
       }
     }
 
+    if (args.file() != null) {
+      try {
+        args.file().seek(fileStartPos);
+      } catch (IOException e) {
+        throw new MinioException(e);
+      }
+    }
```
balamurugana: Why do we need?

api/src/main/java/io/minio/Http.java
```diff
@@ -620,6 +623,10 @@ public static OkHttpClient newDefaultClient() {
             .writeTimeout(DEFAULT_TIMEOUT, TimeUnit.MILLISECONDS)
             .readTimeout(DEFAULT_TIMEOUT, TimeUnit.MILLISECONDS)
             .protocols(Arrays.asList(Protocol.HTTP_1_1))
+            // Our RetryInterceptor handles transient failures with full-jitter backoff; defer
```
balamurugana: Remove this comment. It is obvious.

api/src/main/java/io/minio/Http.java
```diff
@@ -683,10 +690,163 @@ public static OkHttpClient setTimeout(
...
+   * maxRetries} values combined with high backoff caps could outlast either the signing window or a
+   * short-lived STS credential. minio-go's request-level retry rebuilds and re-signs each attempt;
+   * that semantic is intentionally not replicated here in exchange for the simpler interceptor
+   * model.
+   *
+   * <p>To opt out of retries on a stand-alone {@link OkHttpClient}, simply do not register this
+   * interceptor.
+   */
+  public static class RetryInterceptor implements Interceptor {
+    /** Maximum body bytes inspected when probing for an S3 {@code <Code>} value. */
```
balamurugana: We don't want to probe S3 error codes and we should not do this.

api/src/main/java/io/minio/Http.java
```diff
@@ -695,9 +855,16 @@ public static class Body {
...
+     * Creates Body for okhttp3 RequestBody.
+     *
+     * <p><b>Retry caveat.</b> {@link Http.RetryInterceptor} re-invokes {@code writeTo} on each
+     * retry. Pass only replayable bodies; do not pass a body that overrides {@code isOneShot()} to
+     * return {@code true} (or otherwise consumes its source on first write) unless retries are
+     * disabled for that call.
+     */
     public Body(okhttp3.RequestBody requestBody) {
-      this.requestBody = requestBody;
+      this.requestBody = Utils.validateNotNull(requestBody, "request body");
```
balamurugana: No need to throw own exception message. Let it fail.

api/src/main/java/io/minio/Http.java
```diff
@@ -710,6 +877,11 @@ public Body(
         String md5Hash) {
       if (length < 0) throw new IllegalArgumentException("valid length must be provided");
       this.file = file;
+      try {
+        this.fileOffset = file.getFilePointer();
```
balamurugana: Why do we need?

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -92,7 +90,6 @@ public abstract class BaseS3Client implements AutoCloseable {
       "ServerSideEncryptionConfigurationNotFoundError";
   // maximum allowed bucket policy size is 20KiB
   protected static final int MAX_BUCKET_POLICY_SIZE = 20 * 1024;
-  protected static final Random RANDOM = new Random(new SecureRandom().nextLong());
```
balamurugana, line 95: Do not remove `protected`, `public` things.

api/src/main/java/io/minio/Retry.java
```diff
@@ -0,0 +1,136 @@
+/*
```
balamurugana: Why do we need?

docs/API.md
balamurugana, line 74: `10` is too much of retry. `5` is similar to `minio-py`.

api/src/main/java/io/minio/BaseS3Client.java
allanrogerr: If the issue is opting out, the user can do this by setting `.maxRetries(1)`

The current code was done in the spirit of following minio-go, which seems to be the gold standard based on the Slack discussion. minio-go does not wrap the httpClient; retry is applied at a different layer. It also follows the rule that `MaxRetries=1` means no retry. 

What are your thoughts on this? @balamurugana

api/src/main/java/io/minio/BaseS3Client.java
allanrogerr: See above

api/src/main/java/io/minio/BaseS3Client.java
allanrogerr: setTimeout and disableCertCheck use a builder pattern to make a new OkHttpClient instance. The original is not mutated. Im not following your comment @balamurugana

api/src/main/java/io/minio/BaseS3Client.java
allanrogerr: Removing this. This is not achievable it seems. A NPE guard is not needed

api/src/main/java/io/minio/BaseS3Client.java
allanrogerr: Removing this. This is not achievable it seems. A NPE guard is not needed

api/src/main/java/io/minio/BaseS3Client.java
allanrogerr: Each file.read(...) advances the file pointer. After hashing a length-byte file, the pointer is at fileStartPos + length. So whatever position the pointer is at when new Http.Body(...) runs becomes the offset that subsequent retries seek back to before each network attempt. Without the restore at line 1335, that captured offset would be fileStartPos + length instead of fileStartPos, and every PUT — first attempt and retries alike — would write from the wrong offset (typically EOF, producing an empty or truncated upload). Http.Body doesnt handle this.

minio-go does this in api-put-object-streaming.go:682-705

api/src/main/java/io/minio/BaseS3Client.java
allanrogerr: Responded above

api/src/main/java/io/minio/Http.java
allanrogerr: Done

api/src/main/java/io/minio/Http.java
allanrogerr: This is done in minio-go. api.go:712 and api.go:800 both feed errResponse.Code into isS3CodeRetryable(...) and continue the retry loop on a match. So minio-go does probe S3 error codes. Not doing this would cause a divergence in the SDK behaviours. 
Your thoughts @balamurugana ?

api/src/main/java/io/minio/Http.java
allanrogerr: Done

api/src/main/java/io/minio/Http.java
allanrogerr: We dont - since we are keeping createBody save/restore (BaseS3Client.java:1270-1276 and 1334-1340) and RequestBody.position capture and writeTo seek (Http.java:1006, 1053)

Removing

api/src/main/java/io/minio/Retry.java
allanrogerr: Retry.java is a technically correct, direct port from minio-go. Removing it would disrupt that alignment.

docs/API.md
allanrogerr: 10 is a direct port from minio-go.

api/src/main/java/io/minio/BaseS3Client.java
allanrogerr: Reverted

api/src/main/java/io/minio/Retry.java
balamurugana: This doesn't explain why it is needed.

api/src/main/java/io/minio/BaseS3Client.java
balamurugana: It is an user choice. We should not wrap.

api/src/main/java/io/minio/BaseS3Client.java
balamurugana: 1. `this.httpClient = Http.setTimeout(this.httpClient, connectTimeout, writeTimeout, readTimeout);` is not thread safe. This is a technical.
2. A `execute()` call should work consistently internal with retry i.e. no two different timeout between them. This is logical.

api/src/main/java/io/minio/BaseS3Client.java
balamurugana: It is not needed for Java. Please check `RandomAccessFile` and `Http.RequestBody`

api/src/main/java/io/minio/Http.java
balamurugana: Not needed here. The latest approach is to deal with HTTP status code only.

api/src/main/java/io/minio/MinioAsyncClient.java
```diff
@@ -93,6 +93,7 @@
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.ThreadLocalRandom;
```
balamurugana: We already have `RANDOM`. We don't need.

api/src/main/java/io/minio/BaseS3Client.java
allanrogerr: Ok. I have reimplemented this to respect the choice.

api/src/main/java/io/minio/BaseS3Client.java
allanrogerr: Removing this. in order to do so, Checksum.update's RandomAccessFile branch must be rewritten to use FileChannel.read(ByteBuffer, long position) so it no longer mutates the file's pointer. 
PTAL @balamurugana

api/src/main/java/io/minio/Http.java
allanrogerr: Done

api/src/main/java/io/minio/MinioAsyncClient.java
allanrogerr: Removed

api/src/main/java/io/minio/Retry.java
allanrogerr: Strictly speaking none of this is needed. I am seeking to maintain alignment with minio-go as much as possible. It is a single source of truth, with pure-function testability, policy versus mechanism separation, as well as parity with minio-go/retry.go

Perhaps more substantive: it is the SDK's only retry policy: with an HTTP-status set, IOException classifier, and jitter formula - none of which OkHttp provides. They are kept in one package-private class so the synch path and the interceptor can't drift, and so the predicates are unit-testable as pure functions.

As an alternative I could inline the four policy pieces as private statics inside Http.RetryInterceptor. What do you prefer @balamurugana ?

api/src/main/java/io/minio/BaseS3Client.java
allanrogerr: 1. i am making httpClient volatile to ensure thread safety
2. This is already considered because of the builder pattern, the OkHttpClient instance is essentially immutable. The only way to "change the timeout" is to build a new instance and reassign whatever pointer was holding the original. The original itself remains exactly what it was.

allanrogerr:
All comments addressed

api/src/main/java/io/minio/Retry.java
balamurugana: Please remove it. We should not copy/convert go code. We need to use each language features.

allanrogerr: Spoke to Bala - I am removing this entire file and refactoring to suit.

allanrogerr: done

balamurugana:
Addressed in https://github.com/minio/minio-java/pull/1702

PR #1702, opened 2026-05-11
Add retry support for HTTP execution failure

balamurugana:


api/src/main/java/io/minio/Http.java
```diff
@@ -85,6 +90,21 @@ public class Http {
...
+  public static final Set<Integer> RETRIABLE_STATUS_CODES =
+      ImmutableSet.of(
+          408, // Request Timeout
+          429, // Too Many Requests
+          499, // Client Closed Request (nginx)
+          500, // Internal Server Error
+          502, // Bad Gateway
+          503, // Service Unavailable
+          504, // Gateway Timeout
+          520); // Cloudflare unknown error
```
shtripat, line 103: should we use `http.StatusTooManyRequests` kind values instead of hardcoded ones like 429?

balamurugana: okhttp does not have those enums? `java.net.HttpURLConnection` has limited constants. I think we can avoid mixed usage.

shtripat: ok

shtripat (dismissed):
lgtm

api/src/main/java/io/minio/Http.java
```diff
@@ -608,6 +628,171 @@ public static OkHttpClient enableExternalCertificatesFromEnv(OkHttpClient client
...
+    String encodedPath = request.url().encodedPath();
+    String encodedQuery = request.url().encodedQuery();
+    if (encodedQuery != null) encodedPath += "?" + encodedQuery;
+    traceBuilder.append(method.toString()).append(" ").append(encodedPath).append(" HTTP/1.1\n");
+    traceBuilder.append(
+        request
+            .headers()
+            .toString()
+            .replaceAll("Signature=([0-9a-f]+)", "Signature=*REDACTED*")
+            .replaceAll("Credential=([^/]+)", "Credential=*REDACTED*"));
```
allanrogerr: These could be stored as static final Pattern to avoid recompilation every trace event

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -137,6 +132,46 @@ public void close() {
...
+   */
+  public void setRetry(Set<Integer> retryStatusCodes, Long delayMs, Integer maxRetries) {
+    Interceptor interceptor =
+        new Http.StatusRetryInterceptor(
+            retryStatusCodes, delayMs == null ? 0 : delayMs, maxRetries == null ? 0 : maxRetries);
+
+    List<Interceptor> interceptors = this.httpClient.interceptors();
+    int i = 0;
+    boolean found = false;
+    for (i = 0; i < interceptors.size(); i++) {
```
allanrogerr: Any reason for re-reinitialising?

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -268,25 +303,59 @@ private String[] handleRedirectResponse(
     return new String[] {code, message};
   }
 
+  private OkHttpClient getHttpClient(PrintWriter traceStream, Http.S3Request s3request) {
+    if (traceStream == null) return this.httpClient;
+
+    OkHttpClient httpClient = this.httpClient;
+    List<Interceptor> interceptors = httpClient.interceptors();
```
allanrogerr, line 310: This look very close to the code on line 151. Any reason for the duplication? This could be refactored out as a method.

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -137,6 +132,46 @@ public void close() {
     }
   }
 
+  /**
+   * Sets request retry parameters. Any null/invalid values disable retry.
+   *
+   * <pre>Example:{@code
+   * minioClient.setRetryParameters(ImmutableSet.of(408, 504), 250, 3);
```
allanrogerr:
```suggestion
   * minioClient.setRetry(ImmutableSet.of(408, 504), 250, 3);
```

allanrogerr (commented):
The new retry logic including interceptor, setRetry replacement logic, body-replay across retries, NPE-safe trace formatting are not unit-tested. The functional tests cover a happy path, but 
- no test creates a scenario to verify retry actually fires, 
- no test verifies the maxRetries boundary, 
- no test checks setRetry replaces the interceptor correctly, 
- no test exercises a malformed request.method()
Could you have Claude generate some tests please?

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -268,25 +303,59 @@ private String[] handleRedirectResponse(
...
 
-    OkHttpClient httpClient = this.httpClient;
-    // FIXME: enable retry for all request.
-    // if (!s3request.retryFailure()) {
-    //   httpClient = httpClient.newBuilder().retryOnConnectionFailure(false).build();
-    // }
+    StringBuilder traceBuilder = new StringBuilder(request.httpTraces());
+    if (!statusRetryInterceptorFound && traceStream != null) {
+      traceStream.print(request.httpTraces());
+    }
```
allanrogerr: Why not static final if it is never reassigned? Also all branches using this boolean are unreachable. Perhaps there is an unfinished idea here?

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -137,6 +132,46 @@ public void close() {
...
+    if (found) {
+      builder.interceptors().clear();
+      for (int j = 0; j < interceptors.size(); j++) {
+        builder.addInterceptor(i == j ? interceptor : interceptors.get(j));
+      }
+    } else {
+      builder.addInterceptor(interceptor);
+    }
+
+    this.httpClient = builder.build();
```
allanrogerr, line 172: This seems to have the makings of a race:

1. read: List<Interceptor> interceptors = this.httpClient.interceptors();
2. read: OkHttpClient.Builder builder = this.httpClient.newBuilder();
3. write: this.httpClient = builder.build();

Two concurrent setRetry calls can read and assign `protected volatile OkHttpClient httpClient`. This may not be ideal.

api/src/main/java/io/minio/Http.java
```diff
@@ -886,6 +1115,15 @@ public static enum Method {
     POST,
     PUT,
     DELETE;
+
+    public static Method fromString(String value) {
+      if (value == null) return null;
+      try {
+        return Method.valueOf(value.toUpperCase(Locale.US).trim());
+      } catch (IllegalArgumentException e) {
+        return null;
```
allanrogerr, line 1130: Invocations of the method may hit NPE - e.g. `traceBuilder.append(method.toString())`

allanrogerr (commented):
Additional comments @balamurugana  PTAL

api/src/main/java/io/minio/Http.java
balamurugana: This is OkHttp style. If you pass `null` or invalid, it returns `null` which is correct. it is user's responsibility to use it correctly.

api/src/main/java/io/minio/Http.java
balamurugana: done

api/src/main/java/io/minio/BaseS3Client.java
balamurugana: done

api/src/main/java/io/minio/BaseS3Client.java
balamurugana: This method does updates `StatusRetryInterceptor` if available else adds it to default/custom `httpClient`.

api/src/main/java/io/minio/BaseS3Client.java
balamurugana: This method is `synchronized` now.

api/src/main/java/io/minio/BaseS3Client.java
balamurugana: fixed

api/src/main/java/io/minio/BaseS3Client.java
balamurugana: removed

allanrogerr:
Rechecking

api/src/test/java/io/minio/MinioClientTest.java
```diff
@@ -584,4 +586,28 @@ public void testMakeBucketRegionConflicts()
     client.makeBucket(MakeBucketArgs.builder().bucket("mybucket").region("us-west-2").build());
     Assert.fail("exception should be thrown");
   }
+
+  @Test
```
allanrogerr, line 592: Could you add the tests:
- maxRetries boundary (no test exhausts retries and asserts the terminal error)
- setRetry replacement (swap an existing interceptor),
- for a malformed request.method()

api/src/main/java/io/minio/Http.java
allanrogerr: The user will also hit NPE in an unexpected way. e.g. consider the legal `method("OPTIONS", null)`
```
api/src/test/java/io/minio/RetryTraceNpeTest.java
```
```
package io.minio;

import okhttp3.Interceptor;
import okhttp3.OkHttpClient;
import org.junit.Test;

public class RetryTraceNpeTest {
  @Test(expected = NullPointerException.class)
  public void traceOnNpeOnNonS3Method() throws Exception {
    Interceptor verbRewriter =
        chain -> {
          okhttp3.Request r = chain.request();
          return chain.proceed(r.newBuilder().method("OPTIONS", null).build());
        };

    OkHttpClient client = new OkHttpClient.Builder().addInterceptor(verbRewriter).build();

    MinioClient minio =
        MinioClient.builder()
            .endpoint("https://play.min.io")
            .credentials("Q3AM3UQ867SPQQA43P2F", "zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG")
            .httpClient(client)
            .build();
    minio.traceOn(System.out);

    try {
      minio.listBuckets();
    } catch (Throwable t) {
      for (Throwable c = t; c != null; c = c.getCause()) {
        if (c instanceof NullPointerException) throw (NullPointerException) c;
      }
      throw t;
    }
  }
}
```
```
docker run --rm \
-u "$(id -u):$(id -g)" \
-v "$HOME":"$HOME" \
-v /home/ubuntu/github/balamurugana/minio-java:/work \
-w /work \
-e HOME="$HOME" \
-e GRADLE_USER_HOME="$HOME/.gradle" \
eclipse-temurin:17-jdk \
./gradlew :api:test --tests "io.minio.RetryTraceNpeTest" --rerun-tasks -i 2>&1 
```
```
java.lang.NullPointerException: Cannot invoke "io.minio.Http$Method.toString()" because "method" is null
	at io.minio.Http.getRequestTraces(Http.java:642)
	at io.minio.Http$StatusRetryInterceptor.intercept(Http.java:770)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:126)
	at io.minio.RetryTraceNpeTest.lambda$traceOnNpeOnNonS3Method$0(RetryTraceNpeTest.java:13)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:126)
	at okhttp3.internal.connection.RealCall.getResponseWithInterceptorChain$okhttp(RealCall.kt:226)
	at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:574)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:840)
```

allanrogerr: This could be avoidable by a simple null guard e.g.
```
--- a/api/src/main/java/io/minio/Http.java
+++ b/api/src/main/java/io/minio/Http.java
-    traceBuilder.append(method.toString()).append(" ").append(encodedPath).append(" HTTP/1.1\n");
+    traceBuilder.append(method == null ? request.method() : method.toString()).append(" ").append(encodedPath).append(" HTTP/1.1\n");
```

allanrogerr:
Just a couple missing tests if you dont mind and an NPE guard. Thank you for this solution @balamurugana !

api/src/main/java/io/minio/Http.java
balamurugana: This never happens in `minio-java`. No S3/MinIO Admin API use `OPTIONS`. That's the reason `enum Method` does not have it.

api/src/test/java/io/minio/MinioClientTest.java
balamurugana: I added test cover 1 and 2. Case 3 is not applicable for `minio-java`

balamurugana:
> Just a couple missing tests if you dont mind and an NPE guard. Thank you for this solution @balamurugana !

@allanrogerr Updated the PR. Thanks!

allanrogerr (approved):

allanrogerr:
@shtripat @rraulinio PTAL

shtripat (approved):
lgtm

PR #1703, opened 2026-05-22
Retry CompleteMultipartUpload and DeleteObjects on in-body errors

harshavardhana:
Some S3 APIs return HTTP 200 OK but report transient failures inside the XML body. The existing retry interceptor only looks at status codes, so these slip through.

This adds a small set of retriable S3 error codes (InternalError, SlowDown, ServiceUnavailable, RequestTimeout, Throttling, ThrottlingException) and reuses the same backoff/budget already configured for status-code retries.

CompleteMultipartUpload retries the whole request when the 200-OK body is an <Error> with a retriable code. DeleteObjects retries the whole batch when any per-object error is retriable — delete is idempotent so re-sending already-deleted keys is safe, and non-retriable errors still surface as before.


balamurugana:
Already addressed by https://github.com/minio/minio-java/pull/1704

PR #1709, opened 2026-06-20
Move to JDK 25 for build only

balamurugana:
## Summary by CodeRabbit

* **Chores**
  * Updated CI/build to use the latest repository checkout action and refreshed functional-test tooling versions; adjusted functional-test behavior on Windows vs Ubuntu.
  * Enabled builds with Java 25 while keeping Java 8-compatible output, and updated Gradle wrapper to  9.7.1.
  * Refreshed build plugins/dependencies, signing configuration, and artifact handling; updated Git line-ending rules for binaries.
* **Style**
  * Applied formatting-only updates across a few implementation and test call sites.
* **Documentation**
  * Improved Javadoc inline formatting for the HTTP body type.


shtripat (dismissed):
lgtm

.github/workflows/gradle.yml
```diff
@@ -29,11 +29,11 @@ jobs:
             exit 1
           fi
 
-      - name: Setup java 17 for building
+      - name: Setup java 25 for building
         uses: actions/setup-java@v5
```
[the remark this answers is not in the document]
balamurugana: This is indented behavior. The CI should use the latest `v5` for every run.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,7 +1,9 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-9.4.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.6.1-bin.zip
 networkTimeout=10000
+retries=0
+retryBackOffMs=500
```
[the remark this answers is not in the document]
balamurugana: This is intentional for testing and github workflow

spotbugs-filter.xml
```diff
@@ -32,4 +32,8 @@
   <Match>
     <Bug pattern="DMI_DOH"/>
   </Match>
+  <!-- Intentional: subclass *Args override validate* methods as public for the public API surface. -->
+  <Match>
+    <Bug pattern="IAOM_DO_NOT_INCREASE_METHOD_ACCESSIBILITY"/>
+  </Match>
```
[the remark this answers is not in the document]
balamurugana: This is intentional.

build.gradle
```diff
@@ -103,6 +107,10 @@ subprojects {
     check.dependsOn localeTest
 
     java {
+        // Compile/test/javadoc run on JDK 25; bytecode/API level pinned to Java 8 via release below.
+        toolchain {
+            languageVersion = JavaLanguageVersion.of(25)
+        }
```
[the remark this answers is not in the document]
balamurugana: Fixed.

Arubibar:
this version fix the jacskon vulnerailities?

shtripat (dismissed):

functional/TestUserAgent.java
```diff
@@ -17,27 +17,38 @@
...
-    client.bucketExists(BucketExistsArgs.builder().bucket("any-bucket-name-works").build());
-    client.traceOff();
+    // try-with-resources closes the client so its OkHttp dispatcher/connection-pool threads shut
+    // down and the JVM exits promptly instead of waiting out the idle keep-alive.
+    try (MinioClient client = MinioClient.builder().endpoint("http://httpbin.org").build()) {
+      client.setRetry(null, null, null);
+      client.traceOn(baos);
+      try {
+        client.bucketExists(BucketExistsArgs.builder().bucket("any-bucket-name-works").build());
+      } catch (MinioException e) {
```
anjalshireesh, line 35: `catch (MinioException e)` won't catch a connection failure - those surface as `IllegalStateException` from `MinioAsyncClient.java:3700` and escape `main`. Can we make it `catch (Exception e)`?

build.gradle
```diff
@@ -168,6 +176,16 @@ project(':api') {
         dependsOn tasks.named('shadowJar')
     }
 
+    tasks.named('shadowJar', com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar) {
+        // okhttp/okio/kotlin-stdlib each ship a distinct META-INF/*.kotlin_module marker file.
+        // Shadow's KotlinModuleMetadataTransformer expects to see every duplicate, but the
+        // default EXCLUDE strategy would drop them beforehand; WARN lets the transformer run
+        // and still logs if a real duplicate ever shows up (gradleup.com/shadow/configuration/merging).
+        filesMatching('META-INF/*.kotlin_module') {
```
anjalshireesh, line 185: Is this needed? The published `minio:9.0.3-all` already has all five `META-INF/*.kotlin_module` entries without it, and they're distinct paths so there's no duplicate for `duplicatesStrategy` to act on. If it was precautionary, this and the copy at `build.gradle:314` could go.

anjalshireesh (dismissed):
LGTM, couple of minor suggestions

build.gradle
```diff
@@ -47,23 +47,27 @@ subprojects {
...
-        testImplementation 'org.junit.jupiter:junit-jupiter-api:5.14.3'
+        testImplementation 'com.squareup.okhttp3:mockwebserver:5.5.0'
+        testImplementation 'org.junit.jupiter:junit-jupiter-api:5.14.4'
     }
 
     [compileJava, compileTestJava].each() {
         it.options.fork = true
+        // Build runs on JDK 25 (see java.toolchain) but enforces the Java 8 API surface and
+        // emits Java 8 bytecode via --release 8. --release also avoids the "source/target 8 is
+        // obsolete" warning; -Xlint:-options is kept as a belt-and-braces guard for -Werror.
```
allanrogerr: `--release 8` doesn't actually suppress the obsolete-source warning: JDK 25's javac still prints `source value 8 is obsolete and will be removed in a future release`, and with `-Werror` on the next line that fails the build. Dropping `-Xlint:-options` from line 71 as the redundant thing this comment says it is takes `:api:compileJava` down on every lane, so it's worth the comment saying it's load-bearing.

```suggestion
        // Build runs on JDK 25 (see java.toolchain) but enforces the Java 8 API surface and
        // emits Java 8 bytecode via --release 8. JDK 25's javac still warns that source/target 8
        // is obsolete, so the -Xlint:-options below is required rather than optional: without it
        // -Werror turns that warning into a build failure.
```

functional/TestUserAgent.java
allanrogerr: Confirmed — I reproduced it by pointing the client at a dead port instead of a live httpbin: `IllegalStateException: java.net.ConnectException` out of `MinioAsyncClient.throwMinioException` at 3700, straight through `main`, exit 1. It happens because `BaseS3Client`'s OkHttp `onFailure` completes the future with the raw `IOException` rather than wrapping it, so `throwMinioException` finds a cause that isn't a `MinioException` and takes its `IllegalStateException` branch.

`catch (Exception e)` does fix it: with that one change the same forced-failure run exits 0, and it still passes against a reachable host, because the trace already carries the `User-Agent:` line before the socket is used — so the version assertion below still does its job either way.

build.gradle
allanrogerr: You're right that the jar doesn't change — I built the branch with and without both blocks and got the same five `META-INF/*.kotlin_module` entries either way, so nothing is dropped and nothing is merged.

They're not precautionary though. With both blocks deleted, `:api:shadowJar` and `:adminapi:shadowJar` print ten copies of `'META-INF/okhttp.kotlin_module' is matched by KotlinModuleMetadataTransformer_Decorated but its DuplicatesStrategy is EXCLUDE …` (five markers x two modules); with them, none. Shadow emits that on the transformer's pattern match, not on an actual duplicate, so distinct paths don't make it go away. So the trade is ten warning lines per build against two blocks — your call which you'd rather have.

Either way the comment overstates its own case: it says the default EXCLUDE strategy "would drop them beforehand", and nothing is being dropped here. Shadow's own wording is the weaker "duplicates may be silently dropped".

allanrogerr:
Two lines of the description have drifted from the branch:

- It says the Gradle wrapper went to 9.6.1. That was right when it was written, but the wrapper has moved on since — 9.6.1 -> 9.7.0 -> 9.7.1 — and `gradle-wrapper.properties` now points at `gradle-9.7.1-bin.zip`.
- It says that came "with improved download retries". `retries=0` and `retryBackOffMs=500` are the wrapper's own built-in defaults (they're the field initialisers in `WrapperConfiguration`, and `WrapperExecutor` falls back to them when the keys are absent), so the download behaviour is unchanged from before — and at `retries=0` there's still no retry on a failed distribution download.

Could you refresh those two lines?

functional/TestUserAgent.java
balamurugana: No. For connection failure, the test should fail. The objective is to capture a user agent string generated by `minio-java` and validate it is set properly. Ignoring all exception is not a good idea.

build.gradle
balamurugana: Without this change, the build fails.

allanrogerr (approved):
ok

functional/TestUserAgent.java
allanrogerr: Agreed, and your version is the right one. A connection failure escaping and failing the test is the behaviour you want, and swallowing only `MinioException` keeps the httpbin-level error ignored so the User-Agent assertion still runs — `catch (Exception e)` would have collapsed those two cases into one. I was answering whether it would catch the escape rather than whether it should, and that was the wrong question.

One observation and then I'll leave it: the assertion never actually needs the remote host. The trace already carries the `User-Agent:` line before the socket is used, which is why the check still passes when the connection fails. So the only thing httpbin contributes is a way for all five Ubuntu lanes to go red when it happens to be down.

build.gradle
allanrogerr: Understood — leave them in.

For the record, I couldn't get a failure out of removing them on Linux: plain `build`, `build -Prelease` and `publishToMavenLocal` all pass at this head, and so does `build` at the revision where the blocks first went in alongside the shadow 9.4.2 to 9.6.1 bump. All four just print the ten transformer warnings. So whatever you hit was likely Windows or an intermediate state I can't reproduce from the pushed history, and I'd rather the blocks stay than have someone remove them later on the strength of my earlier comment.

If it was a hard failure, it may be worth saying so in the comment. As written it explains itself in terms of duplicate handling, which is what led anjalshireesh and me to read it as precautionary.

harshavardhana (approved):

PR #1710, opened 2026-06-24
Do not set completed if NoSuchVersion in removeObjects

balamurugana:
Previously if batch object removal gets `NoSuchVersion` error only, the iterator stops processing next batches. This is fixed by continue processing next batches for `NoSuchVersion` but stops for other errors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed object deletion to continue processing when delete responses only include ignored errors, preventing the operation from stopping too early.
  * Improved handling of partial delete failures so valid deletions can proceed as expected.


shtripat (approved):
lgtm

PR #1711, opened 2026-06-27
Fix bugs found during code audit across api, adminapi, functional, examples

balamurugana:
api:
- Compose/copy >5GiB: send per-part x-amz-copy-source-range (was loop-invariant)
- Checksum.CRC64NVME.update: bound the slicing loop by len, not p.length
- messages/Filter: keep And(String,Map) tags; "exactly one" via new Utils.exactlyOneNonNull
- Http BaseUrl: ELB endpoints derive the real region (was "com")
- messages/Checksum.headers(): emit x-amz-checksum-<algo>
- PromptObjectArgs prompt setter; ListPartsArgs.Builder extends ObjectArgs.Builder
- maxKeys() null-safe; Arrays.hashCode for array-backed args
- downloadObject error propagation + temp cleanup; snowball/appendObject RAF close
- uploadPartsParallelly buffer return/abort; executeAsync 304 return
- credential providers: AwsConfig/MinioClientConfig/MinioEnvironment NPE -> clean
- GetPresignedObjectUrlArgs expiry overflow; ObjectLockConfiguration duration();
  AccessControlPolicy null owner;
  GetObjectAttributesArgs validation; LifecycleConfiguration via Utils.exactlyOneNonNull;
  VersioningConfiguration.excludeFolders() primitive; ReplicationConfiguration text

adminapi:
- updateServiceAccount newStatus nullable (no silent disable)
- Status.fromString null-guard; thread signing Credentials through execute()
- getBucketQuota asLong(); Crypto.encrypt exact-multiple chunk; map getters
  return empty map when absent

functional:
- PutObjectRunnable surfaces thread failures; notification tests assert;
  legal-hold tests fixed; stream/temp-file cleanup;
  MintLogger throws UncheckedIOException instead of blank log

examples:
- close getObject/Response/progress streams (try-with-resources); fix
  GetObjectProgressBar/SelectObjectContent object names; real file size as object
  size; provider isSuccessful() check; placeholder credentials

Also add CLAUDE.md with build commands and architecture overview.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of missing configuration, credentials, response data, and collections.
  - Corrected checksum headers, CRC64NVME range processing, multipart uploads, encryption boundaries, HTTP errors, and progress reporting.
  - Downloads and uploads now clean up temporary files and resources more reliably.
  - Object listings default to 1,000 keys when unspecified.
  - Improved bucket conflict reporting and checksum compatibility.

- **Validation**
  - Added stricter validation for object attributes, presigned URL expiration values, and environment credentials.

- **API Changes**
  - Renamed the prompt builder method to `prompt`.
  - Service-account status updates now support an unspecified status.
  - Versioning configuration reports folder exclusion as a non-null boolean.


adminapi/src/main/java/io/minio/admin/MinioAdminClient.java
```diff
@@ -702,7 +721,7 @@ public void updateServiceAccount(
       @Nonnull String accessKey,
       @Nullable String newSecretKey,
       @Nullable Map<String, Object> newPolicy,
-      @Nullable boolean newStatus,
+      @Nullable Boolean newStatus,
```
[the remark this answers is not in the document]
balamurugana: `MinioAdminClient` does not preserve backward compatibility which is intentional as the server keeps changing. The users has to recompile with the latest `MinioAdminClient`.

api/src/main/java/io/minio/ListenBucketNotificationArgs.java
```diff
@@ -52,9 +52,7 @@ private void validateEvents(String[] events) {
     }
 
     protected void validate(ListenBucketNotificationArgs args) {
-      if (args.bucketName != null) {
-        super.validate(args);
-      }
+      super.validate(args);
```
shtripat: This drops the bucket-optional path that #1250 added in 2021 for server-wide notification listening. Base had `if (args.bucketName != null) { super.validate(args); }` here; now it's unconditional. `BaseS3Client.java:464` and `:478` still branch on `s3request.bucket() == null`, so the transport layer still supports a null bucket - this only breaks it at the Args validation layer. Nothing in `functional/` or `examples/` exercises the bucket-less path, so this ships without any test catching it.

Was dropping this intentional (did the server side stop supporting bucket-less listen), or should the guard come back? Either way it's a behavior change for existing callers, not just a bug fix, and probably deserves its own line in the description if intentional.

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -475,8 +476,8 @@ private void onResponse(final Response response) throws IOException {
                       break;
                     case 409:
                       if (s3request.bucket() != null) {
-                        code = NO_SUCH_BUCKET;
-                        message = NO_SUCH_BUCKET_MESSAGE;
+                        code = "Conflict";
+                        message = "Bucket not empty";
```
shtripat: The old `NO_SUCH_BUCKET` mapping for 409 was definitely wrong, but "Bucket not empty" is only true for one 409 cause. This is the generic no-XML-body fallback for any bucket-scoped request, so a 409 from `createBucket` (BucketAlreadyOwnedByYou/BucketAlreadyExists) would now also get labeled "Bucket not empty", which is wrong for that case. Since this path has no way to know the actual cause, I'd keep it generic like the non-bucket branch below:

```suggestion
                        code = "Conflict";
                        message = "Request conflicts with the bucket's current state";
```

adminapi/src/main/java/io/minio/admin/Crypto.java
```diff
@@ -209,7 +209,9 @@ public static byte[] encrypt(byte[] payload, String password) throws MinioExcept
     boolean done = false;
     for (int nonceId = 1; !done; nonceId++) {
       int to = from + CHUNK_SIZE;
-      if (to > payload.length) {
+      // Use >= so a payload that is an exact multiple of CHUNK_SIZE marks its final full chunk as
+      // the last one, rather than emitting an extra empty trailing chunk (matches madmin-go/sio).
+      if (to >= payload.length) {
```
shtripat, line 214: This changes which chunk gets marked "last" (and therefore the AAD used for the auth tag) for payloads that are an exact multiple of 16 KiB. It's plausible and matches typical AEAD-chunking design, but I can't check it against the actual madmin-go/sio decoder from this repo, and `CryptoTest.java` doesn't exist anymore (removed in the v9.0.0 refactor), so there's nothing here to catch a regression in either direction. Worth a boundary test (16383/16384/16385/32768-byte payloads) before merge, since a mismatch here would silently break admin API calls whenever a payload happens to land on that size - ideally confirmed against a live server, not just a self round-trip.

.github/workflows/gradle.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - name: Check limited Guava usage
         if: matrix.os == 'ubuntu-latest'
         run: |
-          if grep --with-filename --line-number --no-messages --recursive --exclude-dir=.github "com.google.common.base.Objects" .; then
+          if grep --with-filename --line-number --no-messages --recursive --exclude-dir=.github --exclude=CLAUDE.md "com.google.common.base.Objects" .; then
```
shtripat, line 27: CLAUDE.md doesn't exist anywhere in this diff or in git history (`git log --all --diff-filter=A -- CLAUDE.md` is empty), so this exclude is a no-op right now. The commit message says "Also add CLAUDE.md with build commands and architecture overview" - looks like that file got left out of the final commit.

shtripat (commented):
Went through this end to end. Most of the ~30 fixes are solid and I hand-verified several of them: the CRC64NVME bound (was using `p.length` instead of `len`, so a partial read into a reused buffer corrupted the checksum), the ELB region split (`split(regex, 1)` was a no-op split, hence "com"), the >5GiB compose per-part range headers (the correct `finalHeaders` was computed but never passed to `uploadPartCopy`), and the credential-provider NPE fixes (the `MinioEnvironmentProvider` one especially - `ChainedProvider` only catches `ProviderException`, so the old NPE would've broken chain fallback entirely, not just given a worse message).

Left a few inline comments on things I think need a decision or a test before this merges. Couldn't run the build in my environment (sandboxed, no network to fetch the Gradle 9.7.1 distribution), so no compile/lint/test results from me here - worth letting CI run as usual.

shtripat:
**"Also add CLAUDE.md with build commands and architecture overview."** (last line before the sign-off) — no CLAUDE.md exists in this diff or anywhere in git history (`git log --all --diff-filter=A -- CLAUDE.md` is empty), and `.github/workflows/gradle.yml:27` was updated in this same diff to exclude a `CLAUDE.md` from the Guava-usage grep that was never added.

Suggested fix:
> Either add the CLAUDE.md file this line describes, or drop the line (and the gradle.yml exclude) if it's not coming in this PR.

api/src/main/java/io/minio/ListenBucketNotificationArgs.java
balamurugana: Reverted the change.

adminapi/src/main/java/io/minio/admin/Crypto.java
balamurugana: Verified against the real decoder. madmin-go pins sio-go v0.3.1, whose chunk loop
is `for len(p) > w.bufSize` (strictly greater) — a payload of exactly bufSize stays
buffered and Close() seals it as last. So `>=` is the canonical framing, and it only
affects non-zero exact multiples of 16 KiB. Java's output now matches sio-go's
length exactly (16441 / 32841 / 49241 at 16384 / 32768 / 49152; old code gave
16457 / 32857 / 49257). Interop passes both directions at 0, 1, 16383, 16384,
16385, 32767, 32768, 32769, 49152.

One correction: this wouldn't have broken admin calls. sio-go's DecReader accepts
the old framing too — its lookahead treats the 16-byte trailer as a valid empty
final chunk. It's a canonicalization, not a bug fix.

Skipping a test here: a self round-trip can't catch it (DecryptReader infers the
last chunk from EOF, so it passes either way), and a real one means checking in
Go-generated vectors — not worth the fixture weight when both framings decode.

.github/workflows/gradle.yml
balamurugana: Added CLAUDE.md

api/src/main/java/io/minio/MinioAsyncClient.java
```diff
@@ -3319,11 +3363,32 @@ public CompletableFuture<ObjectWriteResponse> uploadSnowballObjects(
...
+              final RandomAccessFile file;
+              try {
+                file = new RandomAccessFile(args.stagingFilename(), "r");
               } catch (IOException e) {
                 throw new CompletionException(new MinioException(e));
               }
+              return putObject(new PutObjectAPIArgs(args, file, length, headers))
+                  .exceptionally(
+                      e -> {
+                        e = e.getCause();
```
allanrogerr: When this upload fails through `Utils.failedFuture(...)` the throwable arrives here unwrapped, so `getCause()` returns null. That happens on an ordinary path: `Checksum.update` reports a short or truncated staging file as `new MinioException("unexpected EOF")`, which has no cause. The original error is then discarded, and `throwMinioException` below dereferences the null cause, so the caller sees a NullPointerException instead of the real message. If the `file.close()` in this block also fails, `addSuppressed` hits the same null first.

Unwrapping only when there is actually a `CompletionException` to unwrap keeps the real error in both cases and leaves the already-working wrapped path unchanged.

```suggestion
                        if (e instanceof CompletionException) e = e.getCause();
```

The same handler is in `appendObject` at line 3749 and needs the same change.

api/src/main/java/io/minio/MinioAsyncClient.java
```diff
@@ -3666,19 +3731,38 @@ public CompletableFuture<ObjectWriteResponse> appendObject(AppendObjectArgs args
...
+                        null,
+                        args.length(),
+                        file,
+                        partSize,
+                        hashers,
+                        addContentSha256,
+                        addSha256Checksum)
+                    .exceptionally(
+                        e -> {
+                          e = e.getCause();
```
allanrogerr: Same as the handler at line 3375: when `appendObject` fails through `Utils.failedFuture(...)` the throwable arrives unwrapped and `getCause()` is null, so the original error is lost and `throwMinioException` below trips over the null cause. This site is the more reachable of the two, because `Checksum.update(hashers, file, size)` a few lines up returns exactly such a future when the file is shorter than the declared length.

```suggestion
                          if (e instanceof CompletionException) e = e.getCause();
```

api/src/main/java/io/minio/credentials/MinioEnvironmentProvider.java
```diff
@@ -16,11 +16,24 @@
...
+    if (accessKey == null) {
+      throw new ProviderException(
+          "Access key does not exist in MINIO_ACCESS_KEY environment variable");
+    }
+
+    String secretKey = getProperty("MINIO_SECRET_KEY");
+    if (secretKey == null) {
+      throw new ProviderException(
+          "Secret key does not exist in MINIO_SECRET_KEY environment variable");
+    }
```
allanrogerr, line 38: `getProperty` hands back the raw environment value, so `MINIO_ACCESS_KEY=` (exported but empty) gets past these null checks and reaches `Credentials`, which throws `IllegalArgumentException` for an empty key. `ChainedProvider` catches only `ProviderException`, so that escapes the whole chain instead of falling through to the next provider - and `examples/MinioClientWithChainedProvider.java` chains exactly these two providers.

`AwsEnvironmentProvider.getValue` already rejects empty values with a `ProviderException` (AwsEnvironmentProvider.java:25-31); this brings the MinIO provider in line with it.

```suggestion
    if (accessKey == null) {
      throw new ProviderException(
          "Access key does not exist in MINIO_ACCESS_KEY environment variable");
    }
    if (accessKey.isEmpty()) {
      throw new ProviderException("Empty access key in MINIO_ACCESS_KEY environment variable");
    }

    String secretKey = getProperty("MINIO_SECRET_KEY");
    if (secretKey == null) {
      throw new ProviderException(
          "Secret key does not exist in MINIO_SECRET_KEY environment variable");
    }
    if (secretKey.isEmpty()) {
      throw new ProviderException("Empty secret key in MINIO_SECRET_KEY environment variable");
    }
```

api/src/main/java/io/minio/GetObjectAttributesArgs.java
```diff
@@ -68,11 +68,17 @@ public Builder objectAttributes(List<String> objectAttributes) {
...
     public Builder maxParts(Integer maxParts) {
+      if (maxParts != null && maxParts < 1) {
+        throw new IllegalArgumentException("valid max parts must be provided");
+      }
       operations.add(args -> args.maxParts = maxParts);
       return this;
     }
 
     public Builder partNumberMarker(Integer partNumberMarker) {
+      if (partNumberMarker != null && partNumberMarker < 1) {
```
allanrogerr: `ListPartsArgs.Builder.partNumberMarker` bounds the same parameter at both ends - `partNumberMarker < 1 || partNumberMarker > 10000` at ListPartsArgs.java:61 - and throws with this exact message. This copy keeps the lower bound and drops the upper one, so a marker of 100000 is accepted here and rejected by the sibling.

```suggestion
      if (partNumberMarker != null && (partNumberMarker < 1 || partNumberMarker > 10000)) {
```

api/src/main/java/io/minio/BaseS3Client.java
```diff
@@ -475,8 +476,9 @@ private void onResponse(final Response response) throws IOException {
                       break;
                     case 409:
                       if (s3request.bucket() != null) {
-                        code = NO_SUCH_BUCKET;
-                        message = NO_SUCH_BUCKET_MESSAGE;
+                        code = "Conflict";
+                        code = "Conflict";
```
allanrogerr: Applying the suggestion from the now-resolved thread just above left the original `code = "Conflict";` in place, so it is assigned twice in a row and the first store is dead. It compiles and CI is green, so nothing flags it - and because that thread is resolved, no one is watching these lines any more.

```suggestion
                        code = "Conflict";
```

Worth a quick look at the other suggestions applied in this push, in case any of them landed the same way.

allanrogerr:
One line in the description doesn't match the diff:

- **"TestUserAgent guards"** (in the `functional:` section) — nothing in this PR touches `functional/TestUserAgent.java`. The string `UserAgent` doesn't appear anywhere in the diff, and `git log -1 -- functional/TestUserAgent.java` points at `719e476`, which is #1709. Either drop the bullet or add the change it describes.

Everything else in the description checked out against the diff.

shtripat (dismissed):
Went back through this since my last pass. All four things I flagged before are addressed: `ListenBucketNotificationArgs` has the bucket-optional path back, the 409 message in `BaseS3Client` is generic instead of "Bucket not empty", and `CLAUDE.md` is actually in the diff now so the exclude in `gradle.yml` isn't a no-op. On the `Crypto.java` chunk-boundary change, thanks for running it against the real sio-go decoder and posting the byte counts - that's convincing, even without `CryptoTest.java` to pin it down going forward.

The three comments from allanrogerr (on `GetObjectAttributesArgs`, `BaseS3Client`, `MinioEnvironmentProvider`) are anchored to a commit that's no longer in the branch's history - I checked all three against the current head and none of them reproduce, so nothing to do there.

New in this push: the `.exceptionally(e -> ...)` handlers across `MinioAsyncClient`/`BaseS3Client` now guard `e.getCause()` behind `instanceof CompletionException` - correct, since a future completed via `completeExceptionally` hands the raw exception to a directly-attached `exceptionally`, not a wrapped one, so the old unconditional unwrap could silently swap in the wrong exception. Checked all 16 sites, consistent. The credential-provider empty-value checks (throwing `ProviderException` instead of letting an empty string reach `Credentials`) and the `MinioClientConfigProvider` `"aliases"`/`"hosts"` fallback both look right.

I was able to run the full build this time (no network last round): `./gradlew build` is green across all four modules - 49 unit tests, 0 failures, 0 SpotBugs findings, spotless clean.

One thing I noticed that's worth a separate issue but isn't part of this diff: `MinioAsyncClient.removeObjects()`'s iterator sets `completed = error != null` when a batch has an error, which ties loop termination to whether that batch had an error rather than whether there's more to delete. For a >1000-object delete, an error in an early batch abandons everything after it with no signal to the caller. This is identical in `master` today and this PR doesn't touch it, so it shouldn't hold this up.

api/src/main/java/io/minio/MinioAsyncClient.java
```diff
@@ -1861,13 +1877,23 @@ private CompletableFuture<List<UploadPartResponse>> uploadPartsParallelly(
...
+                            } finally {
+                              // Always return the buffer to the pool, even on upload failure.
+                              bufferPool.put(part.args().buffer());
+                            }
                           }
                         } catch (InterruptedException e) {
+                          Thread.currentThread().interrupt();
                           errorOccurred.set(true); // signal to all threads
                           exceptions.add(e);
+                        } catch (RuntimeException e) {
```
anjalshireesh: This still leaks silent corruption on `Error`. An OOM in a worker skips both catches, but the `finally` still counts down `doneLatch`, so `errorOccurred` stays false and `exceptions` stays empty - `uploadPartsParallelly` returns a short part list and `completeMultipartUpload` assembles an object missing a part, with no exception anywhere. Can we make it `catch (Throwable t)`?

api/src/main/java/io/minio/Checksum.java
```diff
@@ -343,12 +343,13 @@ public CRC64NVME() {}
 
     @Override
     public void update(byte[] p, int off, int len) {
+      int limit = off + len;
```
anjalshireesh, line 346: Worth a unit test. `PartReader.java:154` calls `Checksum.update(hashers, buf16k, n)` with `n < 16384` on the last read of every part, so the old `p.length` bound hashed stale bytes and this was wrong for nearly every object - nothing stops that coming back. Same for the other pure fixes here: `Crypto.encrypt` with a payload of exactly `n * CHUNK_SIZE`, `Utils.xor`, the ELB region parsing in `api/src/main/java/io/minio/Http.java:158`, and the expiry overflow in `api/src/main/java/io/minio/GetPresignedObjectUrlArgs.java:76`.

CLAUDE.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Architecture
+
+**Client layering** (`api/src/main/java/io/minio/`):
+- `BaseS3Client` (abstract) — HTTP plumbing, request signing, response handling.
+- `MinioAsyncClient extends BaseS3Client` — the real implementation; every operation returns `CompletableFuture`.
+- `MinioClient` — synchronous facade that **wraps a `MinioAsyncClient`** and blocks on its futures. It does not extend the async client. A new sync operation almost always means adding the async method first, then a blocking wrapper.
+- Both are built via a `.builder()...build()` (endpoint, credentials, region, http client).
+
+**Args pattern** — every public operation takes one immutable `XxxArgs` object built via `XxxArgs.builder()`. The class hierarchy (see `arg-class-structure.txt`) is:
```
anjalshireesh: `arg-class-structure.txt` doesn't exist anywhere in the repo, and neither does `build.sh` on line 60. `docs/API.md` is fine.

anjalshireesh:
The description says "ListenBucketNotificationArgs requires bucket", but that file isn't in the diff. `api/src/main/java/io/minio/ListenBucketNotificationArgs.java:55` still has `if (args.bucketName != null) { super.validate(args); }`, so a builder with no bucket validates clean and only fails at request time. Either make the fix or drop the bullet.

Same kind of thing in the commit message: "functional: ... TestUserAgent guards" — `functional/TestUserAgent.java` is untouched here, that landed in #1709. Worth fixing before squash since it's permanent in `git log`.

CLAUDE.md
shtripat: arg-class-structure.md is in and checks out - went through the whole hierarchy table and the per-parent class lists against the actual `extends` clauses in `api/src/main/java/io/minio/*.java`, no mismatches. The `build.sh` half of this is still open though: `CLAUDE.md:61` says releases are driven by JReleaser with `build.sh` showing the deploy invocation, but there's no `build.sh` anywhere in the repo or its history. The actual deploy config is the `jreleaser { deploy { ... } }` block in `build.gradle:446`.

api/src/main/java/io/minio/Checksum.java
shtripat: Tests landed for three of these - `ChecksumTest` covers the `p.length` bug (poisons the scratch buffer tail so a regression would actually fail it), `GetPresignedObjectUrlArgsTest` covers the expiry overflow, and `MinioClientTest` covers the ELB region parsing. Ran all three fresh, 38/38 pass. `Crypto.encrypt`'s exact-multiple-chunk boundary and `Utils.xor` are still without a pinning test - same gap I flagged on `Crypto.java` last round.

shtripat (commented):
Went through the latest push addressing anjalshireesh's comments. The `MinioAsyncClient` catch-`Throwable` fix is correct and type-consistent. The new tests (`ChecksumTest`, `GetPresignedObjectUrlArgsTest`, `MinioClientTest` additions) are solid - not vacuous, ran them fresh and they'd catch a regression. Left two replies on open threads: `build.sh` referenced in `CLAUDE.md` still doesn't exist, and `Crypto.encrypt`/`Utils.xor` are still without pinning tests. Full `./gradlew build` is green.

balamurugana:
> Went through the latest push addressing anjalshireesh's comments. The `MinioAsyncClient` catch-`Throwable` fix is correct and type-consistent. The new tests (`ChecksumTest`, `GetPresignedObjectUrlArgsTest`, `MinioClientTest` additions) are solid - not vacuous, ran them fresh and they'd catch a regression. Left two replies on open threads: `build.sh` referenced in `CLAUDE.md` still doesn't exist, and `Crypto.encrypt`/`Utils.xor` are still without pinning tests. Full `./gradlew build` is green.

1. No unit test for Crypto.decrypt() here, deliberately.

A self round-trip can't catch this — DecryptReader infers the last chunk from EOF,
so it accepts both the old and new framing and passes either way. Pinning it for
real means checking in a ~16 KB ciphertext vector generated by sio-go (v0.3.1, as
madmin-go pins), an external toolchain nobody can regenerate from inside the repo.

I verified against that real decoder out-of-band instead, both directions, at 0, 1,
16383, 16384, 16385, 32767, 32768, 32769 and 49152 bytes — all pass, and Java's
framing now matches sio-go exactly. sio-go also accepts the old framing, so this
was a canonicalization rather than a live bug.

If you want something in-tree, asserting the encrypted length is
41 + n + 16*ceil(n/16384) catches this regression with no fixture. Happy to add it.


2. Renamed it to Utils.exactlyOneNonNull() instead of adding a test.

The name was the real problem: "xor" reads as parity (true for 1, 3, 5 non-null)
but the method means exactly one — and at the three-arg call sites
(PutObjectAPIBaseArgs, Filter) a parity reading would let file + buffer + data all
pass validation. The new name can't be misread. A test would only have guarded
against that misreading; the rename removes it.

Not an API break either — xor() isn't in 9.0.3, so it never shipped.

shtripat (dismissed):
lgtm

anjalshireesh:
Both bullets are gone from the description now — thanks. Dropping the commit-message half; not worth churning history over.

CLAUDE.md
```diff
@@ -0,0 +1,62 @@
...
+./gradlew spotlessApply          # auto-format (google-java-format); fixes most lint failures
+./gradlew :api:test              # unit tests for one module
+./gradlew :api:test --tests 'io.minio.MinioClientTest'        # single test class
+./gradlew :api:test --tests 'io.minio.MakeBucketArgsTest.method'  # single test method
+./gradlew :api:localeTest        # re-runs tests under locale de-DE (part of `check`)
+./gradlew runFunctionalTest      # integration tests vs play.min.io by default
+./gradlew runFunctionalTest -Pendpoint=http://localhost:9000 -PaccessKey=... -PsecretKey=... -Pregion=us-east-1
+```
+
+Unit tests are JUnit 5 (Jupiter) in `*/src/test/java`. `functional/` tests hit a real server and are not part of `build`.
```
anjalshireesh: These are JUnit 4, not Jupiter - every test in `api/src/test` and `adminapi/src/test` uses `org.junit.Test`/`org.junit.Assert`, including the three files added here, and the `test {}` block at `build.gradle:92` has no `useJUnitPlatform()`. The catch is that getting it wrong is silent: I dropped a Jupiter test class with a body that throws into `api/src/test/java/io/minio/` and `:api:test` gave BUILD SUCCESSFUL with no result file and no warning. Since this file is what an agent reads first, can we either correct the line or add `useJUnitPlatform()` plus the engine and migrate?

anjalshireesh:
Description is stale after the rename — lines 4 and 14 still say `Utils.xor`, but the method is `Utils.exactlyOneNonNull` now.

api/src/main/java/io/minio/MinioAsyncClient.java
allanrogerr: Implemented in 0b581976 (`api/src/main/java/io/minio/MinioAsyncClient.java:3377`, in `uploadSnowballObjects`). That line now reads `if (e instanceof CompletionException) e = e.getCause();`, and no bare `e = e.getCause();` line is left in the file.

You can re-check it with `git show 0b581976:api/src/main/java/io/minio/MinioAsyncClient.java | sed -n 3377p`.

api/src/main/java/io/minio/MinioAsyncClient.java
allanrogerr: Implemented in 0b581976 (`api/src/main/java/io/minio/MinioAsyncClient.java:3751`, in `appendObject`). The handler now reads `if (e instanceof CompletionException) e = e.getCause();`, the same guard as the `uploadSnowballObjects` handler at line 3377.

You can re-check it with `git show 0b581976:api/src/main/java/io/minio/MinioAsyncClient.java | sed -n 3751p`.

api/src/main/java/io/minio/credentials/MinioEnvironmentProvider.java
allanrogerr: Implemented in 0b581976 (`api/src/main/java/io/minio/credentials/MinioEnvironmentProvider.java:26-41`). The provider now throws `ProviderException` for an empty value as well as a missing one; line 31 reads `throw new ProviderException("Empty access key in MINIO_ACCESS_KEY environment variable");`.

You can re-check it with `git show 0b581976:api/src/main/java/io/minio/credentials/MinioEnvironmentProvider.java | sed -n 26,41p`.

api/src/main/java/io/minio/GetObjectAttributesArgs.java
allanrogerr: Implemented in 0b581976 (`api/src/main/java/io/minio/GetObjectAttributesArgs.java:79-80`), with the shared constant in place of the literal. The check now reads `partNumberMarker < 1 || partNumberMarker > ObjectWriteArgs.MAX_MULTIPART_COUNT`, and `api/src/main/java/io/minio/ObjectWriteArgs.java:37` sets `MAX_MULTIPART_COUNT = 10000`. `ListPartsArgs.java:61-62` uses the same check.

I also ran it at that commit: `GetObjectAttributesArgs.builder().partNumberMarker(10001)` and `.partNumberMarker(100000)` both throw `IllegalArgumentException: part number marker must be between 1 and 10000`, and `.partNumberMarker(10000)` is accepted.

api/src/main/java/io/minio/BaseS3Client.java
allanrogerr: Implemented in 0b581976 (`api/src/main/java/io/minio/BaseS3Client.java:479-480`). The 409 bucket branch now assigns `code = "Conflict";` once, followed by `message = "Request conflicts with the bucket's current state";`, and no two consecutive `code =` assignments remain in the file.

I also re-checked the other suggestions applied on this PR, and none of them left the replaced lines behind. You can re-check this one with `git show 0b581976:api/src/main/java/io/minio/BaseS3Client.java | sed -n 477,485p`.

CLAUDE.md
allanrogerr: Implemented in 0b581976 (`CLAUDE.md:31`). The line now reads: "Unit tests are JUnit 4 (`org.junit.Test`, `org.junit.Assert`) in `*/src/test/java`. **Do not write Jupiter tests there.**"

I repeated your check at that commit: a Jupiter test whose body throws, placed in `api/src/test/java/io/minio/`, compiled and produced no result file, and `./gradlew :api:test` still reported BUILD SUCCESSFUL.

adminapi/src/main/java/io/minio/admin/MinioAdminClient.java
```diff
@@ -453,14 +470,15 @@ public long getBucketQuota(String bucketName) throws MinioException {
...
-      return OBJECT_MAPPER
-          .<Map<String, JsonNode>>readValue(response.body().bytes(), mapType)
-          .entrySet()
-          .stream()
-          .filter(entry -> "quota".equals(entry.getKey()))
-          .findFirst()
-          .map(entry -> Long.valueOf(entry.getValue().toString()))
-          .orElseThrow(() -> new IllegalArgumentException("found not quota"));
+      Map<String, JsonNode> quotaEntity = OBJECT_MAPPER.readValue(response.body().bytes(), mapType);
+      JsonNode quota = quotaEntity.get("quota");
```
allanrogerr, line 474: `getBucketQuota` cannot read back a quota from AIStor RELEASE.2026-08-07T18-34-35Z. I set a 1 MiB quota: the PUT with `{"quotatype":"hard","quota":1048576}` returned 200, and the GET returned `{"size":1048576,"rate":0,"requests":0,"quotatype":"hard"}`, so this lookup throws `quota not found in response`. madmin-go v3 still carried `quota` as a deprecated field, and v4 removed it and kept `size`. The merge base reads the same key, so this is not a regression from this PR, but this PR is already rewriting the method.

You may want to read `size` and fall back to `quota` for servers that still send it. What are your thoughts?

api/src/main/java/io/minio/PromptObjectArgs.java
```diff
@@ -56,7 +56,7 @@ protected void validate(PromptObjectArgs args) {
       Utils.validateNotNull(args.promptArgs, "prompt argument");
     }
 
-    public Builder offset(String prompt) {
+    public Builder prompt(String prompt) {
```
allanrogerr, line 59: Renaming `offset(String)` to `prompt(String)` breaks existing callers. The 9.0.3 tag has `public Builder offset(String prompt)`, and a caller of `PromptObjectArgs.builder().bucket("b").object("o").offset("hello")` fails to compile against this branch with `cannot find symbol`. CodeRabbit asked in its first review to keep `offset(String)` as an alias that forwards to `prompt(String)`. The new name is the right one, and an alias would keep those callers compiling.

Would you keep the alias until the next major version?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -2,7 +2,7 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 distributionUrl=https\://services.gradle.org/distributions/gradle-9.7.1-bin.zip
 networkTimeout=10000
-retries=0
+retries=3
```
allanrogerr: `retries=3` is lost the next time someone runs `./gradlew wrapper`. I ran it on this branch, and the task rewrote this file with `retries=0`. With `tasks.named('wrapper') { retries = 3; retryBackOffMs = 500 }` added to `build.gradle`, the same command left this file unchanged, byte for byte.

Would you add that block so regenerating the wrapper keeps the retries?

allanrogerr (commented):
These two points have no changed line to attach a comment to.

1. The notification test at `functional/TestMinioClient.java` lines 2654-2659 does not check the filter. I evaluated its condition against a returned config whose suffix filter is `jpg` instead of `pg`, and against one with no filter, and the test passes both. Comparing `Xml.marshal(expectedConfig)` with `Xml.marshal(config)`, as CodeRabbit suggested in its first review, fails both and still passes an identical config.

2. No unit test pins the fixes made for the review threads. I reverted six of them together (the `CompletionException` unwrap in `uploadSnowballObjects` and `appendObject`, the empty `MINIO_ACCESS_KEY` and `MINIO_SECRET_KEY` checks, the `partNumberMarker` upper bound, the integral quota check, the `Crypto.encrypt` chunk boundary, and `catch (Throwable)` in `uploadPartsParallelly`), and `:api:test` (53 tests) and `:adminapi:test` (2 tests) still passed. The `Crypto.encrypt` boundary is listed only for completeness; the thread on it explains why a round-trip test cannot pin it. I checked three of the others on this branch, and each has a short test to write: an append from a file truncated after `AppendObjectArgs.builder().filename()` recorded its size fails with `unexpected EOF`, an empty `MINIO_ACCESS_KEY` makes `ChainedProvider` fall through to the next provider, and `partNumberMarker(10001)` throws `IllegalArgumentException`.

What are your thoughts?

allanrogerr:
I also raised https://github.com/minio/minio-java/issues/1716 @balamurugana

adminapi/src/main/java/io/minio/admin/MinioAdminClient.java
balamurugana: This is completely different issue. Adding newer madmin-go support needs to be addressed in separate work.

api/src/main/java/io/minio/PromptObjectArgs.java
balamurugana: Not necessarily. The arg class was wrongly implemented and users need to fix it anyway.

gradle/wrapper/gradle-wrapper.properties
balamurugana: Reverted the change. Lets keep the default behavior.

balamurugana:
> I also raised #1716 @balamurugana

Commented. IMO we don't need.

api/src/main/java/io/minio/GetObjectAttributesArgs.java
```diff
@@ -68,11 +68,19 @@ public Builder objectAttributes(List<String> objectAttributes) {
...
+        throw new IllegalArgumentException("max parts must be greater than 0");
+      }
       operations.add(args -> args.maxParts = maxParts);
       return this;
     }
 
     public Builder partNumberMarker(Integer partNumberMarker) {
+      // Matches minio-go, which sends x-amz-part-number-marker only when the value is positive
+      // and imposes no upper bound. Null means unset; zero and negatives carry no meaning.
+      if (partNumberMarker != null && partNumberMarker <= 0) {
```
anjalshireesh, line 81: This now disagrees with `api/src/main/java/io/minio/ListPartsArgs.java:62`, which still rejects anything above `ObjectWriteArgs.MAX_MULTIPART_COUNT`. Same parameter, same S3 concept, and the divergence is new in this PR - the earlier push added the shared bound here to match `ListPartsArgs`, this one removed it from just this side. If the minio-go argument holds it applies to both, so can we align them either way?

anjalshireesh:
`gradle/wrapper/gradle-wrapper.properties` went from `retries=3` back to `retries=0` in the last push, so the file now matches master exactly and the change has dropped out of the PR entirely. Nothing in the discussion asked for that — looks like collateral from regenerating the wrapper, since `retries=0` is Gradle 9.7.1's generated default. Worth putting `retries=3` back unless it was deliberate.

api/src/main/java/io/minio/GetObjectAttributesArgs.java
shtripat: Looked into this: part numbers are capped at 10,000 across every S3 multipart operation, so a marker above `MAX_MULTIPART_COUNT` can never match a real part on either call - leaving it unbounded here doesn't unlock anything the server will honor, it just moves the rejection from client to server. `ListPartsArgs.java:61-62` already validates this bound and nothing broke doing that. I'd align `GetObjectAttributesArgs` back to the same check rather than the other way, and either way add a builder test for `partNumberMarker` on both classes - right now there's zero coverage for either one, which is exactly how this flipped twice with nothing catching it.

shtripat (commented):
Went through everything since round 3 (`84f8e209`). Small delta - `CLAUDE.md`'s build.sh/JUnit claims are both fixed now, the notification filter test actually checks prefix/suffix, and there's a real regression test for the empty-env-var case on `MinioEnvironmentProvider`. The `Utils.xor`->`exactlyOneNonNull` rename is cosmetic. Ran the full build, targeted unit tests, spotless, and spotbugs locally - all green, matches CI.

One thing did regress: `GetObjectAttributesArgs`'s `partNumberMarker` upper bound, added earlier in response to review feedback, was removed again in the latest push. Left a reply on that thread with a recommendation.

api/src/main/java/io/minio/GetObjectAttributesArgs.java
balamurugana: I do not see any validation in `minio-go` for `ListParts` S3 API and I do not want to do it here, hence the validation in ListPartsArgs is correct.

balamurugana:
> `gradle/wrapper/gradle-wrapper.properties` went from `retries=3` back to `retries=0` in the last push, so the file now matches master exactly and the change has dropped out of the PR entirely. Nothing in the discussion asked for that — looks like collateral from regenerating the wrapper, since `retries=0` is Gradle 9.7.1's generated default. Worth putting `retries=3` back unless it was deliberate.

We do not want to change the default behavior of generated wrapper. Let it be the default.

balamurugana:
> Went through everything since round 3 (`84f8e209`). Small delta - `CLAUDE.md`'s build.sh/JUnit claims are both fixed now, the notification filter test actually checks prefix/suffix, and there's a real regression test for the empty-env-var case on `MinioEnvironmentProvider`. The `Utils.xor`->`exactlyOneNonNull` rename is cosmetic. Ran the full build, targeted unit tests, spotless, and spotbugs locally - all green, matches CI.
> 
> One thing did regress: `GetObjectAttributesArgs`'s `partNumberMarker` upper bound, added earlier in response to review feedback, was removed again in the latest push. Left a reply on that thread with a recommendation.

`GetObjectAttributesArgs`'s `partNumberMarker` is checked as per `minio-go`.

shtripat (approved):

anjalshireesh (approved):

